PR #365, opened 2026-03-30
Support external Bitcoin nodes in hashi-localnet CLI

0xsiddharthks:
Extends the hashi-localnet CLI to support running against external Bitcoin nodes (signet, testnet4), instead of only spawning a local regtest instance.

## Usage

```
  # Regtest (unchanged default behavior)                                                                                                          
  hashi-localnet start                                                                                                                          
                                                                                                                                                  
  # Signet (connects to local signet node)
  hashi-localnet start --bitcoin-network signet \                                                                                                 
    --btc-rpc-url http://127.0.0.1:38332 \                                                                                                      
    --btc-wallet my_signet_wallet \                                                                                                               
    -v
                                                                                                                                                  
  # Then in another terminal:                                                                                                                     
  hashi-localnet deposit --amount 10000
```

### DIFF
--- a/crates/e2e-tests/src/hashi_network.rs
+++ b/crates/e2e-tests/src/hashi_network.rs
@@ -17,11 +17,25 @@ use sui_transaction_builder::ObjectInput;
 use sui_transaction_builder::TransactionBuilder;
 use tracing::debug;
 
-use crate::BitcoinNodeHandle;
 use crate::SuiNetworkHandle;
 
 const POLL_INTERVAL: std::time::Duration = std::time::Duration::from_millis(500);
 const TEST_WEIGHT_DIVISOR: u16 = 100;
+
+/// Trait for Bitcoin node connectivity used by the hashi network builder.
+pub trait BitcoinNodeInfo {
+    fn rpc_url(&self) -> &str;
+    fn p2p_address(&self) -> String;
+}
+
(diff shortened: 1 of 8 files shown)


crates/e2e-tests/src/hashi_network.rs
```diff
@@ -17,12 +17,26 @@ use sui_transaction_builder::ObjectInput;
...
+}
+
+impl BitcoinNodeInfo for crate::BitcoinNodeHandle {
+    fn rpc_url(&self) -> &str {
+        self.rpc_url()
+    }
+    fn p2p_address(&self) -> String {
+        self.p2p_address()
+    }
+}
```
bmwill, line 38: Why do we need this trait?

crates/e2e-tests/src/hashi_network.rs
```diff
@@ -295,7 +325,7 @@ impl HashiNetworkBuilder {
         self,
         dir: &Path,
         sui: &SuiNetworkHandle,
-        bitcoin: &BitcoinNodeHandle,
+        bitcoin: &impl BitcoinNodeInfo,
```
bmwill, line 328: would it maybe be better to just pass in the rpc and p2p values directly vs needing to go through a trait?

PR #389, opened 2026-04-02
Add backup save and restore commands for config

dlukegordon:
- This PR adds a backup subcommand, with two subcommands under it, save and restore.
- This is step 1 in a series of PRs to enable backup and restore of all critical information for operating a node. Later PRs will include backing up the database, and anything else necessary.
- Save takes the config.toml file, and any file it references (such as the bitcoin private key path), tars them with a manifest to allow restoring to their original locations later, and encrypts them to an age pubkey.
- Restore reverses this processes, decrypting and untarring the backup, and optionally restoring the files to their original location.
- Includes unit tests, and I have also manually tested with an age private key stored on a yubikey. The restore process properly works with pinentry and waits for a yubikey tap before proceeding.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ anyhow = "1.0.98"
 toml = "0.9.2"
 futures = "0.3.31"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
+age = { version = "0.11", features = ["cli-common", "plugin"] }
 rustls = { version = "0.23.29", default-features = false, features = ["ring"] }
 ed25519-dalek = { version = "2.2.0", features = ["pem", "pkcs8", "rand_core"] }
 x509-parser = { version = "0.17.0", features = ["verify"] }
@@ -67,6 +68,7 @@ test-strategy = "0.4"
 proptest = { version = "1.6.0", default-features = false, features = ["std"] }
 tempfile = "3.23.0"
 jiff = "0.2"
+tar = "0.4"
 
 # fastcrypto
 fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto.git", rev = "ff39c4b8e7fa041ecc43789be214e3b2dc8027c2" }
--- a/crates/hashi/src/cli/commands/backup.rs
+++ b/crates/hashi/src/cli/commands/backup.rs
@@ -18,6 +18,7 @@ use std::fs::OpenOptions;
 use std::io;
 use std::io::ErrorKind;
 use std::io::Read;
+use std::os::unix::fs::OpenOptionsExt;
 use std::path::Path;
 use std::path::PathBuf;
 use std::str::FromStr;
@@ -203,6 +204,7 @@ fn write_manifest_to_extract_dir(extract_dir: &Path, manifest_toml: &str) -> Res
     let mut file = OpenOptions::new()
         .write(true)
         .create_new(true)
+        .mode(0o600)
         .open(&manifest_path)
         .map_err(|e| match e.kind() {
             ErrorKind::AlreadyExists => anyhow::anyhow!(
@@ -263,16 +265,39 @@ fn build_backup_manifest(files: &[PathBuf]) -> Result<BackupManifest> {
     let mut manifest_files = Vec::with_capacity(files.len());
 
     for file in files {
-        let archive_name = PathBuf::from(file.file_name().ok_or_else(|| {
-            anyhow::anyhow!("Backup input does not have a file name: {}", file.display())
-        })?);
+        let base_name = file
+            .file_name()
+            .ok_or_else(|| {
+                anyhow::anyhow!("Backup input does not have a file name: {}", file.display())
+            })?
+            .to_string_lossy();
 
-        if !archive_names.insert(archive_name.clone()) {
-            anyhow::bail!(
-                "Backup input file name collision in archive: {}",
-                archive_name.display()
-            );
-        }
+        let archive_name = if archive_names.contains(base_name.as_ref()) {
+            let stem = Path::new(base_name.as_ref())
+                .file_stem()
+                .unwrap_or_default()
+                .to_string_lossy();
+            let ext = Path::new(base_name.as_ref())
+                .extension()
+                .map(|e| format!(".{}", e.to_string_lossy()))
+                .unwrap_or_default();
+
+            let mut suffix = 2u32;
+            loop {
+                let candidate = format!("{stem}-{suffix}{ext}");
+                if archive_names.insert(candidate.clone()) {
+                    print_info(&format!(
+                        "Archive name collision for {base_name}: renamed to {candidate} (original: {})",
+                        file.display()
+                    ));
+                    break PathBuf::from(candidate);
+                }
+                suffix += 1;
+            }
+        } else {
+            archive_names.insert(base_name.to_string());
+            PathBuf::from(base_name.as_ref())
+        };
 
         manifest_files.push(BackupManifestFile {
             archive_name,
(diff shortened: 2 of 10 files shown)


bmwill:
2k lines is a lot and i haven't had time to look at everything yet. As a first pass I had claude do a review:

```
Issues

  1. Restored files lose their original mode (security). crates/hashi/src/cli/commands/backup.rs:425 and :482
  open restored files via OpenOptions::new().write().create_new(), which uses the default umask (typically 0644).
   Private keys — keypair_path and bitcoin.private_key_path — are the whole point of this backup, and restoring
  them world-readable is a real downgrade from the originals. Either preserve the mode from the tar header (tar
  crate exposes entry.header().mode()) or explicitly set 0o600 on restored files. This matters especially for
  copy_to_original_paths, where the user reasonably expects the original permissions back.

  2. Basename collision is a hard failure. build_backup_manifest (backup.rs:323) bails if any two backup inputs
  share a file name. If a user ever points keypair_path and bitcoin.private_key_path at two files both called,
  say, keypair.pem in different directories, backup breaks with no workaround short of renaming a file on disk.
  Consider disambiguating (e.g., prefixing with a short hash of the parent path, or numbering collisions) rather
  than failing. At minimum the error message should suggest the fix.

  3. Restore doesn't validate tar entry types. restore_backup_entries (backup.rs:388) iterates entries and
  io::copys them, but never checks entry.header().entry_type(). A backup built today only contains regular files,
   but if a future change (or a hand-crafted archive) introduces a symlink or directory entry, io::copy silently
  writes zero bytes and the restore "succeeds" with empty files. A quick ensure!(header.entry_type().is_file(),
  ...) guards the invariant. The path-root check at :410 already blocks traversal, which is good.

  4. Minor: IdentityFile::from_file via display().to_string(). load_backup_identities (backup.rs:359) converts
  the path through display(), which is lossy for non-UTF-8 paths. Unlikely to hit in practice, but
  to_string_lossy().into_owned() is equally non-strict and more honest, or you could bail early on non-UTF-8.

  5. Test coverage gaps. The tests cover round-trip and overwrite refusal nicely, but nothing exercises: wrong
  identity file (decrypt failure path), basename collision, or the "no config loaded" bail. The decrypt-failure
  case in particular would catch regressions in error messaging, which users will hit when they fat-finger the
  identity file.
```

dlukegordon:
> 2k lines is a lot and i haven't had time to look at everything yet. As a first pass I had claude do a review:
> 
> ```
> Issues
> 
>   1. Restored files lose their original mode (security). crates/hashi/src/cli/commands/backup.rs:425 and :482
>   open restored files via OpenOptions::new().write().create_new(), which uses the default umask (typically 0644).
>    Private keys — keypair_path and bitcoin.private_key_path — are the whole point of this backup, and restoring
>   them world-readable is a real downgrade from the originals. Either preserve the mode from the tar header (tar
>   crate exposes entry.header().mode()) or explicitly set 0o600 on restored files. This matters especially for
>   copy_to_original_paths, where the user reasonably expects the original permissions back.
> 
>   2. Basename collision is a hard failure. build_backup_manifest (backup.rs:323) bails if any two backup inputs
>   share a file name. If a user ever points keypair_path and bitcoin.private_key_path at two files both called,
>   say, keypair.pem in different directories, backup breaks with no workaround short of renaming a file on disk.
>   Consider disambiguating (e.g., prefixing with a short hash of the parent path, or numbering collisions) rather
>   than failing. At minimum the error message should suggest the fix.
> 
>   3. Restore doesn't validate tar entry types. restore_backup_entries (backup.rs:388) iterates entries and
>   io::copys them, but never checks entry.header().entry_type(). A backup built today only contains regular files,
>    but if a future change (or a hand-crafted archive) introduces a symlink or directory entry, io::copy silently
>   writes zero bytes and the restore "succeeds" with empty files. A quick ensure!(header.entry_type().is_file(),
>   ...) guards the invariant. The path-root check at :410 already blocks traversal, which is good.
> 
>   4. Minor: IdentityFile::from_file via display().to_string(). load_backup_identities (backup.rs:359) converts
>   the path through display(), which is lossy for non-UTF-8 paths. Unlikely to hit in practice, but
>   to_string_lossy().into_owned() is equally non-strict and more honest, or you could bail early on non-UTF-8.
> 
>   5. Test coverage gaps. The tests cover round-trip and overwrite refusal nicely, but nothing exercises: wrong
>   identity file (decrypt failure path), basename collision, or the "no config loaded" bail. The decrypt-failure
>   case in particular would catch regressions in error messaging, which users will hit when they fat-finger the
>   identity file.
> ```

Fyi, ~1200 lines of the diff are from the `Cargo.lock`, so this PR isn't as big as it seems.

To address the feedback:

1. I made it so that all restored files have 0600 permissions.
2. I changed things so that in the event of a filename collision, a numeric suffix is appended (ie `-2`) to the name of the file in the tarball. If restore is then used with `--copy-to-original-paths`, the original file name and path will be restored, the suffix dropped.
3. I added a check on restore to ensure each tar entry is a regular file, in the unlikely event of restoring a manually crafted malformed encrypted tarball.
4. I added a check to ensure the passed `backup_age_identity` values are valid utf-8.
5. I added a test for basename collision. For decryption failure, and no config loaded, the program cleanly prints an error message and exits, so it seems like overkill to add test cases for those. Can do if you want.

Cargo.toml
```diff
@@ -45,6 +45,7 @@ anyhow = "1.0.98"
 toml = "0.9.2"
 futures = "0.3.31"
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
+age = { version = "0.11", features = ["cli-common", "plugin"] }
```
bmwill, line 48: this does add a bunch of deps to the main binary. We should probably do a check to see if these additions still result in a deterministicly built binary or not.

crates/hashi/src/cli/config.rs
```diff
@@ -292,3 +364,43 @@ sui_rpc_url = "https://fullnode.mainnet.sui.io:443"
...
+/// An age recipient that can be used as the target of a config backup.
+///
+/// Supports both native x25519 recipients (`age1...`) and plugin recipients
+/// (`age1<plugin-name>1...`, e.g. `age1yubikey1...`). Plugin recipients are only
+/// resolved against a plugin binary at encryption time, so storing one in the
+/// config does not require the plugin to be installed.
+#[derive(Clone)]
+pub enum BackupRecipient {
+    Native(x25519::Recipient),
+    Plugin(plugin::Recipient),
```
bmwill, line 377: can you explain a little bit about how plugins work here?

crates/hashi/src/cli/commands/backup.rs
```diff
@@ -0,0 +1,762 @@
...
+    print_success(&format!("Backup completed: {}", output_path.display()));
+
+    Ok(())
+}
+
+/// Materialize a `BackupRecipient` into a concrete `age::Recipient` trait object
+/// suitable for passing to `Encryptor::with_recipients`.
+///
+/// For plugin recipients, this is where `age-plugin-<name>` is looked up on
+/// `$PATH`, so this call will fail if the plugin binary is not installed.
```
bmwill, line 99: ah i missed this

crates/hashi/src/cli/config.rs
dlukegordon: Age encryption supports storing a private key on a yubikey through [age-plugin-yubikey](https://github.com/str4d/age-plugin-yubikey).

If we are using this, the recipient contains a different hrp to indicate it's for a yubikey, and the identity file contains a stub pointing to the yubikey instead of the actual private key. To support this case, we need to handle things a bit differently, hence the enum.

bmwill (approved):
Awesome, thanks for this!

PR #413, opened 2026-04-06
Make `t` and `allowed_delta` configurable

zhouwfang:
#368

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -121,7 +121,7 @@ impl MpcManager {
         encryption_key: PrivateKey<EncryptionGroupElement>,
         signing_key: Bls12381PrivateKey,
         public_message_store: Box<dyn PublicMessagesStore>,
-        threshold_basis_points: u16,
+        threshold_in_basis_points: u16,
         weight_reduction_allowed_delta: u16,
         chain_id: &str,
         weight_divisor: Option<u16>,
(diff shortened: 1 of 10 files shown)


bmwill (commented):
move changes lgtm @Bridgerz want to take a look as well?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -144,8 +146,12 @@ impl MpcManager {
             .get(&epoch)
             .ok_or_else(|| MpcError::InvalidConfig(format!("no committee for epoch {epoch}")))?
             .clone();
-        // TODO: Pass t and f as arguments instead of computing them
-        let (nodes, threshold) = build_reduced_nodes(&committee, allowed_delta, weight_divisor)?;
+        let (nodes, threshold) = build_reduced_nodes(
+            &committee,
+            threshold_basis_points,
```
jonas-lj: The name confused me a bit. You call it basis points here because you're defining the threshold as some ratio with a fixed denominator, MAX_BASIS_POINTS, and then use this to compute the threshold with the actual given weights, right? if so, I'd call this something else, like TOTAL_BASIS_WEIGHT, or just document the relationship with THRESHOLD_BASIS_POINTS.

zhouwfang: I renamed `threshold_basis_points` to be `threshold_in_basis_points` to make it clear this is "the threshold, expressed in basis points" rather than "basis points of the threshold." I kept `MAX_BASIS_POINTS` rather than `TOTAL_BASIS_WEIGHT`, which sounds like a sum of something that could vary rather than a fixed constant.

jonas-lj (approved):

PR #417, opened 2026-04-07
Support varying `t` and `allowed_delta` across epochs

zhouwfang:
#368


jonas-lj:
Looks alright to me, but I'm not too confident in how the committee config is on-chain, so perhaps someone else should also take a look?

zhouwfang:
> Looks alright to me, but I'm not too confident in how the committee config is on-chain, so perhaps someone else should also take a look?

@bmwill Could you take a look? Thanks.

packages/hashi/sources/core/committee/committee.move
```diff
@@ -36,10 +36,19 @@ public struct Committee has copy, drop, store {
     members: vector<CommitteeMember>,
     /// Total voting weight of the committee.
     total_weight: u64,
+    /// MPC threshold in basis points
+    mpc_threshold_in_basis_points: u64,
+    /// Allowed delta for weight reduction
+    mpc_weight_reduction_allowed_delta: u64,
```
bmwill, line 42: @Bridgerz can you take a look at this PR? We should evaluate if we need to have extendability for epoch-specific configs like this one.

Bridgerz: We could absolutely make this more extensible via applying our same config pattern we already have, something like: 
```
  public struct Committee has copy, drop, store {
      epoch: u64,
      members: vector<CommitteeMember>,
      total_weight: u64,
      epoch_config: VecMap<String, ConfigValue>,  // snapshot of relevant config keys
  }
```

However I question the need for extensible epoch pinned configs at all. The two fields added to Committee are baked into DKG itself so makes sense... do we really expect more DKG specific configurations that require pinning to an epoch? To be clear if we did need another field on committee there is a migration path forward (not an easy one, but doable). 

We don't know what we don't know... but just want to make sure we aren't over complicating. @zhouwfang what do you think?

zhouwfang: For MPC we only expect 3: t, f (added in #462), and allowed_delta. I guess the extensibility doesn't pay off for a fixed set of 3.

bmwill (approved):

PR #421, opened 2026-04-08
Add MPC profiling metrics

zhouwfang:
#515


crates/hashi/src/grpc/client.rs
```diff
@@ -105,34 +106,41 @@ impl Client {
...
         let response = self
             .mpc_service_client()
             .retrieve_messages(proto_request)
             .await?;
-        RetrieveMessagesResponse::try_from(response.get_ref())
-            .map_err(|e| tonic::Status::internal(e.to_string()))
+        let response_size = response.get_ref().encoded_len();
+        let resp = RetrieveMessagesResponse::try_from(response.get_ref())
+            .map_err(|e| tonic::Status::internal(e.to_string()))?;
+        Ok((resp, request_size, response_size))
```
bmwill: If we want to add metrics to the clients/servers then we should leverage middleware as it will be much cleaner and keep the business logic free from this instrumentation. 

You can take a look at https://github.com/MystenLabs/sui-rust-sdk/pull/237 for how middleware can be added to grpc clients and https://github.com/bmwill/sui/blob/bf4f4c2e66964ba3668baa2accbf0521c9aa0e81/crates/sui-rpc-api/src/lib.rs#L229-L243 for how middleware can be added to the server side (I think we do already do some server side metrics using this paradigm.)

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -720,6 +745,10 @@ impl MpcManager {
             })
             .await?
         };
+        metrics
+            .mpc_dealer_crypto_duration_seconds
+            .with_label_values(&[MPC_LABEL_DKG])
+            .observe(crypto_start.elapsed().as_secs_f64());
```
bmwill: For timers we probably should use https://docs.rs/prometheus/latest/prometheus/struct.Histogram.html#method.start_timer

crates/hashi/src/grpc/client.rs
zhouwfang: Reverted the `client.rs` changes. Byte counting now lives in `RpcP2PChannel` which is the metrics wrapper around the client, so `grpc::Client` stays clean.

I looked at both references. The challenge with applying `CallbackLayer` on the client side for our use case: `Client` instances are cached in `CommitteeSet` and shared across protocols. The middleware wouldn't know which protocol is active, so we'd lose protocol labels on the bytes metrics. We need per-protocol labels to compare communication cost across protocols.                                                                                         
                                                            
`RpcP2PChannel` is created per-protocol with the label, so it's the natural place for this.

crates/hashi/src/grpc/client.rs
```diff
@@ -43,6 +43,18 @@ pub type BoxedChannel = BoxCloneService<http::Request<Body>, http::Response<Body
 
 const DEFAULT_MAX_DECODING_MESSAGE_SIZE: usize = 4 * 1024 * 1024;
 
+pub const MPC_PROTOCOL_METADATA_KEY: &str = "mpc-protocol";
```
bmwill:
```suggestion
pub const MPC_PROTOCOL_METADATA_KEY: &str = "x-hashi-mpc-protocol";
```

PR #423, opened 2026-04-09
[7/n][guardian-integration] wire guardian call into withdrawal flow

0xsiddharthks:
## Summary

In #495 and #523, we wired the local limiter and the on-chain `WithdrawalSignedEvent` observer. This PR closes the loop by actually calling the guardian during the withdrawal flow.

Post-MPC, before `sign_withdrawal`, the leader fans out a committee-signed `StandardWithdrawalRequest`, forwards it to the guardian, and verifies the `GuardianSigned` response against the pubkey we cached at bootstrap.

## Note

As of today, we are doing two fan outs: 1 for the MPC signing. followed by 1 for the BLS  cert generation for the guardian call.

Right now they are happening sequentially. However, we can parallelize them.

Adding that as a TODO for later on.

## Changes
- New `SignGuardianWithdrawalRequest` BridgeService RPC. The leader fans this out to every committee member; each one reconstructs the request deterministically from on-chain state via `build_guardian_withdrawal_request`, bounds the leader-supplied `timestamp_secs` against its own latest checkpoint (tolerance: 600s), and BLS-signs. The leader aggregates into a committee cert before forwarding to the guardian.

- `finalize_withdrawal_through_guardian` runs post-MPC, pre-`sign_withdrawal`. Failures are labeled under the new `withdrawal_signing` operation: `GuardianSeqMismatch`, `GuardianRateLimited`, `GuardianUnavailable`.
- `max_concurrent` cap of 1 (introduced in #523 for limiter advance ordering) now also covers the guardian's strict `timestamp_secs` ordering.
- e2e tests: `assert_eq!(local_state, guardian_state)` after a successful withdrawal — the round-trip integration check.


crates/hashi-guardian/src/withdraw.rs
```diff
@@ -99,9 +99,11 @@ async fn normal_withdrawal_inner(
 
     info!("Checking rate limits.");
     let consumed_amount_sats = request.utxos().external_out_amount().to_sat();
+    let wid = *request.wid();
```
bmwill: why is this a u64 if the actual id is 32 bytes?

crates/hashi-types/proto/sui/hashi/v1alpha/bridge_service.proto
```diff
@@ -165,3 +168,21 @@ message SignWithdrawalConfirmationRequest {
...
+
+// The leader sends the withdrawal-transaction id along with the guardian-specific
+// fields (timestamp, seq) so each validator can independently reconstruct and
+// BLS-sign the same `StandardWithdrawalRequest`.
+message SignGuardianWithdrawalRequestRequest {
+  // The id of the WithdrawalTransaction on Sui (32 bytes).
+  bytes withdrawal_txn_id = 1;
+
+  // Timestamp in unix seconds (used for guardian rate limiting).
+  uint64 timestamp_secs = 2;
```
bmwill: So this is expected to be a semi-recent timestamp from a checkpoint correct?

crates/hashi/src/withdrawals.rs
```diff
@@ -1244,3 +1265,93 @@ fn script_pubkey_from_raw_address(address_bytes: &[u8]) -> anyhow::Result<bitcoi
...
+
+    let utxos = TxUTXOs::new(inputs, outputs)
+        .map_err(|e| anyhow!("Failed to build guardian TxUTXOs: {e}"))?;
+
+    let wid = compute_withdrawal_wid(&txn.request_ids);
+
+    Ok(hashi_types::guardian::StandardWithdrawalRequest::new(
+        wid,
+        utxos,
+        timestamp_secs,
```
bmwill, line 1363: The timestamp needs to be validated that it is within some acceptable range from this node's view of time.

crates/hashi/src/withdrawals.rs
```diff
@@ -1244,3 +1265,93 @@ fn script_pubkey_from_raw_address(address_bytes: &[u8]) -> anyhow::Result<bitcoi
...
+    ))
+}
+
+/// Deterministic 64-bit id for a withdrawal — leading bytes of Blake2b256 over
+/// the BCS-encoded request ids. Stable across restarts and leader rotations.
+pub fn compute_withdrawal_wid(request_ids: &[Address]) -> u64 {
+    let bytes = bcs::to_bytes(&request_ids).expect("serialization should succeed");
+    let hash = Blake2b256::digest(&bytes);
+    u64::from_le_bytes(hash.digest[..8].try_into().unwrap())
+}
```
bmwill: We shouldn't need this, a Withdrawal already has a stable ID, its UID. Lets use that instead of coming up with a new id mechanism.

crates/hashi-guardian/src/enclave.rs
```diff
@@ -310,6 +310,7 @@ impl EnclaveState {
 
     pub async fn consume_from_limiter(
         &self,
+        wid: u64,
```
mskd12: But wid is not actually used inside consume, no? is this for future logging purposes?

crates/hashi-guardian/src/withdraw.rs
0xsiddharthks: Fixed in #522 .
`WithdrawalID` is now `[u8; 32]` end-to-end

crates/hashi-types/proto/sui/hashi/v1alpha/bridge_service.proto
0xsiddharthks: Yes. The leader sources it from the on-chain `WithdrawalTransaction.timestamp_ms` (the checkpoint timestamp at which the withdrawal-tx object was committed) and converts to seconds

Each hashi node then bounds it against it's own latest checkpoint with a 600s tolerance before BLS-signing .

crates/hashi-guardian/src/enclave.rs
0xsiddharthks: Yeah, the core reason for the `wid` is to be used as a cache key lateron.

crates/hashi/src/leader/mod.rs
```diff
@@ -1355,7 +1361,21 @@ impl LeaderService {
             .map(|s| s.to_byte_array().to_vec())
             .collect();
 
-        // 3. Build the WithdrawalTxSigning and get BLS certificate via fan-out
+        // 3. Post-MPC: forward to guardian for the enclave signature.
+        if let (Some(guardian), Some(seq)) = (inner.guardian_client(), expected_limiter_seq) {
```
bmwill, line 1365: do we want to also send the mpc signatures to the guardian?

bmwill (approved):
lgtm with an extra question

crates/hashi/src/leader/mod.rs
bmwill: I do think that we probably want this. I wonder if we could probably skip this extra round of committee signatures for getting approval from the guardian to sign if we do something slightly different.

We already have a committee signature when we commit to the txn itself so i wonder if that (assuming its the same epoch?) signature plus the actual MPC schnorr signatures would be enough to have the guardian prove to provide a signature.

bmwill: As discussed we can leave this as an optimization later

PR #425, opened 2026-04-09
Harden presig pool

zhouwfang:
#193

### DIFF
--- a/crates/hashi/src/leader/mod.rs
+++ b/crates/hashi/src/leader/mod.rs
@@ -1029,6 +1029,34 @@ impl LeaderService {
             }
         };
 
+        // Verify that enough presigs are available before committing
+        let num_inputs = approval.selected_utxos.len() as u64;
+        {
+            let num_consumed = inner.onchain_state().state().hashi().num_consumed_presigs;
+            let needed_end = num_consumed + num_inputs;
+            match inner.try_signing_manager() {
+                Some(sm) => {
+                    let mgr = sm.read().unwrap();
+                    let available_end = mgr.available_presig_end_index();
+                    if needed_end > available_end {
+                        warn!(
+                            "Presig pool insufficient: need index {needed_end}, \
+                             pool ends at {available_end}. \
+                             Waiting for refill before committing withdrawal.",
+                        );
+                        return Ok(());
+                    }
+                }
+                None => {
+                    warn!(
+                        "SigningManager not initialized; \
+                         cannot commit withdrawal without presigs.",
+                    );
+                    return Ok(());
+                }
+            }
+        }
+
         // Submit commit_withdrawal_tx to Sui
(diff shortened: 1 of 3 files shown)


crates/hashi/src/mpc/service.rs
```diff
@@ -39,6 +39,7 @@ const RETRY_INTERVAL: Duration = Duration::from_secs(10);
 const RPC_TIMEOUT: Duration = Duration::from_secs(5);
 const MAX_PROTOCOL_ATTEMPTS: u32 = 3;
 const START_RECONFIG_POLL_INTERVAL: Duration = Duration::from_millis(500);
+const MPC_PROTOCOL_TIMEOUT: Duration = Duration::from_secs(120);
```
bmwill: do you think this is long enough of a time window? What happens if these timeouts don't line up perfectly between different nodes on the network?

crates/hashi/src/leader/mod.rs
```diff
@@ -1029,6 +1029,34 @@ impl LeaderService {
...
+                Some(sm) => {
+                    let mgr = sm.read().unwrap();
+                    let available_end = mgr.available_presig_end_index();
+                    if needed_end > available_end {
+                        warn!(
+                            "Presig pool insufficient: need index {needed_end}, \
+                             pool ends at {available_end}. \
+                             Waiting for refill before committing withdrawal.",
+                        );
+                        return Ok(());
```
bmwill: Given we can ask for large clips of signatures, we probably want to be proactive about this otherwise we can end up in a situation where we need X presigs but that extends past the available range and we have enough available so that we are not going to proactively add a new batch leading to a live-lock of a sort.

The test that is timing out is likely due to this. So we probably don't want to actually gate or stop trying to create and commit to transactions even though we don't have enough presigs. What we do want is, if we need a presig past the available range when we get to signing (or MPC can monitor the onchain value and be more proactive about refilling) we kick off another refill so that signing can complete.

crates/hashi/src/mpc/service.rs
zhouwfang: The timeout doesn't need to line up between nodes. Each node independently times out its own call. If node A times out but node B is still running the protocol, nothing bad happens. That being said, I changed it to 300s.

crates/hashi/src/mpc/signing.rs
```diff
@@ -225,6 +238,10 @@ impl SigningManager {
                     {
                         &mut mgr.batches[b]
                     } else {
+                        if mgr.next_batch.is_none() {
+                            let next = mgr.batch_index() + 1;
+                            let _ = mgr.refill_tx.send(next);
```
bmwill: can we just call trigger_refill here to dedup the logic?

bmwill (approved):

PR #435, opened 2026-04-10
Fix key rotation stuck

zhouwfang:
#95

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -94,7 +94,8 @@ pub struct MpcManager {
     previous_output: Option<MpcOutput>,
     pub batch_size_per_weight: u16,
 
-    // Mutable during the epoch
+    // TODO: Rename these fields so it is clear at the call site which are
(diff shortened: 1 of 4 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -258,11 +258,19 @@ impl MpcManager {
             if let Some(response) = self.message_responses.get(&sender) {
                 return Ok(response.clone());
             }
-            return Err(MpcError::InvalidMessage {
-                sender,
-                reason: "Message previously received but no valid response was produced"
-                    .to_string(),
-            });
+            // No cached response, but the incoming hash matches what we have
```
benr-ml: nice catch! maybe let's rename things so it would be clear in the call site what is cached and what is stored persistently (we might have similar bugs elsewhere)

zhouwfang: Added a TODO for renaming fields. The only similar situation I can find now is `handle_complain_request`, but it looks already correct.

bmwill (approved):

PR #424, opened 2026-04-09
Rename PendingWithdrawal (WithdrawalTransaction) and persist + flatten WithdrawalStatus

Bridgerz:
## Summary
- Renames `PendingWithdrawal` to `WithdrawalTransaction` with `key, store` abilities and UID for direct object querying
- Splits single `pending_withdrawals` Bag into `withdrawal_txns` (hot ObjectBag) + `confirmed_txns` (cold ObjectBag) for hot/cold separation
- Flattens `WithdrawalStatus` enum to C-style variants (no fields) — redundant data now lives on `WithdrawalRequest.withdrawal_txn_id` and `WithdrawalTransaction.txid`
- Updates all Rust types, scraping, leader, watcher, gRPC, CLI, metrics, proto definitions, and docs

## Test plan
- [x] 71/71 Move unit tests pass
- [x] `cargo check` clean
- [x] `cargo clippy` clean
- [x] Move and Rust formatters pass
- [ ] CI green
- [ ] Devnet deployment test

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #439, opened 2026-04-10
Unify user_requests index on BitcoinState

Bridgerz:
## Summary
Stacked on #424.

Moves the per-user request index out of `DepositRequestQueue` and `WithdrawalRequestQueue` and into a single unified `user_requests: Table<address, Bag>` on `BitcoinState`. Both deposit and withdrawal request IDs now go into the same per-user Bag, giving clients a single place to look up all requests for an address.

- New `index_user_request` / `unindex_user_request` / `has_user_requests` / `user_has_request` helpers on `BitcoinState`
- `WithdrawalRequestQueue.insert_withdrawal` no longer takes `ctx` (indexing moved to entry function)
- `DepositRequestQueue.insert_processed` returns `(request_id, recipient)` so the caller can index
- Indexing calls moved to entry functions in `withdraw.move` and `deposit.move`
- Rust BCS types updated to match

## Test plan
- [x] 71/71 Move tests pass
- [x] `cargo check` and `cargo clippy` clean
- [x] Move and Rust formatters pass
- [ ] CI green

### DIFF
--- a/packages/hashi/sources/btc/bitcoin_state.move
+++ b/packages/hashi/sources/btc/bitcoin_state.move
@@ -8,6 +8,7 @@ use hashi::{
     utxo_pool::UtxoPool,
     withdrawal_queue::WithdrawalRequestQueue
 };
+use sui::{bag::Bag, table::Table};
 
 public struct BitcoinStateKey has copy, drop, store {}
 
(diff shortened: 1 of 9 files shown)


packages/hashi/sources/btc/bitcoin_state.move
```diff
@@ -50,3 +55,46 @@ public(package) fun utxo_pool(self: &BitcoinState): &UtxoPool {
...
+public(package) fun unindex_user_request(
+    self: &mut BitcoinState,
+    user: address,
+    request_id: address,
+) {
+    if (self.user_requests.contains(user)) {
+        let user_bag: &mut Bag = &mut self.user_requests[user];
+        if (user_bag.contains(request_id)) {
+            let _: bool = user_bag.remove(request_id);
+        };
```
bmwill, line 84: maybe add a "if bag empty delete" here?

bmwill (approved):

PR #443, opened 2026-04-13
Add backup and restore of fjall db snapshot

dlukegordon:
## Summary
Adds encrypted backup and restore of the fjall database snapshot plus the node's config files, and proves the round trip end-to-end with a new e2e test that takes a node offline, backs it up, destroys its state, lets the rest of the network rotate several epochs without it, restores from the backup, and verifies it rejoins the committee.
## What's new
- `hashi backup save` produces an age-encrypted `.tar.age` archive containing the CLI config, node config, any referenced key files, and a consistent MVCC snapshot of the fjall database (all four keyspaces). Accepts native x25519 and age plugin recipients (YubiKey etc.).
- `hashi backup restore` decrypts the archive, validates the manifest, and either extracts into a target directory or puts every file back at its original path with `--copy-to-original-paths`. Extraction and DB restoration are both atomic (stage into a sibling temp dir, rename on success) so a failure mid-flight never leaves half-populated state.
- `Database::snapshot_to_path` takes a fjall MVCC snapshot while the source DB stays open for writes. Chunks the copy by a 16 MiB payload budget to bound memory independently of value size. Refuses a pre-existing destination so a typo can't silently merge.
- Symlinks rejected on both archive and restore paths. Restored files written with mode 0o600.
- Locked-DB detection produces a clear "stop the node first" error instead of a fjall internal.
## Testing
- 14 unit tests in `crates/hashi/src/backup.rs` and `crates/hashi/src/cli/commands/backup.rs` cover manifest disambiguation, round-trip DB integrity, refusal to overwrite, inline-PEM vs path classification for node config fields, and the locked-DB error path.
- New e2e test `test_backup_restore_round_trip_and_rejoin` takes node 0 offline, runs `backup save`, deletes its db and config files, forces two rotations without it, runs `backup restore`, restarts node 0, and asserts all 4 nodes agree on the MPC pubkey after one more rotation. Runs in ~100s under nextest.
## Notes for reviewers
- New `backup_age_pubkey` field on `CliConfig` accepts native or plugin recipients via a single serde representation.
- `backup_file_paths()` is the single source of truth for "what must be archived", and bails when `node_config_path` is unset or points at a broken/incomplete node config — so a typo can't produce a silently-incomplete backup.
- Added `tempfile` and `walkdir` to `hashi`'s runtime dependencies (previously only dev), needed for atomic staging during restore.
- Added `age` as a dev-dep on `e2e-tests` for the new test.

### DIFF
--- a/crates/hashi/src/backup.rs
+++ b/crates/hashi/src/backup.rs
@@ -10,7 +10,6 @@
 
 use anyhow::Context;
 use anyhow::Result;
-use std::collections::HashMap;
 use std::collections::HashSet;
 use std::fs;
 use std::fs::File;
@@ -19,6 +18,7 @@ use std::io;
 use std::io::ErrorKind;
 use std::io::Read;
 use std::os::unix::fs::OpenOptionsExt;
+use std::path::Component;
 use std::path::Path;
 use std::path::PathBuf;
 use tracing::info;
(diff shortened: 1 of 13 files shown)


crates/hashi/src/backup.rs
```diff
@@ -0,0 +1,718 @@
...
+}
+
+pub fn encrypt_files_to_age_archive(
+    manifest: &BackupManifest,
+    db_snapshot_dir: &Path,
+    recipient: &dyn age::Recipient,
+    output_path: &Path,
+) -> Result<()> {
+    let output = File::create(output_path)
+        .with_context(|| format!("Failed to create {}", output_path.display()))?;
```
bmwill: Elsewhere we try hard to avoid clobbering existing files but this will silently overwrite an existing file if it already exists, we could use `create_new` instead if we want to  not clobber any existing file.

crates/hashi/src/backup.rs
```diff
@@ -0,0 +1,718 @@
...
+        let mut entry = entry?;
+        let archive_path = entry.path()?.into_owned();
+
+        if archive_path.starts_with(db_prefix) {
+            restore_db_entry(&mut entry, &archive_path, output_dir)?;
+        } else {
+            restore_config_entry(&mut entry, &archive_path, output_dir, &expected_files)?;
+            restored_count += 1;
+        }
+    }
```
bmwill, line 404: We're not doing any explicit checking the the restored DB here, we only validate the config entry count. Do we want to also do any checks for the db entries?

crates/hashi/src/backup.rs
```diff
@@ -0,0 +1,718 @@
...
+        format!(
+            "Failed to copy database snapshot from {} to staging directory {}",
+            source.display(),
+            staging.path().display()
+        )
+    })?;
+
+    // Disarm the TempDir guard before rename so it doesn't try to delete the
+    // path we just moved away.
+    let staging_path = staging.keep();
```
bmwill: I'm not sure i follow what the main reason is for why we need to disarm the drop guard.

crates/hashi/src/backup.rs
```diff
@@ -0,0 +1,718 @@
...
+    fs::rename(&staging_path, dest).map_err(|e| {
```
bmwill, line 679: `fs::rename`'s docs say:

> This will not work if the new name is on a different mount point.

Probably worth a comment or docs around this. I also wonder how portable this is if the TEMP dir is on a different fs mount.

crates/hashi/src/backup.rs
```diff
@@ -0,0 +1,718 @@
...
+    }
+    let mut output_file = create_file_strict(&output_path)?;
+    io::copy(entry, &mut output_file).with_context(|| {
+        format!(
+            "Failed to write restored file contents to {}",
+            output_path.display()
+        )
+    })?;
+    Ok(())
+}
```
bmwill, line 486: Do we want to do a similar defense in depth approach as is done in `extract_dir_name` where we avoid `..` path segments in order to avoid breaking out of the directory we are restoring to?

crates/hashi/src/backup.rs
dlukegordon: Clarified in comments.

crates/hashi/src/backup.rs
dlukegordon: Added

crates/hashi/src/backup.rs
dlukegordon: Clarified in comments.

crates/hashi/src/backup.rs
dlukegordon: Using create_file_strict now

crates/hashi/src/backup.rs
dlukegordon: Now validating db files via the manifest

bmwill (approved):

PR #446, opened 2026-04-13
Allow cancellation of approved withdrawal requests

Bridgerz:
## Summary
- Removes the `ECannotCancelAfterApproval` guard from `cancel_withdrawal()`
- Withdrawal requests can now be cancelled while in `Requested` **or** `Approved` state
- Once the committee commits the request to a `WithdrawalTransaction` (status moves to `Processing`, BTC is burned, request moves to `processed` bag), cancellation is no longer possible — `borrow_request()` will abort since the request is no longer in the `requests` bag

## Why
Previously, approval was treated as the point of no return. But between `Approved` and `Processing` there's a window where the request is still in the active `requests` bag with its full BTC balance intact. The actual irreversible step is `commit_requests()`, which drains the BTC and moves the request to the `processed` bag. Allowing cancellation in the `Approved` state gives users more flexibility without any safety risk.

## Test plan
- [x] `test_approve_then_cancel` — now verifies successful cancellation returns full BTC
- [x] `sui move test -p packages/hashi` — 71/71 pass
- [x] `prettier-move -c` — formatting clean

### DIFF
--- a/packages/hashi/tests/withdraw_tests.move
+++ b/packages/hashi/tests/withdraw_tests.move
@@ -180,7 +180,8 @@ fun test_approve_request_bad_signature() {
 }
 
 #[test]
-#[expected_failure(abort_code = hashi::withdraw::ECannotCancelAfterApproval)]
+/// Cancelling an approved (but not yet processing) request should succeed
+/// and return the full BTC balance to the requester.
 fun test_approve_then_cancel() {
     let epoch = 0u64;
     let ctx = &mut test_utils::new_tx_context(REQUESTER, epoch);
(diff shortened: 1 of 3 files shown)


packages/hashi/tests/withdraw_tests.move
```diff
@@ -180,7 +180,8 @@ fun test_approve_request_bad_signature() {
 }
 
 #[test]
-#[expected_failure(abort_code = hashi::withdraw::ECannotCancelAfterApproval)]
+/// Cancelling an approved (but not yet processing) request should succeed
+/// and return the full BTC balance to the requester.
 fun test_approve_then_cancel() {
```
bmwill, line 185: can we add a test for trying to cancel a request that was entered into the processed state?

Bridgerz:
@bmwill ready for another round

bmwill (approved):

PR #452, opened 2026-04-14
Process deposits on new btc blocks, not new sui checkpoint

dlukegordon:


crates/hashi/src/leader/mod.rs
```diff
@@ -134,19 +132,31 @@ impl LeaderService {
...
-                    }
-
-                    self.process_deposit_requests(checkpoint_timestamp_ms);
                     self.process_unapproved_withdrawal_requests(checkpoint_timestamp_ms);
                     self.process_approved_withdrawal_requests(checkpoint_timestamp_ms);
                     self.process_unsigned_withdrawal_txns();
                     self.process_signed_withdrawal_txns();
                     self.check_delete_proposals(checkpoint_timestamp_ms);
                 }
+                Ok(()) = btc_block_rx.changed() => {
```
bmwill: we should probably check the error status of this like we do above and log something and bail.

dlukegordon: Pushed

bmwill (approved):

PR #453, opened 2026-04-14
[telemetry] Add structured logging with span-based tracing

0xsiddharthks:
Add a new `telemetry` module in hashi-types .

## additional changes:
- tty vs JSON logging
  - using `std::io::IsTerminal` to detect stderr (terminal -> ttl , pipe -> JSON)
  - override possible via `RUST_LOG_JSON`
- service identification
  - no longer requiring explicit service-name in the log body
  - using two mechanisms for service identification:
    - loki stream labels (in k8s)
    - `target` field on every event. (tracing subscriber emits the module path in the field)

## Also add `#[tracing::instrument]` at the main unit-of-work boundaries:
- gRPC handlers in `bridge_service`
- Deposit / withdrawal validation paths
- Every leader task (deposit, withdrawal approval / commitment / signing / confirmation fan-outs)
- Bitcoin and Sui indexers
- MPC run loops, `handle_reconfig`, and `mpc::signing::sign`
- Every `execute_*` in `sui_tx_executor` — the generic `execute` records `sui_digest` on success so one span tree covers handler → validator → Sui submission

## deployment changes:
MystenLabs/sui-operations#7626


crates/hashi-telemetry/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "hashi-telemetry"
```
bmwill: I don't necessarily think this is needed as a separate crate vs sticking this in a module in hashi-types

0xsiddharthks: Yeah, makes sense. updated as a module.

crates/hashi-types/src/telemetry.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        TelemetryGuard { _private: () }
+    }
+}
+
+#[must_use = "dropping the guard immediately will lose buffered log output"]
+pub struct TelemetryGuard {
+    _private: (),
+}
```
bmwill: Why do we have this guard? I don't think that its actually doing what this says it is doing.

crates/e2e-tests/src/main.rs
```diff
@@ -279,13 +279,10 @@ async fn cmd_start(
...
-        .with_env_filter(
-            tracing_subscriber::EnvFilter::builder()
-                .with_default_directive(default_level.into())
-                .from_env_lossy(),
-        )
+    let _guard = hashi_types::telemetry::TelemetryConfig::new()
+        .with_default_level(default_level)
         .with_target(false)
+        .with_env()
         .init();
```
bmwill, line 289: I think this makes more sense to go at the top of the `main` function in this file rather than here right?

crates/hashi-types/src/telemetry.rs
```diff
@@ -0,0 +1,100 @@
...
+        } else {
+            let fmt_layer = tracing_subscriber::fmt::layer()
+                .with_file(self.file_line)
+                .with_line_number(self.file_line)
+                .with_target(self.target)
+                .with_ansi(stderr().is_terminal())
+                .with_filter(env_filter);
+
+            tracing_subscriber::registry().with(fmt_layer).init();
+        }
```
bmwill, line 100: One last thing, that was a pre-existing issue, we need to explicitly configure the Formatter to output to stderr as it for some reason defaults to stdout

bmwill (approved):

PR #454, opened 2026-04-14
[metrics] Add Bitcoin confirmation-count gauges for deposits

0xsiddharthks:
Add a new metric: `hashi_deposit_request_confirmations` that buckets deposit transaction into the buckets: `not_found | mempool | 0 | 1 | 2 | 3 | 4 | 5 | 6_plus`.

No extra RPC call done. just caching the block heights per active transaction as the leader is checking each transaction.

Adding proper unit tests for the `deposit_observation_cache` in #460 
update metrics docs : MystenLabs/sui-operations#7645

### DIFF
--- a/crates/hashi/src/btc_monitor/monitor.rs
+++ b/crates/hashi/src/btc_monitor/monitor.rs
@@ -1,6 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


bmwill (approved):

bmwill:
This is great. I wonder if it would be worthwhile to do something similarish with UTXOs in our pool. That is continually monitor the UTXOs in the pool (specifically the change ones where we do a lookup already on the txn construction path for how many confirmations they have). This would maybe save some queries/latency as well as give us a way to monitor utxos and if they get spent by a txn the system doesn't know about we can raise an alarm.

PR #455, opened 2026-04-15
Make `f` configurable

zhouwfang:
#368

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -3632,9 +3632,11 @@ where
     F: FnOnce() -> T + Send + 'static,
     T: Send + 'static,
 {
-    tokio::task::spawn_blocking(f)
-        .await
-        .expect("spawn_blocking task panicked")
+    match tokio::task::spawn_blocking(f).await {
+        Ok(v) => v,
+        Err(e) if e.is_cancelled() => std::future::pending().await,
+        Err(e) => std::panic::resume_unwind(e.into_panic()),
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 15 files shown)


zhouwfang:
~~Pending on further `fastcrypto` changes.~~

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -3606,8 +3607,19 @@ fn build_reduced_nodes(
     let total_weight: u16 = nodes_vec.iter().map(|n| n.weight).sum();
     let threshold =
         (total_weight as u32 * threshold_in_basis_points as u32).div_ceil(MAX_BASIS_POINTS) as u16;
-    Nodes::new_reduced(nodes_vec, threshold, weight_reduction_allowed_delta, 1)
-        .map_err(|e| MpcError::CryptoError(e.to_string()))
+    let max_faulty =
+        (total_weight as u32 * max_faulty_in_basis_points as u32).div_ceil(MAX_BASIS_POINTS) as u16;
```
benr-ml, line 3611: isn't this always 10000? when is that different, only in tests?

benr-ml (approved):

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Yes, only tests see a different `total_weight`, because `test_weight_divisor` (asserted None on mainnet/testnet) scales each member's weight down before the sum.

bmwill: total_weight is going to be capped at 10_000 units maximum. But since validator participation in hashi will not be 100%, its likely that the total number of voting units will be less than 10_000 in production settings.

zhouwfang: PTAL at #471, in case `total_weight` is less than 100 on mainnet.

PR #456, opened 2026-04-15
Add upgrade e2e test and fix PackageUpgradedEvent parsing

Bridgerz:
## Summary
- Adds an e2e test that exercises the full governance-gated upgrade lifecycle and verifies cascading effects
- **Fixes a bug**: `PackageUpgradedEvent` was missing from `HashiEvent::try_parse()`, so validators never learned about package upgrades via the event stream — only on full rescrape

## Bug fix
`PackageUpgradedEvent` had a struct definition, `MoveType` impl, and was handled in the watcher's event loop (`watcher.rs:228`), but was never added to the `try_parse()` match arms in `move_types/mod.rs`. The event was silently dropped at the catch-all `_ => return Ok(None)`. After an upgrade, validators would not update their package version map until a subscription reconnect triggered a full rescrape.

## Test details
The test programmatically patches the deployed package at test time (bumps `PACKAGE_VERSION`, injects a canary module), then verifies:

1. **Watcher picks up new package** — all 4 nodes detect the new version via `PackageUpgradedEvent`
2. **Package ID routing** — `OnchainState.package_id()` returns the new package on all nodes
3. **Validator deposit confirmation** — full deposit flow (BTC send → mine → request → validator auto-confirm) works against the upgraded package
4. **New code reachable** — v2-only canary module is callable
5. **Version gating** — disabling v1 rejects old entry points with `EVersionDisabled`

## Files changed
- `crates/hashi-types/src/move_types/mod.rs` — add `PackageUpgradedEvent` to `try_parse()`
- `crates/e2e-tests/src/upgrade_flow.rs` — upgrade helpers (prepare, build, propose/vote/execute, disable)
- `crates/e2e-tests/src/upgrade_tests.rs` — the e2e test
- `crates/e2e-tests/src/lib.rs` — module declarations + `dir()` accessor

## Test plan
- [x] `cargo test -p e2e-tests -- test_upgrade_v1_to_v2 --ignored` — passes in ~44s
- [x] `cargo check -p hashi` — compiles clean

### DIFF
--- a/crates/e2e-tests/src/upgrade_flow.rs
+++ b/crates/e2e-tests/src/upgrade_flow.rs
@@ -10,8 +10,13 @@
 use anyhow::Result;
 use hashi::cli::client::CreateProposalParams;
 use hashi::cli::client::build_create_proposal_transaction;
+use hashi::cli::client::build_vote_transaction;
+use hashi::cli::upgrade::build_execute_proposal_transaction;
+use hashi::cli::upgrade::build_upgrade_execution_transaction;
+use hashi::cli::upgrade::build_upgrade_package;
+use hashi::cli::upgrade::extract_new_package_id_from_response;
+use hashi::cli::upgrade::extract_proposal_id_from_response;
 use hashi::config::HashiIds;
-use hashi::sui_tx_executor::SUI_CLOCK_OBJECT_ID;
 use hashi::sui_tx_executor::SuiTxExecutor;
 use std::path::Path;
 use std::path::PathBuf;
@@ -19,9 +24,6 @@ use sui_sdk_types::Address;
 use sui_sdk_types::Identifier;
 use sui_sdk_types::StructTag;
 use sui_sdk_types::TypeTag;
-use sui_transaction_builder::Function;
-use sui_transaction_builder::ObjectInput;
-use sui_transaction_builder::TransactionBuilder;
 
 use crate::TestNetworks;
 use crate::sui_network::sui_binary;
@@ -94,57 +96,6 @@ pub fn prepare_upgrade_package(test_dir: &Path, original_package_id: Address) ->
     Ok(dst)
 }
 
-/// Build the upgraded package and return the compiled modules + digest.
-pub fn build_upgrade_package(
-    package_path: &Path,
-    client_config: Option<&Path>,
-) -> Result<(sui_sdk_types::Publish, Vec<u8>)> {
-    let mut cmd = std::process::Command::new(sui_binary());
-    cmd.arg("move");
-
-    if let Some(config) = client_config {
-        cmd.arg("--client.config").arg(config);
-    }
-
-    cmd.arg("-p")
-        .arg(package_path)
-        .arg("build")
-        .arg("-e")
-        .arg("testnet")
-        .arg("--dump-bytecode-as-base64");
-
-    let output = cmd.output()?;
-    anyhow::ensure!(
-        output.status.success(),
-        "sui move build failed:\nstdout: {}\nstderr: {}",
-        output.stdout.escape_ascii(),
-        output.stderr.escape_ascii()
-    );
-
-    #[derive(serde::Deserialize)]
-    struct MoveBuildOutput {
-        modules: Vec<String>,
-        dependencies: Vec<Address>,
-        digest: Vec<u8>,
-    }
-
-    let build_output: MoveBuildOutput = serde_json::from_slice(&output.stdout)?;
-    let digest = build_output.digest.clone();
-    let modules = build_output
-        .modules
-        .into_iter()
-        .map(|b64| <base64ct::Base64 as base64ct::Encoding>::decode_vec(&b64))
-        .collect::<Result<Vec<_>, _>>()?;
-
-    Ok((
-        sui_sdk_types::Publish {
-            modules,
-            dependencies: build_output.dependencies,
-        },
-        digest,
-    ))
-}
-
 /// Run the full upgrade lifecycle: prepare → build → propose → vote → execute+publish+finalize.
 ///
 /// Returns the new package ID on success.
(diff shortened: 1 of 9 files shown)


crates/e2e-tests/src/upgrade_tests.rs
```diff
@@ -0,0 +1,362 @@
...
+        Ok(hbtc_recipient)
+    }
+
+    /// Test the full upgrade lifecycle, exercising real cascading effects.
+    ///
+    /// 1. Watcher picks up new package — PackageUpgradedEvent updates OnchainState
+    /// 2. Validators confirm deposits post-upgrade — leader routes calls correctly
+    /// 3. Package ID routing — OnchainState.package_id() returns the new package
+    #[tokio::test]
+    #[ignore = "requires localnet (run with --ignored)"]
```
bmwill: what is missing from being able to run this in CI?

crates/e2e-tests/src/upgrade_tests.rs
```diff
@@ -0,0 +1,362 @@
...
+            .await?
+            .into_inner();
+        Ok(response.balance().balance_opt().unwrap_or(0))
+    }
+
+    async fn wait_for_deposit_confirmation(
+        sui_client: &mut sui_rpc::Client,
+        request_id: Address,
+        timeout: Duration,
+    ) -> Result<()> {
```
bmwill: These functions should already exist elsewhere in the e2e-tests, maybe they just need to be factored out into a common place to that they can be reused?

crates/e2e-tests/src/upgrade_tests.rs
```diff
@@ -0,0 +1,362 @@
...
+
+//! End-to-end test for the package upgrade lifecycle.
+//!
+//! Exercises real cascading effects of upgrading the hashi package:
+//! - Rust watcher picks up the new package version via PackageUpgradedEvent
+//! - Validators auto-confirm deposits against the upgraded package
+//! - Package ID routing updates correctly in OnchainState
+
+#[cfg(test)]
+mod tests {
```
bmwill: This file is probably better placed in `crates/e2e-tests/tests/upgrade_tests.rs` then you shouldn't even need the `cfg(test)` dance.

crates/e2e-tests/src/upgrade_flow.rs
```diff
@@ -0,0 +1,412 @@
...
+        String::from_utf8_lossy(&output.stderr)
+    );
+
+    // Patch config.move: bump PACKAGE_VERSION from 1 to 2
+    let config_path = dst.join("sources/core/config/config.move");
+    let config_src = std::fs::read_to_string(&config_path)?;
+    let patched = config_src.replace(
+        "const PACKAGE_VERSION: u64 = 1;",
+        "const PACKAGE_VERSION: u64 = 2;",
+    );
```
bmwill, line 64: nice

crates/e2e-tests/src/upgrade_tests.rs
```diff
@@ -0,0 +1,362 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! End-to-end test for the package upgrade lifecycle.
+//!
+//! Exercises real cascading effects of upgrading the hashi package:
+//! - Rust watcher picks up the new package version via PackageUpgradedEvent
+//! - Validators auto-confirm deposits against the upgraded package
```
bmwill: I'm actually semi-surprised that our system is already smart enough to do this as a part of our txn construction.

crates/e2e-tests/src/upgrade_flow.rs
```diff
@@ -0,0 +1,412 @@
...
+use crate::TestNetworks;
+use crate::sui_network::sui_binary;
+
+/// Prepare an upgrade package by copying the deployed source and patching it.
+///
+/// 1. Copies `<test_dir>/packages/hashi` to `<test_dir>/packages/hashi-upgrade`
+/// 2. Bumps `PACKAGE_VERSION` from 1 to 2 in `config.move`
+/// 3. Sets `published-at` in `Move.toml` to the original package ID
+///
+/// Returns the path to the patched package directory.
```
bmwill, line 37: Wonder if it would make sense to factor some, not all, of this logic into a reusable functions that we can then leverage in the cli to do upgrades in the future (create, vote, execute proposals). Specifically to do some explicit pre-flight checks (check that VERSION is exactly what we expect it to be `+1` of the latest published version, etc.

Thoughts?

crates/e2e-tests/src/upgrade_tests.rs
Bridgerz: nothing. good catch

crates/e2e-tests/src/upgrade_flow.rs
Bridgerz: not a bad idea. I'll go ahead and extract the useful functions to hashi and then in my coming cli update I'll add these checks.

Bridgerz:
@bmwill this is ready for another round!

crates/hashi/src/cli/upgrade.rs
```diff
@@ -0,0 +1,236 @@
...
+/// `client_config` is passed as `--client.config` when supplied, otherwise the
+/// `sui` binary's default client config is used.
+///
+/// Returns the compiled `Publish` (modules + dependencies) plus the package
+/// digest — the latter is what goes into the `Upgrade` proposal.
+pub fn build_upgrade_package(
+    sui_binary: &Path,
+    package_path: &Path,
+    client_config: Option<&Path>,
+) -> Result<(Publish, Vec<u8>)> {
```
bmwill, line 48: I may be not seeing it, but can we add explicit checks for a few things? Specifically ensuring that the upgraded package version is exactly +1 from the last published version?

Bridgerz: will add it to #458

bmwill (approved):

PR #458, opened 2026-04-16
CLI: governance improvements

Bridgerz:
## Summary

A batch of CLI UX/ergonomics improvements for the governance (proposal) flow,
plus one correctness fix for on-chain state scraping and a small localnet
helper to make this stuff actually drivable against a local dev network.

Stacked on top of #456 — set base back to `main` once that lands.

### Commits

1. **Fix CLI on-chain state init for Bag-backed dynamic fields.** The
   scraper was reading `child_object.contents` for every dynamic field,
   but that's only populated for `ObjectBag` (DynamicObject) kinds. For
   plain `Bag` entries (proposals), the real payload is on the
   `DynamicField` itself via `value_type` + `value`. Without this,
   every governance CLI command errored with `unable to parse type ""`
   once the proposals bag had any entries.
2. **Friendlier prompt + print proposal ID on create.** `(y/N)` →
   `Press enter to X, or Ctrl+C to cancel`. After `proposal create ...`,
   print the new `Proposal ID: 0x…` so it can be copy/pasted into the
   follow-up `vote` / `view` / `execute` commands.
3. **Show votes + quorum progress in `proposal view`.** One extra
   `list_dynamic_fields` call pulls the full `Proposal<T>` envelope and
   displays creator, voter addresses, voted-vs-threshold weight, and
   whether quorum is reached. Scoped to the CLI — no new fields on
   `onchain::types::Proposal` or changes to the watcher.
4. **`--execute` flag on `proposal vote`.** Casts the vote and, if
   that vote pushes the proposal over quorum, chains an immediate
   `proposal::<module>::execute` in the same invocation. Skips silently
   (with an info message) when quorum isn't reached. Upgrade proposals
   are rejected — they need the dedicated upgrade flow.
5. **localnet: dump validator operator keys to disk.** Writes each
   validator's `operator_private_key` to
   `<data-dir>/validators/validator_<N>.pem` (0600) at `hashi-localnet
   start` time. Without this, the keys only live in-memory on the
   running validator processes, so there's no way to point the CLI at a
   committee-member key for `proposal create` / `vote` / `execute` on
   localnet.

## Test plan

Tested end-to-end against a 4-validator localnet:

- [x] `proposal create update-config bitcoin_confirmation_threshold u64:3`
      prints the new proposal ID and uses the press-enter prompt.
- [x] `proposal view <id>` shows `Creator`, `Votes: N voter(s) — <w> of total
      weight <total>`, `Threshold: <bps> (<pct>%)`, and the voter list.
- [x] `proposal vote <id> --execute` from the second validator prints
      "Quorum not reached yet (…); skipping --execute."
- [x] `proposal vote <id> --execute` from the third validator prints
      "Quorum reached (…); executing…" and submits both the vote and the
      execute transactions.
- [x] On-chain `bitcoin_confirmation_threshold` updated to `3` after
      the auto-execute.

### DIFF
--- a/crates/e2e-tests/src/main.rs
+++ b/crates/e2e-tests/src/main.rs
@@ -303,22 +303,22 @@ async fn cmd_start(
         .to_pem()
         .context("Failed to serialize funded key as PEM")?;
     std::fs::create_dir_all(data_dir)?;
-    {
-        use std::io::Write;
-        use std::os::unix::fs::OpenOptionsExt;
-        let mut file = std::fs::OpenOptions::new()
-            .write(true)
-            .create(true)
-            .truncate(true)
-            .mode(0o600)
-            .open(&funded_key_path)
-            .with_context(|| {
-                format!(
-                    "Failed to write funded key to {}",
-                    funded_key_path.display()
-                )
-            })?;
-        file.write_all(pem.as_bytes())?;
+    write_pem_to_disk(&funded_key_path, &pem)?;
+
+    // Write each validator's operator key so CLI governance commands can
+    // use them. The `hashi` CLI needs a committee-member key to propose /
+    // vote / execute; these aren't persisted anywhere else for localnet.
+    let validators_dir = data_dir.join("validators");
+    std::fs::create_dir_all(&validators_dir)?;
+    for (i, node) in test_networks.hashi_network().nodes().iter().enumerate() {
+        let operator_pem = node
+            .hashi()
+            .config
+            .operator_private_key
+            .as_ref()
+            .context("validator has no operator_private_key")?;
+        let path = validators_dir.join(format!("validator_{i}.pem"));
+        write_pem_to_disk(&path, operator_pem)?;
     }
 
     let state = LocalnetState {
@@ -805,6 +805,22 @@ fn cli_config_path(data_dir: &Path) -> std::path::PathBuf {
     data_dir.join("hashi-cli.toml")
 }
 
+/// Write a PEM-encoded key to disk with restrictive permissions (0600 on unix).
+fn write_pem_to_disk(path: &Path, pem: &str) -> Result<()> {
+    use std::io::Write;
+    use std::os::unix::fs::OpenOptionsExt;
+    let mut file = std::fs::OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .mode(0o600)
+        .open(path)
+        .with_context(|| format!("Failed to open {} for writing", path.display()))?;
+    file.write_all(pem.as_bytes())
+        .with_context(|| format!("Failed to write PEM to {}", path.display()))?;
+    Ok(())
+}
+
 /// Write a `hashi-cli.toml` config file that the main `hashi` CLI can read.
 fn write_cli_config(data_dir: &Path, state: &LocalnetState) -> Result<()> {
     let config = hashi::cli::config::CliConfig {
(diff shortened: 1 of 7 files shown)


crates/e2e-tests/src/main.rs
```diff
@@ -303,22 +303,22 @@ async fn cmd_start(
...
-                )
-            })?;
-        file.write_all(pem.as_bytes())?;
+    write_pem_to_disk(&funded_key_path, &pem)?;
+
+    // Write each validator's operator key so CLI governance commands can
+    // use them. The `hashi` CLI needs a committee-member key to propose /
+    // vote / execute; these aren't persisted anywhere else for localnet.
+    let validators_dir = data_dir.join("validators");
+    std::fs::create_dir_all(&validators_dir)?;
```
bmwill, line 312: Do we not persist the configs to disk?

crates/hashi/src/cli/client.rs
```diff
@@ -190,6 +199,93 @@ impl HashiClient {
...
+
+    /// Live-fetch the full on-chain `Proposal<T>` (votes + quorum threshold +
+    /// metadata) for a specific proposal via one `list_dynamic_fields` call on
+    /// the proposals bag. Separate from the cached `fetch_proposal` because
+    /// validators don't need these fields in their in-memory state.
+    pub async fn fetch_proposal_details(
+        &self,
+        proposal_id: Address,
+        proposal_type: &crate::onchain::types::ProposalType,
+    ) -> Result<ProposalDetails> {
```
bmwill: What if there are multiple proposals of type T?

crates/hashi/src/onchain/mod.rs
```diff
@@ -707,6 +707,19 @@ pub(crate) async fn scrape_hashi_config(
     Ok(convert_move_config(config))
 }
 
+/// Return the child object's type as a string, preferring the dedicated
+/// `object_type` field (always set by Sui indexing) and falling back to the
+/// BCS `contents.name` (which some fullnodes leave empty in dynamic-field
+/// list responses).
```
bmwill: This is likely just because the field mask needs to be properly set right?

crates/hashi/src/onchain/mod.rs
```diff
@@ -1209,29 +1232,40 @@ async fn scrape_proposals(
 ) -> Result<types::Proposals> {
     let mut proposals: BTreeMap<Address, types::Proposal> = BTreeMap::new();
 
+    // Proposals live in a `0x2::bag::Bag` (dynamic FIELD, value stored inline),
+    // so we read `value_type` + `value` off the `DynamicField` proto itself —
+    // NOT `child_object` (that's only populated for DynamicObject kinds).
```
bmwill, line 1224: Which i think your next PR is changing back right?

bmwill (approved):

crates/e2e-tests/src/main.rs
Bridgerz: its nice for testing governance flows.

crates/hashi/src/cli/client.rs
Bridgerz: good catch. type is now derived

crates/hashi/src/onchain/mod.rs
Bridgerz: now requests object_type so the primary path always works.

crates/hashi/src/onchain/mod.rs
Bridgerz: correct! just wanted to split the changes up a bit

PR #460, opened 2026-04-16
[metrics] Add unit tests for confirmation-count bucket logic

0xsiddharthks:
Unit tests for the bucket logic in `rebuild_confirmation_metrics_inner` added in #454.

### DIFF
--- a/crates/hashi/src/btc_monitor/monitor.rs
+++ b/crates/hashi/src/btc_monitor/monitor.rs
@@ -1062,3 +1062,148 @@ enum MonitorMessage {
     // Drop a deposit from `deposit_observation_cache`
     ForgetDeposit(bitcoin::OutPoint),
 }
+
+#[cfg(test)]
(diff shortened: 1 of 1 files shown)


crates/hashi/src/btc_monitor/monitor.rs
```diff
@@ -1062,3 +1062,146 @@ enum MonitorMessage {
...
+
+        rebuild_confirmation_metrics_inner(&mut cache, 100, &metrics);
+
+        for label in crate::metrics::CONFIRMATION_STATUS_LABELS {
+            assert_eq!(bucket(&metrics, label), 0, "label {label} should be zero");
+        }
+    }
+
+    #[test]
+    fn observation_to_bucket_mapping() {
```
dlukegordon, line 1111: Should we cover the "0" bucket as well?

dlukegordon (approved):

PR #449, opened 2026-04-13
[1/n][guardian-integration] guardian local and dev deployment infra setup

0xsiddharthks:
## Summary

- Plain-Kubernetes (non-enclave) Containerfile for hashi-guardian + `grpc.health.v1.Health` registration in the binary
- New `non-enclave-dev` cargo feature that stubs the AWS Nitro NSM attestation call so the guardian runs outside an enclave; enabled by the K8s image
- `docker/hashi-guardian-k8s/compose.yaml` — MinIO-based local dev stack with an Object-Lock bucket pre-created;
- `crates/hashi-guardian/examples/bootstrap_operator_init.rs` — example script that drives `OperatorInit` against the local stack

The service is already running in the dev cluster: 
- deployment: https://app.pulumi.com/mysten/hashi-guardian/dev/updates/3
- PR: MystenLabs/sui-operations#7617

## Run locally

```bash
# Terminal 1 — MinIO
docker compose -f docker/hashi-guardian-k8s/compose.yaml up -d

# Terminal 2 — guardian
AWS_ENDPOINT_URL_S3=http://localhost:9000 \
cargo run --release -p hashi-guardian --features non-enclave-dev

# Terminal 3 — bootstrap operator_init
GUARDIAN_ENDPOINT=http://localhost:3000 \
AWS_S3_BUCKET=hashi-guardian-dev \
AWS_REGION=us-east-1 \
AWS_ACCESS_KEY_ID=minioadmin \
AWS_SECRET_ACCESS_KEY=minioadmin \
BITCOIN_NETWORK=signet \
cargo run --release -p hashi-guardian --features non-enclave-dev --example bootstrap_operator_init

# Teardown:
docker compose -f docker/hashi-guardian-k8s/compose.yaml down -v
```

PR #463, opened 2026-04-17
[4/n][guardian-integration] wid-keyed idempotency cache

0xsiddharthks:
> Stacked on #423 → #466 → #449.

## Summary

Retries of the same withdrawal (leader restart, leader rotation, lost response, or a seq-mismatch retry on the hashi side) previously debited the bucket once per retry — even though `wid` is deterministic and the actual on-chain withdrawal only happens once.

This PR adds an LRU cache of signed responses keyed by `wid` on `EnclaveState` (cap 1024). `standard_withdrawal` consults the cache before touching committee-sig verification, the limiter, or BTC signing; a hit returns the previously signed response untouched. Entries are inserted only after a successful S3 log commit, so the cache and bucket always agree.

## Why wid

`wid` is already computed as `u64::from_le_bytes(blake2b256(bcs(request_ids))[..8])` deterministically on the hashi side. The guardian re-derives the same value from the on-chain request set, so a retry from any leader at any epoch collides to the same wid.

## What this unlocks

- The hard-reserve seq-mismatch retry loop added in #423 is now safe to fire repeatedly without double-debit.
- PR-4's soft reserve (pre-MPC `SoftReserveWithdrawal`) can rely on wid idempotency to dedupe soft-reservation attempts across nodes.

## Tests

- `test_standard_withdrawal_wid_cache_is_idempotent` — same wid twice returns an identical signed response; bucket reflects exactly one debit.
- `test_standard_withdrawal_failures_not_cached` — failed withdrawals are NOT cached; the next attempt runs the live path.

## Follow-ups (not in this PR)

- S3-replay rehydration of the cache on guardian restart (PR-5 in the stack).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2893,6 +2893,7 @@ dependencies = [
  "hashi-types",
  "hpke",
  "k256",
+ "lru",
  "rand 0.8.5",
  "serde",
  "serde_bytes",
--- a/crates/hashi-guardian/Cargo.toml
+++ b/crates/hashi-guardian/Cargo.toml
@@ -13,6 +13,7 @@ tracing.workspace = true
 tracing-subscriber.workspace = true
 tonic.workspace = true
 tonic-health.workspace = true
+lru = "0.16"
 hashi-types = { path = "../hashi-types" }
 
 # Crypto dependencies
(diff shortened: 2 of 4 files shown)


0xsiddharthks:
Deferred alongside #465. The wid-keyed response cache is a real correctness improvement — without it, cross-tick retries of the same withdrawal double-debit the bucket. That is a rate-limiter accuracy concern, not a safety one: BTC withdrawal remains exactly-once (enforced by Sui Move `.withdrawal_txns` state + MPC signing independent of the guardian).

For the MVP signet/devnet landing we are shipping #423 with both soft + hard guardian touchpoints and accepting the retry double-debit as a known limitation. This PR will be re-applied before we care about strict bucket accuracy (pre-scale / pre-mainnet).

See `.claude/plans/golden-finding-castle.md` for the full investigation + decision rationale. Branch `siddharth/guardian-wid-cache` preserved locally for the re-application.

PR #464, opened 2026-04-17
[5/n][guardian-integration] soft-reserve RPC + pre-commit integration

0xsiddharthks:
> Stacked on #463 → #423 → #466 → #449.

## Summary

PR-2 (hard reserve post-MPC) means an over-limit withdrawal only surfaces AFTER we've done committee fan-out + MPC signing — minutes of wasted work per leader iteration. This PR adds a lightweight pre-commit soft reserve so over-limit withdrawals abort cheaply.

## Guardian changes (`hashi-types` + `hashi-guardian`)

- `RateLimiter` gains `pending_reserves: HashMap<wid, PendingReserve>` and a `soft_reserve(wid, timestamp, amount, now)` method. Idempotent on wid (repeat probes return the existing reservation with a refreshed TTL). Capacity check: `refill_capacity(timestamp) - Σ pending.amount_sats ≥ amount`.
- `consume` now takes `wid` and drops any matching pending entry — a hard reserve atomically converts a soft reserve.
- `expire_pending(now)` drops stale entries; `SOFT_RESERVE_TTL_SECS = 5 min`. A 1s sweep task runs in \`main.rs\`.
- New `SoftReserveWithdrawal` RPC. **No committee signature required** for soft reserves — the TTL bounds DoS blast radius and wid idempotency handles retries.

## Hashi changes

- `GuardianClient::soft_reserve_withdrawal` wrapper.
- `leader::soft_reserve_withdrawal_through_guardian` probes the guardian after coin selection and before the committee BLS fan-out for commit. Rate-limited / unavailable → abort iteration; next tick retries with the same wid (guardian returns the same reservation).
- `compute_withdrawal_wid` extracted into \`withdrawals.rs\` so both touchpoints derive the same deterministic id.

## Move package

No changes.

## Tests

- `test_soft_reserve_is_idempotent_on_wid` — repeat probes return existing reservation, TTL refreshed
- `test_soft_reserve_rejects_over_commitment_across_wids` — pending amounts subtract from capacity
- `test_soft_reserve_enforces_monotonic_timestamp` — out-of-order timestamps rejected
- `test_expire_pending_drops_stale_reservations` — TTL sweep drops past-due entries
- `test_consume_removes_matching_soft_reserve` — hard reserve atomically drops pending
- `test_soft_reserve_leaves_room_for_hard_reserve_of_same_wid` — end-to-end: soft-reserve capacity, then hard-reserve succeeds
[...]

PR #465, opened 2026-04-17
[6/n][guardian-integration] restart safety: rehydrate limiter + wid cache from S3

0xsiddharthks:
> Stacked on #464 → #463 → #423 → #466 → #449.

## Summary

On a guardian restart the new session generated fresh ephemeral keys and started with a mock `LimiterState` (bucket at max capacity, `next_seq = 0`) plus an empty wid-keyed response cache (PR-3). That combination meant retries of prior-session withdrawals would either be rejected outright (seq mismatch) or double-debit the bucket. This PR closes both gaps.

## Log schema

- `WithdrawalLogMessage::Success` carries an optional `limiter_state_post` — the full `LimiterState` snapshot after the withdrawal was applied. \`#[serde(default)]\` keeps the format backward compatible with logs written before this field existed.
- Withdrawal log Object Lock retention bumped from 5 min to 1 hour so the KP can read prior-session tails across realistic restart windows without racing retention expiry.

## KP-side rehydration (`crates/hashi-monitor/src/kp`)

- \`heartbeat_checks::collect_recent_sessions\` exposes the per-session heartbeat summary so rehydration can enumerate prior sessions alongside the new live one.
- \`kp::rehydrate_limiter_state\` scans prior-session withdrawal logs in the recent window and picks the highest-seq \`limiter_state_post\` snapshot. Falls back to a fresh max-capacity state when no prior logs exist (first-ever provisioning).

## Guardian-side rehydration (`crates/hashi-guardian/src/init`)

- After \`finalize_init\` and before marking \`provisioner_init_logging_complete\`, the guardian:
  1. Lists prior-session heartbeats to enumerate prior session ids.
  2. Resolves each prior session's attested Ed25519 signing pubkey from \`init/{session_id}-oi-attestation-unsigned.json\`.
  3. Reads the prior session's withdrawal success logs, verifying each against that session's pubkey.
  4. Re-signs the unsigned \`response\` with the CURRENT session's key.
  5. Caches the re-signed response by wid.
- Reads are tolerant of missing directories; failures log a warning and continue rather than blocking withdrawal serving. Worst case we fall back to the pre-PR-5 behavior of an empty cache.

## Why re-sign on rehydration

The log's \`response\` field is the unsigned \`StandardWithdrawalResponse\` (BTC Schnorr sigs for each input). Those sigs are deterministic in the enclave BTC key, which is shared across sessions. Only the Ed25519 envelope (timestamp + signature) is session-specific. Re-signing with the new session's key means clients that \`get_guardian_info\`-fetched the new pubkey verify cached retries cleanly.

## Tests

All existing unit tests still pass (hashi 240/240, hashi-guardian 17/17, hashi-types 47/47, hashi-monitor 11/11 excluding the pre-existing \`lookup_btc_confirmation_with_local_regtest\` which requires a local regtest node). Rehydration integration tests require real S3 (or MinIO) so they land with the guardian e2e harness (task #10).

## Follow-ups

- Hashi leader should re-fetch \`guardian_signing_pubkey\` on verification failure so it picks up a restarted guardian's new key without a hashi restart. Out of scope for this PR.
- E2E harness that exercises a full guardian restart end-to-end — captured as task #10 / PR-2b.

### DIFF
--- a/crates/hashi-guardian/src/init.rs
+++ b/crates/hashi-guardian/src/init.rs
@@ -8,14 +8,25 @@ use hashi_types::guardian::crypto::combine_shares;
 use hashi_types::guardian::crypto::commit_share;
 use hashi_types::guardian::crypto::decrypt_share;
 use hashi_types::guardian::crypto::Share;
+use hashi_types::guardian::s3_utils::S3HourScopedDirectory;
+use hashi_types::guardian::time_utils::now_timestamp_secs;
 use hashi_types::guardian::InitLogMessage::OIAttestationUnsigned;
 use hashi_types::guardian::InitLogMessage::OIGuardianInfo;
 use hashi_types::guardian::InitLogMessage::PIEnclaveFullyInitialized;
 use hashi_types::guardian::InitLogMessage::PISuccess;
 use hashi_types::guardian::*;
+use std::collections::HashMap;
 use std::sync::Arc;
 use tracing::info;
+use tracing::warn;
 use GuardianError::*;
+
+/// How many hour-scoped directories back we scan when rehydrating the
+/// wid-keyed response cache from prior-session withdrawal logs. Matches
+/// the Object Lock retention window for withdrawal logs (see
+/// `S3_OBJECT_LOCK_DURATION_WITHDRAW`) so we never try to read a
+/// directory whose entries have expired out from under us.
+const REHYDRATE_LOG_HOURS: u64 = 2;
 
 /// Receives S3 API keys & share commitments.
 /// Returns an error for malformed requests / dup call & panics for the rest.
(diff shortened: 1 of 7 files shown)


0xsiddharthks:
Deferred alongside #463. The main restart-safety value is wid-cache rehydration, which is meaningless without the cache itself. `limiter_state_post` in logs + KP-side `rehydrate_limiter_state` + Object Lock retention bump (5 min → 1 hour) stand on their own, but are not required for initial signet/devnet bring-up — a restarted dev guardian resetting its bucket is acceptable at this stage.

To re-apply: restore the wid-cache first (re-open #463), then re-apply this branch. See `.claude/plans/golden-finding-castle.md`. Branch `siddharth/guardian-restart-safety` preserved locally.

PR #466, opened 2026-04-17
[2/n][guardian-integration] in-process e2e harness

0xsiddharthks:
## Summary

- lift `Enclave` (+ `EnclaveConfig` / `EnclaveState` / `Scratchpad`) from `main.rs` into a new `enclave.rs` lib module
  — @mskd12 this is just a simple move. no changes made to the code.
    - the only change is that `enclave_btc_keypair` and `hashi_btc_master_pubkey` went from private to `pub(crate)`. 
    - This is required because `init.rs::tests` reads these fields directly via `enclave.config.<field>.get()`
  - moving these to `enclave.rs` allows the e2e-tests to construct and derive an enclave in-process
- New `test-utils` feature on `hashi-guardian` exposing constructors for partially/fully-initialized enclaves.
- add new guardian client for hashi

## Tests harness changes
- `e2e-tests::GuardianHarness` runs an in-process guardian over real gRPC.
- `TestNetworksBuilder::with_guardian()` orchestrates start → DKG → finalize.
- `test_bitcoin_withdrawal_with_guardian_e2e_flow` does full e2e flow with guardian


crates/e2e-tests/src/e2e_flow.rs
```diff
@@ -255,10 +255,40 @@ mod tests {
...
     async fn test_bitcoin_withdrawal_e2e_flow() -> Result<()> {
+        run_bitcoin_withdrawal_e2e(false).await
+    }
+
+    #[tokio::test]
+    async fn test_bitcoin_withdrawal_with_guardian_e2e_flow() -> Result<()> {
+        run_bitcoin_withdrawal_e2e(true).await
+    }
+
+    async fn run_bitcoin_withdrawal_e2e(with_guardian: bool) -> Result<()> {
```
bmwill, line 266: I suppose this makes sense while we are getting the integration hooked up. We may want to evaluate if we want to have all tests run with guardian by default eventually.

crates/e2e-tests/src/guardian_harness.rs
```diff
@@ -0,0 +1,136 @@
...
+    enclave: Arc<Enclave>,
+    endpoint: String,
+    shutdown_tx: Option<oneshot::Sender<()>>,
+    server_handle: Option<JoinHandle<()>>,
+}
+
+impl GuardianHarness {
+    /// Start an operator-init'd guardian. Withdrawal RPCs stay gated until
+    /// [`Self::finalize`] completes provisioner-init.
+    pub async fn start(network: Network) -> Result<Self> {
```
bmwill, line 38: At which stage is the guardian pubkey available? Here or after finalize? we do need to commit to it, and likely a url of some such, onchain and it would probably be beneficial to be able to have the key available at contract init time.

crates/hashi-guardian/Cargo.toml
```diff
@@ -27,11 +27,15 @@ aws-config = "1.1.7"
 aws-sdk-s3 = "1.12.0"
 aws-credential-types = "1"
 
+aws-smithy-mocks = { version = "0.1", optional = true }
```
mskd12, line 30: nit: any way to remove the mock library from the main dependency list? IIUC it is just used by the test harness, right?

crates/hashi-guardian/src/enclave.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+/// Configuration set during initialization (immutable after set)
+pub struct EnclaveConfig {
+    /// Ephemeral keypair (set on boot)
+    eph_keys: EphemeralKeyPairs,
+    /// S3 client & config (set in operator_init)
+    s3_logger: OnceLock<S3Logger>,
+    /// Enclave BTC private key (set in provisioner_init)
+    pub(crate) enclave_btc_keypair: OnceLock<Keypair>,
```
mskd12: i think we can get rid of the `pub(crate)` and add functions in EnclaveConfig like `is_enclave_btc_keypair_set()` and `is_hashi_btc_master_pubkey_set()`, no?

crates/hashi-guardian/Cargo.toml
0xsiddharthks: Yeah, afaik we can remove it, but then we have to pass `--features test-utils` everytime we want to run a test.

but i think we might be able to manage that via nextest profiles. 

let me check.

crates/hashi-guardian/src/enclave.rs
0xsiddharthks: Yeah! thats much better than this.
Let me add this.

crates/e2e-tests/src/guardian_harness.rs
bmwill: From a conversation we had today, the committing onchain will be done post this series

crates/e2e-tests/src/guardian_harness.rs
```diff
@@ -0,0 +1,136 @@
...
+        let enclave = create_operator_initialized_enclave(
+            OperatorInitTestArgs::default().with_network(network),
+        )
+        .await;
+
+        let listener = TcpListener::bind("127.0.0.1:0")
+            .await
+            .context("bind guardian harness listener")?;
+        let addr: SocketAddr = listener.local_addr()?;
+        let endpoint = format!("http://{addr}");
```
bmwill, line 48: Just a note, do we want to use traditional TLS termination or require mTLS from the validators to the guardian?

crates/e2e-tests/src/lib.rs
```diff
@@ -265,6 +294,41 @@ impl TestNetworksBuilder {
...
+
+    let nodes = networks.hashi_network.nodes();
+    anyhow::ensure!(
+        !nodes.is_empty(),
+        "no hashi nodes to provision guardian from"
+    );
+
+    nodes[0]
+        .wait_for_mpc_key(std::time::Duration::from_secs(120))
+        .await?;
```
bmwill, line 309: We do we need this if we fetch the key from onchain below?

crates/e2e-tests/src/lib.rs
```diff
@@ -265,6 +294,41 @@ impl TestNetworksBuilder {
...
+
+    let hashi = nodes[0].hashi();
+    let committee = hashi
+        .onchain_state()
+        .current_committee()
+        .ok_or_else(|| anyhow::anyhow!("no current committee after DKG"))?;
+    let master_pubkey = hashi.get_onchain_mpc_pubkey()?;
```
bmwill, line 316: s/master/mpc

crates/hashi/src/config.rs
```diff
@@ -134,6 +134,11 @@ pub struct Config {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub screener_endpoint: Option<String>,
 
+    /// URL of the `hashi-guardian` gRPC endpoint. When not set, the guardian
+    /// integration is bypassed.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub guardian_endpoint: Option<String>,
```
bmwill, line 140: maybe add a TODO that this is temporary and will be removed once we commit it onchain?

crates/hashi/src/grpc/guardian_client.rs
```diff
@@ -0,0 +1,40 @@
...
+use tonic::transport::Endpoint;
+
+type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
+
+/// Lazy gRPC channel to a `hashi-guardian`.
+#[derive(Clone, Debug)]
+pub struct GuardianClient {
+    endpoint: String,
+    channel: Channel,
+}
```
bmwill, line 17: We should instrument this with metrics like we do our other gRPC client. Happy to add a TODO and add them as a follow up PR.

crates/hashi-guardian/src/enclave.rs
```diff
@@ -0,0 +1,488 @@
...
+    rate_limiter: OnceLock<Arc<tokio::sync::Mutex<RateLimiter>>>,
+}
+
+/// Scratchpad used only during initialization.
+/// Note: We don't clear it post-init because it does not have a lot of data.
+#[derive(Default)]
+pub struct Scratchpad {
+    /// The received shares
+    /// TODO: Investigate if it can be moved to std::sync::Mutex
+    pub shares: tokio::sync::Mutex<Vec<Share>>,
```
bmwill, line 72: yeah we should have this be `std::sync::Mutex`

crates/hashi-guardian/src/lib.rs
```diff
@@ -8,4 +8,28 @@ pub const HEARTBEAT_INTERVAL: Duration = Duration::from_mins(1);
...
 pub const MAX_HEARTBEAT_FAILURES_INTERVAL: Duration = Duration::from_mins(5);
 
+pub mod enclave;
+pub mod getters;
+pub mod heartbeat;
+pub mod init;
+pub mod rpc;
 pub mod s3_logger; // used by the monitor
+pub mod setup;
+pub mod withdraw;
```
bmwill, line 18: were these files just not getting compiled?

crates/hashi-guardian/src/main.rs
```diff
@@ -2,102 +2,20 @@
...
 use tonic_health::server::health_reporter;
 use tracing::info;
 
-mod getters;
-mod heartbeat;
-mod init;
-mod rpc;
-mod s3_logger;
-mod setup;
-mod withdraw;
```
bmwill, line 32: ah this is where they were.

bmwill (approved):
not major issues, mostly just some nits or asks to add TODOs so we don't forget things.

PR #468, opened 2026-04-17
metrics: observe RPC body sizes

bmwill:
Add a tower `CallbackLayer` wired up for both the axum server router and outbound tonic clients (via sui-http's `callback::MakeCallbackHandler`) that records per-route body size and byte-transfer metrics:

- `hashi_request_size_bytes` / `hashi_response_size_bytes` histograms
- `hashi_bytes_sent_total` / `hashi_bytes_received_total` counters

All four are labeled `path` and `role`, where `role` is RFC 9110 terminology for whether this node is acting as the `client` or `server` on the observed connection. The existing `hashi_inflight_requests`, `hashi_requests`, and `hashi_request_latency` metrics gain the same `role` dimension so client-side and server-side traffic can be separated in queries.

Client-side observation requires handing the `Metrics` registry to every outbound `grpc::Client`. `CommitteeSet::set_metrics` plumbs the registry through `OnchainState`, and `Client::with_metrics` attaches the callback layer to the `tonic_rustls::Channel` transport. A boxed transport type (`BoxedChannel`) keeps `Client` cloneable regardless of whether the layer is attached.

Also extract the tower wiring into `src/grpc/metrics_layer.rs` and bump `sui-http` to 0.2.0 for the updated `MakeCallbackHandler` trait.

### DIFF
--- a/crates/hashi/src/grpc/metrics_layer.rs
+++ b/crates/hashi/src/grpc/metrics_layer.rs
@@ -1,0 +1,382 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Tower callback layer plumbing for HTTP/gRPC metrics.
+//!
+//! This module owns every [`sui_http::middleware::callback`] handler the
+//! crate uses. A single [`RpcMetricsMakeCallbackHandler`] type drives
+//! both the server-side handler attached to the axum router in
+//! `crate::grpc::HttpService::start` and the client-side handler attached
+//! to outbound `tonic_rustls::Channel`s in `crate::grpc::client::Client`;
+//! the per-role behavior is selected via the [`Role`] passed at
+//! construction. The actual prometheus registration lives in
+//! [`crate::metrics::Metrics`]; this module is purely the wiring that
+//! observes traffic and writes into those metrics.
+
+use std::borrow::Cow;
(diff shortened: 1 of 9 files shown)


crates/hashi/src/grpc/metrics_layer.rs
```diff
@@ -0,0 +1,376 @@
...
+
+        self.metrics
+            .requests
+            .with_label_values(&[self.path.as_ref(), status, self.role.label()])
+            .inc();
+
+        self.counted_response = true;
+    }
+
+    fn on_service_error<E>(&mut self, _error: &E)
```
dlukegordon, line 271: This shouldn't ever occur on the server side but I think it could still occur on the client side?

crates/hashi/src/grpc/metrics_layer.rs
```diff
@@ -0,0 +1,376 @@
...
+        }
+    }
+}
+
+/// Direction of HTTP body bytes from the perspective of this node.
+/// Selects which byte counter (`bytes_sent_total` vs
+/// `bytes_received_total`) the observed bytes are flushed to, and —
+/// together with [`Role`] — which body (request vs response) the tally
+/// was tracking.
+#[derive(Clone, Copy)]
```
dlukegordon: Nit but might as well add Debug if we're adding it for Role

PR #469, opened 2026-04-17
Make `SigningManager` access epoch-typed and reconcile from on-chain state

zhouwfang:
#179

### DIFF
--- a/crates/e2e-tests/src/lib.rs
+++ b/crates/e2e-tests/src/lib.rs
@@ -442,6 +442,33 @@ mod tests {
         nodes[0].current_epoch().unwrap()
     }
 
+    async fn wait_for_signing_manager(
+        nodes: &[HashiNodeHandle],
+        epoch: u64,
+        timeout: std::time::Duration,
+    ) -> Result<()> {
+        let deadline = tokio::time::Instant::now() + timeout;
+        loop {
+            let all_ready = nodes
+                .iter()
+                .all(|node| node.hashi().signing_manager_for(epoch).is_some());
+            if all_ready {
+                return Ok(());
+            }
+            if tokio::time::Instant::now() >= deadline {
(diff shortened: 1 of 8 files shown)


crates/e2e-tests/src/lib.rs
```diff
@@ -1267,6 +1297,60 @@ mod tests {
...
+        const RECOVERY_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
+
+        tracing_subscriber::fmt()
+            .with_test_writer()
+            .with_env_filter(
+                tracing_subscriber::EnvFilter::from_default_env()
+                    .add_directive(tracing::Level::INFO.into()),
+            )
+            .try_init()
+            .ok();
```
bmwill: you probably should just use init_test_logging

crates/hashi/src/mpc/service.rs
```diff
@@ -440,17 +434,60 @@ impl MpcService {
...
+                     falling back to protocol"
+                );
+                if let Err(e) = self.prepare_signing(epoch, &output).await {
+                    error!("sync_if_stale: protocol fallback failed for epoch {epoch}: {e}");
+                    return;
+                }
+                info!("sync_if_stale: recovered epoch {epoch} via protocol path");
+            }
+        }
+        let _ = self.key_ready_tx.send(Some(output.public_key));
```
bmwill, line 477: I assume this should be safe to reset multiple times since its expected that the pubkey is exactly the same? (the recovery path does validate this correct?)

zhouwfang: Yes, `watch::send` is idempotent for receivers, and `complete_key_rotation` errors on pubkey divergence, so any Ok from `recover_mpc_state` matches the lineage.

bmwill (approved):

PR #482, opened 2026-04-22
Fix rotation reconstruction panic when `previous_output` is unset

zhouwfang:
#95

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -392,12 +392,16 @@ impl MpcManager {
         {
             return Ok(cached_response.clone());
         }
-        let messages = if cache_is_current
-            && let Some(m) = self.get_dealer_messages(
-                request.protocol_type,
-                &request.dealer,
-                request.batch_index,
-            ) {
+        let cached_messages = cache_is_current
+            .then(|| {
+                self.get_dealer_messages(
+                    request.protocol_type,
+                    &request.dealer,
+                    request.batch_index,
+                )
+            })
+            .flatten();
+        let messages = if let Some(m) = cached_messages {
             m
         } else {
             let from_db = match request.protocol_type {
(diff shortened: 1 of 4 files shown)


crates/hashi/src/mpc/mpc_except_signing_tests.rs
```diff
@@ -3676,54 +3624,6 @@ async fn test_recover_shares_via_complaint_insufficient_signers() {
     );
 }
 
-#[tokio::test]
-#[should_panic(expected = "cannot have complaint without message")]
-async fn test_recover_shares_via_complaint_no_dealer_message() {
```
jonas-lj, line 3681: Can you add a test to demonstrate that the panic has been fixed?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -387,17 +387,46 @@ impl MpcManager {
...
         // It is safe to return a response from cache since we already know that dealer was malicious.
-        if let Some(cached_response) = self.complaint_responses.get(&cache_key) {
+        let cache_is_current = request.epoch == self.mpc_config.epoch;
+        if cache_is_current && let Some(cached_response) = self.complaint_responses.get(&cache_key)
+        {
             return Ok(cached_response.clone());
         }
-        let messages = self
-            .get_dealer_messages_with_db_fallback(
+        let messages = if cache_is_current
```
jonas-lj: nit: I think this can be written a bit simpler with cache_is_current.then

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -3333,112 +3365,130 @@ impl MpcManager {
...
+                        mpc_manager,
+                        &dealer_address,
+                        &rotation_msgs,
+                        signers,
+                        p2p_channel,
+                        source_epoch,
+                    )
+                    .await?;
+                    let mut mgr = mpc_manager.write().unwrap();
+                    for (share_index, output) in recovered {
```
jonas-lj, line 3462: nit: This `share_index` shadows the one from `NeedsRotationComplaintRecovery`, so the latter should perhaps have another name?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -387,17 +387,46 @@ impl MpcManager {
...
+                    .public_messages_store
+                    .get_dealer_message(request.epoch, &request.dealer)
+                    .map_err(|e| MpcError::StorageError(e.to_string()))?
+                    .map(Messages::Dkg),
+                ProtocolTypeIndicator::KeyRotation => self
+                    .public_messages_store
+                    .get_rotation_messages(request.epoch, &request.dealer)
+                    .map_err(|e| MpcError::StorageError(e.to_string()))?
+                    .map(Messages::Rotation),
+                ProtocolTypeIndicator::NonceGeneration => match request.batch_index {
```
jonas-lj: nit: I think this can be written simpler as request.batch_index.map(...)

jonas-lj (approved):

crates/hashi/src/mpc/mpc_except_signing_tests.rs
zhouwfang: `e2e_tests::test_rotation_reconstruction_complaint_recovery_after_restart` exercises the scenario, which panicked intermittently in CI. I'd love to add a unit test, but it is too hard.

PR #484, opened 2026-04-23
[3/n][guardian-integration] add local limiter (synced with guardian) to hashi nodes

0xsiddharthks:
Adds an local emulator of the guardian's token bucket (and sequence).

- Hashi:
  - Adds `LocalLimiter` (`crates/hashi/src/guardian_limiter.rs`)
  - Bootstrap the local limiter during hashi startup
- Guardian:
  - `GetGuardianInfoResponse` extended with `LimiterState` + `LimiterConfig`

We are not currently updating the local limiter when processing new transactions. that is being done in #495


crates/hashi-types/proto/sui/hashi/v1alpha/guardian_service.proto
```diff
@@ -200,6 +206,15 @@ message LimiterState {
   optional uint64 next_seq = 3;
 }
 
+// Immutable limiter configuration.
```
bmwill, line 209: This isn't exactly immutable since the guardian can opt to change it

bmwill (approved):

PR #486, opened 2026-04-23
Restarted dealer resumes with the same messages instead of regenerating

zhouwfang:
#14 #95 #193

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -297,7 +297,7 @@ impl MpcManager {
         }
         let result = match &request.messages {
             Messages::Dkg(msg) => {
-                self.store_dkg_message(self.mpc_config.epoch, sender, msg)?;
+                self.cache_and_persist_dkg_message(self.mpc_config.epoch, sender, msg)?;
                 self.try_sign_dkg_message(sender, &request.messages)
             }
             Messages::Rotation(msgs) => {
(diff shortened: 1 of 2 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2173,17 +2173,27 @@ impl MpcManager {
...
+                .public_messages_store
+                .get_dealer_message(self.mpc_config.epoch, &self.address)
+            {
+                Ok(Some(msg)) => {
+                    self.dkg_messages.insert(self.address, msg.clone());
+                    Messages::Dkg(msg)
+                }
+                Ok(None) => {
+                    let msg = self.create_dealer_message(rng);
+                    self.store_dkg_message(self.mpc_config.epoch, self.address, &msg)?;
```
jonas-lj: I don't think it's obvious what this function is doing, so since you're using self.dkg_messages and self.public_messages_store directly elsewhere, it'll be easier to read if you just insert into those directly here.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2196,11 +2206,21 @@ impl MpcManager {
...
+                .public_messages_store
+                .get_rotation_messages(self.mpc_config.epoch, &self.address)
+            {
+                Ok(Some(msgs)) => {
+                    self.rotation_messages.insert(self.address, msgs.clone());
+                    Messages::Rotation(msgs)
+                }
+                Ok(None) => {
+                    let msgs = self.create_rotation_messages(previous, rng);
+                    self.store_rotation_messages(self.mpc_config.epoch, self.address, &msgs)?;
```
jonas-lj: Same as for DKG messages

jonas-lj (approved):

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Good point. I renamed them to be `cache_and_persist_*`. They are used at multiple other places too.

PR #487, opened 2026-04-24
Extract MPC into a separate crate

zhouwfang:
#16


bmwill:
Whats the main motivation for this? our code shouldn't be big enough yet to really warrant splitting this yet to get a perf bump

zhouwfang:
> Whats the main motivation for this? our code shouldn't be big enough yet to really warrant splitting this yet to get a perf bump

Not perf, layering for code health. It makes the 3 fields in Hashi struct owned by MpcServiceState in the new crate, and Hashi delegates. So Hashi becomes a consumer of an MPC library rather than the owner of MPC internals. State changes become local to the mpc crate instead of touching Hashi.

zhouwfang:
Another motivation is to make the mpc crate more reusable for other potential use cases.

bmwill:
I don't think this is a great idea, and its very out of date so i'm going to close this for now.

PR #492, opened 2026-04-27
Keep executed proposals on-chain and emit data in ProposalExecutedEvent

Bridgerz:
## Summary
- Stop deleting governance proposals on `execute` — they stay in the on-chain `ObjectBag` with an `executed: bool` flag so the actual action that was taken remains inspectable forever.
- `ProposalExecutedEvent<T>` now carries the typed `data: T` so off-chain consumers can see what was executed without re-fetching the (now-deleted) proposal.
- Adds `copy` to the five proposal data types so they can be embedded in the event.

## Why `copy` on the proposal data types is safe
The data types (`UpdateConfig`, `EnableVersion`, `DisableVersion`, `EmergencyPause`, `Upgrade`) are pure *descriptors* of governance intent: a key/value pair, a version number, a digest, a bool. The authority to act on them lives in the `Proposal<T>` lifecycle (quorum + `execute`), not in the descriptor itself, so duplicating the value carries no privilege. Every field already supports `copy` (`String`, `u64`, `bool`, `Value`, `vector<u8>`), so this is mechanically valid.

**Forward trade-off:** a future proposal type that carries a non-`copy` resource (e.g. an `EnableNewCoin` proposal that holds a `TreasuryCap<X>`) will not satisfy `T: copy + drop + store`. When that arrives we can add a separate `execute_consuming<T: store>` path that removes from the bag and emits a data-less event, leaving this descriptor path untouched. The current shape is opinionated for the descriptor case, not a ceiling.

## Re-execution and expiry guards
Keeping proposals around after execute requires guarding against re-execution. `execute` now asserts `!executed`, sets it to `true`, and emits/returns the data. `delete_expired` also refuses already-executed proposals so they're kept forever even past the 7-day expiry.

## Off-chain (Rust) changes
- `move_types::Proposal<T>` mirrors the new `executed` field — BCS positional, must match the on-chain layout.
- `move_types::ProposalExecutedEvent` exposes `data_bcs: Vec<u8>` (the proposal id is a fixed 32-byte `Address`, so the remaining bytes are the typed payload — decode per `proposal_type` downstream).
- `onchain::types::Proposal` carries `executed`; `scrape_proposals` plumbs it through.
- Watcher: `ProposalCreatedEvent` initializes `executed: false`; `ProposalExecutedEvent` no longer removes the entry, it marks `executed = true`.
- Existing config-refetch on `UpdateConfig`/`EmergencyPause` is left in place — applying the data directly from the event is a follow-up.

## Coordination
This change is wire-incompatible with previously deployed packages (BCS layout of `Proposal<T>` and `ProposalExecutedEvent` both grow). Rolling out requires a coordinated package upgrade + validator/CLI deploy.

## Test plan
- [x] `sui move test` — all 76 Move tests pass, including new `test_re_execute_proposal_fails` and `test_delete_expired_executed_proposal_fails`
- [x] `cargo check --workspace --all-targets` — clean
- [x] `cargo test -p hashi-types -p hashi --lib` — 42/42 pass
- [ ] End-to-end: run an `update_config` proposal on localnet, verify `ProposalExecutedEvent` carries the typed data and the proposal stays queryable on-chain with `executed: true`

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -42,7 +42,7 @@ pub struct Hashi {
     pub committees: CommitteeSet,
     pub config: Config,
     pub treasury: Treasury,
-    pub proposals: ObjectBag,
+    pub proposals: Proposals,
     /// TOB certificates by (epoch, batch_index) -> EpochCertsV1
     pub tob: Bag,
     /// Number of presignatures consumed in the current epoch.
@@ -222,6 +222,15 @@ pub struct MetadataCap {
 pub struct Coin {
     pub id: Address,
     pub balance: u64,
+}
+
+/// Rust version of the Move hashi::proposals::Proposals type.
+#[derive(Debug, serde_derive::Deserialize)]
+pub struct Proposals {
+    /// Proposals that have been created but not yet executed.
+    pub active: ObjectBag,
+    /// Proposals that have executed successfully (kept indefinitely).
+    pub executed: ObjectBag,
 }
 
 /// Rust version of the Move hashi::deposit_queue::DepositRequestQueue type.
@@ -410,7 +419,6 @@ pub struct Proposal<T> {
     pub timestamp_ms: u64,
     pub metadata: VecMap<String, String>,
     pub data: T,
-    pub executed: bool,
 }
 
 /// Rust version of the Move hashi::update_config::UpdateConfig type.
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -410,6 +410,7 @@ pub struct Proposal<T> {
     pub timestamp_ms: u64,
     pub metadata: VecMap<String, String>,
     pub data: T,
+    pub executed: bool,
 }
 
 /// Rust version of the Move hashi::update_config::UpdateConfig type.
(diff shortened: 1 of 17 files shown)


crates/hashi-types/src/move_types/mod.rs
```diff
@@ -410,6 +410,7 @@ pub struct Proposal<T> {
     pub timestamp_ms: u64,
     pub metadata: VecMap<String, String>,
     pub data: T,
+    pub executed: bool,
```
bmwill: Option<Timestamp> would probably be more ideal?

Also we should do the same trick and have a "completed" Bag for these so listing current proposals is easier to do.

Bridgerz: Ya I thought to do the same but we don't actually ever need to iterate over proposals like we do btc requests. That being said, it would be nice to keep the symmetry. Will do.

crates/hashi-types/src/move_types/mod.rs
```diff
@@ -410,7 +419,6 @@ pub struct Proposal<T> {
     pub timestamp_ms: u64,
     pub metadata: VecMap<String, String>,
     pub data: T,
-    pub executed: bool,
```
bmwill, line 413: I do still think that its worthwhile to include some extra metadata about the proposal on execution. Whether we include this as explicit fields or if we append the info to the `metadata` map it doesn't matter.

In particular I think we should include:
- execution timestamp
- execution hashi epoch (this is important because it allows us to link the `votes` to the actual committee that approved it)

bmwill (approved):
lgtm, though we should look at adding some extra metadata at some point.

PR #493, opened 2026-04-27
Read bitcoin_confirmation_threshold live in deposit confirmation

Bridgerz:
## Summary
The BTC monitor was caching `bitcoin_confirmation_threshold` into `MonitorConfig` once at validator startup, so an on-chain governance change to the threshold required every validator to restart before pending deposits could pick up the new value. Withdrawals already read the threshold live on each evaluation in the leader (`leader/mod.rs:1313`); deposits did not. This PR closes that asymmetry.

## Approach
Replace the cached `u32` with a tiny trait that the monitor calls per deposit evaluation:

```rust
pub trait ConfirmationThresholdSource: Send + Sync {
    fn current(&self) -> u32;
}
```

- `OnchainState` implements it by delegating to its existing `bitcoin_confirmation_threshold()` accessor — same source of truth as withdrawals, no parallel cache to keep in sync.
- `StaticThreshold(u32)` for tests and the standalone `testnet_monitor` example.
- The monitor stays Sui-unaware: it only sees `Arc<dyn ConfirmationThresholdSource>` injected via the builder.

## Why this and not other options
- **`watch::Receiver<u32>` synced by the watcher** — works but introduces a second source of truth that has to be kept in lockstep with `state.hashi.config`. Subtle drift bug risk for no real benefit; the read isn't on a hot path.
- **Move the comparison out of the monitor entirely** — gives architectural symmetry with withdrawals (Sui-aware layer does the compare) but requires unwinding the deposit pipeline. Larger blast radius for marginal benefit.
- **Inject `Arc<OnchainState>` directly** — couples the monitor crate to Sui state. Violates the existing boundary.

The trait approach gets us actual symmetry (both deposit and withdrawal threshold reads now bottom out in `OnchainState::bitcoin_confirmation_threshold()`) at minimum surface area.

## Audit of other governance-controlled configs
I traced every `OnchainState` consumer to find any other cache-at-boot bugs:

- ✅ Already live: `paused()`, `bitcoin_confirmation_threshold()` for withdrawals, deposit/withdrawal minimums (enforced on-chain at request creation), withdrawal/deposit queues, epoch, committee membership.
- ❌ The bug fixed here: deposit-side `bitcoin_confirmation_threshold`.
- ⚠️ Per-epoch snapshot, **likely intentional**: `mpc_threshold_in_basis_points`, `mpc_weight_reduction_allowed_delta`, `mpc_max_faulty_in_basis_points` are baked into the `Committee` struct at construction. Threshold-signature schemes generally need parameters stable within a protocol run, so this is probably deliberate — flagging for the MPC owners but not changing it here.

Nothing else cached.

## Test plan
- [x] `cargo test -p hashi --lib` — 274/274 pass
- [x] New unit tests in `btc_monitor::config::tests`:
  - `static_threshold_returns_stored_value`
  - `builder_with_u32_wraps_in_static_source`
  - `live_source_picks_up_updates` — flips an `AtomicU32` post-build and verifies the monitor sees the new value
- [ ] End-to-end on localnet: start a validator, fire an `update_config` proposal for `bitcoin_confirmation_threshold`, verify a pending deposit confirms at the new threshold without a restart

### DIFF
--- a/crates/hashi/src/btc_monitor/config.rs
+++ b/crates/hashi/src/btc_monitor/config.rs
@@ -240,7 +240,8 @@ pub fn network_from_chain_id(chain_id: &str) -> Option<Network> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::sync::atomic::{AtomicU32, Ordering};
+    use std::sync::atomic::AtomicU32;
+    use std::sync::atomic::Ordering;
 
     /// Test fake whose threshold can be mutated post-construction so we
     /// can verify the monitor reads the value live, not at boot.
--- a/crates/hashi/src/btc_monitor/config.rs
+++ b/crates/hashi/src/btc_monitor/config.rs
@@ -5,19 +5,45 @@
 
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::Arc;
 
 pub use bitcoin::BlockHash;
 pub use bitcoin::Network;
 use bitcoin::blockdata::constants::genesis_block;
 pub use corepc_client;
 
-#[derive(Debug, Clone)]
+/// Live source of the BTC confirmation threshold, read on every deposit
+/// evaluation so on-chain governance changes (`update_config` for
+/// `bitcoin_confirmation_threshold`) take effect without a validator
+/// restart.
+///
+/// Implemented by `OnchainState` for production and by `StaticThreshold`
+/// for tests/examples that don't need live updates.
+pub trait ConfirmationThresholdSource: Send + Sync {
+    fn current(&self) -> u32;
+}
+
+/// Fixed-value implementation of `ConfirmationThresholdSource` for use
+/// in tests, the standalone `testnet_monitor` example, and as the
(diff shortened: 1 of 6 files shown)


crates/hashi/src/btc_monitor/config.rs
```diff
@@ -5,19 +5,45 @@
...
 pub struct MonitorConfig {
     /// Bitcoin network to connect to
     pub network: Network,
 
-    /// Number of confirmations required for a transaction to be considered canonical
-    pub confirmation_threshold: u32,
+    /// Live source of the number of confirmations required for a
+    /// transaction to be considered canonical. Read at every deposit
+    /// evaluation so governance updates are picked up immediately.
+    pub confirmation_threshold: Arc<dyn ConfirmationThresholdSource>,
```
bmwill: Instead of this, why don't we have the API which core uses to check for confirmation, pass in the confirmation value explicitly. It removes some of this odd boilerplate too.

bmwill (approved):
lgtm

PR #495, opened 2026-04-28
[4/n][guardian-integration] validate local limiter at MPC signing time on every node

0xsiddharthks:
## Summary

- Each committee member validates its own local limiter (capacity + timestamp + leader-supplied `expected_limiter_seq`) before participating in MPC signing.
- Leader peeks `next_seq`, validates locally before broadcasting, and skips the round if its own limiter would reject (retries on the next checkpoint).


crates/hashi-types/proto/sui/hashi/v1alpha/bridge_service.proto
```diff
@@ -128,6 +128,12 @@ message SignWithdrawalTxConstructionResponse {
 message SignWithdrawalTransactionRequest {
   // The id of the WithdrawalTransaction on Sui (32 bytes).
   bytes withdrawal_txn_id = 1;
+
+  // Set when the leader's local guardian limiter is initialized. Each
+  // committee member validates this against its own limiter's `next_seq`
+  // before participating in MPC signing, and uses it when advancing local
+  // state after signing succeeds.
+  optional uint64 expected_limiter_seq = 2;
 }
```
bmwill, line 136: I think it makes sense to want to do a preflight before we ask the guardian, at this point we've already committed to signing a txn so the worst thing that happens is we wait.

IIUC we probably also want to do this preflight check during the txn construction phase too right? Or is that coming in a follow up PR?

crates/hashi/src/leader/mod.rs
```diff
@@ -1167,6 +1167,29 @@ impl LeaderService {
         }
         info!("MPC signing withdrawal transaction");
 
+        // Leader-side limiter check before broadcasting MPC: peek next_seq
+        // and validate locally so we don't waste an MPC round if our own
+        // limiter would reject the consume. The same expected_seq is
+        // forwarded to every committee member via the MPC request.
+        let expected_limiter_seq = if let Some(limiter) = inner.local_limiter() {
```
bmwill: We should ensure that we either only schedule a single txn to sign at a time or take into account that we may have more in flight at a time.

crates/hashi/src/withdrawals.rs
```diff
@@ -628,6 +651,14 @@ impl Hashi {
 
             signatures_by_input.push(signature);
         }
+        if let Some((limiter, seq, ts, amt)) = limiter_consume {
+            limiter.apply_consume(seq, ts, amt).await.map_err(|e| {
```
bmwill: How are we planning on handle the application of the consume? we probably don't want to actually consume the limit permanently until signing is confirmed onchain

crates/hashi/src/leader/mod.rs
bmwill: Also thinking on this a bit more we probably need the limiter amount to be included as an input into the txn construction process so that we can properly select requests which fit into the limit. We can probably do so as a follow up though, can you leave a comment for this as a TODO?

crates/hashi-types/proto/sui/hashi/v1alpha/bridge_service.proto
0xsiddharthks: Yes, we are doing both.

This PR adds the preflight at MPC time (leader peeks `next_seq` + `validate_consume` before broadcasting MPC). 

The txn-construction-phase preflight is in #423 `process_approved_withdrawal_requests` truncates the approved batch to the longest timestamp-sorted prefix that fits the local limiter's currently-available capacity.

0xsiddharthks: Yes, thats done in the followup PRs in the stack ( #523 ) :)

crates/hashi/src/leader/mod.rs
0xsiddharthks: Yes, this is also done in the followup pr #523 . we are setting max_concurrent to 1 if guardian is concigured. 

more explanation is in the PR summary of #523

crates/hashi/src/withdrawals.rs
0xsiddharthks: I've refactored my strategy a bit here:

1. **Within this PR**: the signing path is now validate-only — both the leader pre-MPC check and the per-member MPC check call `LocalLimiter::validate_consume` and never mutate state. See `crates/hashi/src/withdrawals.rs:601` and the comment block at line 598-600.

2. **In #523**: the *only* mutator of `LocalLimiter` after bootstrap is `start_guardian_limiter_observer` (`crates/hashi/src/lib.rs:625`). It listens to the on-chain watcher's `Notification::WithdrawalSigned` and calls `apply_consume` on every observed signed event. So the local limiter advances iff the BLS-signed `sign_withdrawal` lands on-chain, which the leader only submits after `finalize_withdrawal_through_guardian` returns Ok (#423). Effectively: on-chain visibility of the signed withdrawal is the durable proof that the guardian has acked, and that's what advances the limiter — symmetrically across leader and follower since every node consumes the same checkpoint stream.

---

The last step is to introduce a softreserve per WID at the local limiter level. thats also followup in the stack.

bmwill (approved):

PR #497, opened 2026-04-28
Add abort reconfig governance proposal

Bridgerz:
## Summary
- add an AbortReconfig governance proposal that clears pending reconfig state through current-committee quorum
- emit AbortReconfigEvent and keep the committed epoch/MPC key unchanged
- wire the proposal through Rust on-chain parsing, CLI, metrics labels, and proposal cleanup
- document reconfig failure scenarios that can require an abort

## Tests
- sui move test
- cargo check --workspace

### DIFF
--- a/packages/hashi/sources/core/proposal/types/abort_reconfig.move
+++ b/packages/hashi/sources/core/proposal/types/abort_reconfig.move
@@ -9,25 +9,42 @@
 /// committee with stable on-chain voting power.
 module hashi::abort_reconfig;
 
-use hashi::{hashi::Hashi, proposal, reconfig};
+use hashi::{hashi::Hashi, proposal};
 use std::string::String;
 use sui::{clock::Clock, vec_map::VecMap};
 
 const THRESHOLD_BPS: u64 = 6667;
+const ENotReconfiguring: u64 = 0;
+const EWrongReconfigEpoch: u64 = 1;
 
-public struct AbortReconfig has drop, store {}
+public struct AbortReconfig has copy, drop, store {
+    epoch: u64,
+}
 
 public fun propose(
     hashi: &mut Hashi,
+    epoch: u64,
     metadata: VecMap<String, String>,
     clock: &Clock,
     ctx: &mut TxContext,
 ): ID {
     hashi.config().assert_version_enabled();
-    proposal::create(hashi, AbortReconfig {}, THRESHOLD_BPS, metadata, clock, ctx)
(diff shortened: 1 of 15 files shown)


zhouwfang:
#163

packages/hashi/sources/core/proposal/types/abort_reconfig.move
```diff
@@ -0,0 +1,33 @@
...
+/// next committee cannot complete DKG/key rotation or cannot produce the
+/// `end_reconfig` certificate, the last committed committee is the only
+/// committee with stable on-chain voting power.
+module hashi::abort_reconfig;
+
+use hashi::{hashi::Hashi, proposal, reconfig};
+use std::string::String;
+use sui::{clock::Clock, vec_map::VecMap};
+
+const THRESHOLD_BPS: u64 = 6667;
```
bmwill, line 16: Not related to this PR but do we want to revisit the idea of having configurable thresholds so that if we want to update a threshold we don't have to do a full package upgrade?

packages/hashi/sources/core/reconfig.move
```diff
@@ -57,9 +56,11 @@ entry fun end_reconfig(
...
 }
 
-// TODO: Re-enable with committee certificate verification.
-entry fun abort_reconfig(_self: &mut Hashi, _ctx: &TxContext) {
-    abort EAbortReconfigDisabled
+public(package) fun abort_reconfig(self: &mut Hashi, ctx: &TxContext) {
+    self.config().assert_version_enabled();
+    assert!(self.committee_set().is_reconfiguring(), ENotReconfiguring);
+    let epoch = self.committee_set_mut().abort_reconfig(ctx);
+    sui::event::emit(AbortReconfigEvent { epoch });
```
bmwill: I don't think there is any value in having this function anymore, should we just have the proposal explicitly call self.committee_set_mut().abort_reconfig(ctx)?

packages/hashi/sources/core/proposal/types/abort_reconfig.move
```diff
@@ -0,0 +1,33 @@
...
+
+public struct AbortReconfig has drop, store {}
```
bmwill: Do we want to be explicit and have the proposal commit to the epoch that we want to abort?

Bridgerz:
@bmwill ready for another round

bmwill (approved):

PR #499, opened 2026-04-29
Optimize db backups with bulk ingestion and excluding nonces

dlukegordon:


crates/hashi/src/db.rs
```diff
@@ -137,35 +127,23 @@ impl Database {
             ))
         })?;
 
-        for (name, source_ks) in self.all_keyspaces() {
+        for (name, source_ks) in self.backup_keyspaces() {
```
zhouwfang, line 130: The old code had an explicit `BATCH_BYTE_BUDGET`. With `nonce_messages` excluded, the remaining keyspaces are small in steady state so OOM is unlikely, but the explicit safety-net is gone. What about adding a comment that fjall's bulk ingestion buffers are bounded internally with a reference?

dlukegordon: Added a comment

crates/hashi/src/db.rs
```diff
@@ -137,35 +127,27 @@ impl Database {
...
-        // rows live in the active memtable and are recovered from the journal
-        // on next open, so the destination directory is self-contained and
-        // safe to archive after this call returns. We do not fsync the dest
-        // directory entry here: this function's caller immediately tars the
-        // directory in the same process, so a crash before archival just
+        // `persist(SyncAll)` fsyncs fjall's active journal. The copied rows are
+        // bulk-ingested into table files rather than journaled, but fjall still
+        // creates an active journal for the destination DB. We do not fsync the
+        // dest directory entry here: this function's caller immediately tars
+        // the directory in the same process, so a crash before archival just
```
zhouwfang: nit: It is accurate that the rows skip the journal, but it leaves a reader wondering what `persist(SyncAll)` is actually fsyncing in a journal that has no copied-row writes. Could you spell out that persist(SyncAll) is still needed to fsync the manifest/SST file metadata so the bulk-ingested SSTs are reachable on next open? Otherwise it reads as if the call might be vestigial.

zhouwfang (approved):

PR #500, opened 2026-04-29
Add MPC config proposal CLI

zhouwfang:
#368

### DIFF
--- a/crates/hashi/src/cli/commands/proposal.rs
+++ b/crates/hashi/src/cli/commands/proposal.rs
@@ -509,6 +509,79 @@ pub async fn create_update_config_proposal(
     Ok(())
 }
 
+pub async fn create_update_mpc_config_proposal(
+    config: &CliConfig,
+    threshold_bps: Option<u64>,
+    max_faulty_bps: Option<u64>,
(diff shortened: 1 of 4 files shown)


crates/hashi/src/cli/commands/proposal.rs
```diff
@@ -509,6 +509,79 @@ pub async fn create_update_config_proposal(
...
+            f <= MAX_BPS,
+            "--max-faulty-bps must be in 0..={MAX_BPS}, got {f}"
+        );
+    }
+    if let Some(d) = weight_reduction_allowed_delta {
+        anyhow::ensure!(
+            d <= MAX_BPS,
+            "--weight-reduction-allowed-delta must be in 0..={MAX_BPS}, got {d}"
+        );
+    }
```
bmwill, line 538: do we have checks in move for these invariants? Or at least during reconfig so that we can abort it and restart with new values once they've been set?

zhouwfang: Yes, these are already enforced on-chain by `mpc_config::is_valid_config_entry`.

bmwill (approved):

PR #501, opened 2026-04-29
Write fjall backup directly into encrypted tarball

dlukegordon:


crates/hashi/src/db.rs
```diff
@@ -86,74 +84,24 @@ impl Database {
...
-        // function's caller immediately tars the directory in the same process,
-        // so a crash before archival just discards an incomplete backup and
-        // loses no committed state.
-        dest_db.persist(fjall::PersistMode::SyncAll)?;
+    pub(crate) fn backup_keyspace_names() -> [&'static str; 3] {
+        [
+            ENCRYPTION_KEYS_CF_NAME,
+            DEALER_MESSAGES_CF_NAME,
+            ROTATION_MESSAGES_CF_NAME,
+        ]
```
bmwill: Can we consolidate these to avoid drift?

dlukegordon: Yeah, that's what I originally had but it's just verbose if you want to be able to get the names without a Database instance. Will push.

dlukegordon: https://github.com/MystenLabs/hashi/pull/501/changes/c20b6b022edf243a6bef26bbc33581f47a19fa27

bmwill (approved):

crates/hashi/src/backup.rs
```diff
@@ -67,7 +72,13 @@ fn create_dir_strict(path: &Path) -> Result<()> {
 #[derive(serde::Deserialize, serde::Serialize)]
 pub struct BackupManifest {
```
zhouwfang, line 73: What about adding format_version and bumping on future format changes? The validation path can then reject incompatible backups with a clear error instead of cryptic deserialize failures.

crates/hashi/src/backup.rs
```diff
@@ -464,27 +473,64 @@ fn restore_db_entry<R: Read>(
...
-            output_path.display()
+            "Failed to restore database entry {} into {}",
+            archive_path.display(),
+            db_path.display()
         )
     })?;
     Ok(())
 }
 
+fn restore_db_backup_keyspace<R: Read>(
```
zhouwfang, line 512: It calls `fjall::Database::builder(dest).open()` per call, which means three full WAL-recovery / manifest-read cycles for the three backup keyspaces. Minor performance, no correctness issue. We can open the destination DB once in the caller, pass `&fjall::Database` down, persist once at the end.

dlukegordon: Good point, fixing

zhouwfang (approved):

PR #502, opened 2026-04-30
Generate encryption key per epoch and register on-chain

zhouwfang:
#95 #163

### DIFF
--- a/crates/hashi/src/db.rs
+++ b/crates/hashi/src/db.rs
@@ -357,10 +357,9 @@ impl Database {
 
     /// Prune all MPC keyspaces.
     pub fn prune_messages_below(&self, cutoff_epoch: u64) -> Result<()> {
-        // TODO(IOP-194): Drop this extra retention once `start_reconfig` rejects
-        // stale-tagged registrations and the per-epoch stake threshold is scoped
-        // to initial committee creation only. Until then, retention covers the
-        // race.
+        // Encryption keys retain one extra epoch to support graceful rotation: the
+        // committee may capture a previous epoch's pub key, and the matching
+        // private key must still be in DB.
         prune_keyspace(&self.encryption_keys, cutoff_epoch.saturating_sub(1))?;
         prune_keyspace(&self.dealer_messages, cutoff_epoch)?;
         prune_keyspace(&self.rotation_messages, cutoff_epoch)?;
--- a/crates/hashi/src/db.rs
+++ b/crates/hashi/src/db.rs
@@ -13,6 +13,7 @@ use fjall::Result;
 use sui_sdk_types::Address;
 
 use hashi_types::committee::EncryptionPrivateKey;
+use hashi_types::committee::EncryptionPublicKey;
 
 use serde::de::DeserializeOwned;
 
@@ -169,6 +170,32 @@ impl Database {
             ))
         })?;
         Ok(Some(EncryptionPrivateKey::from(scalar)))
+    }
+
+    pub fn find_encryption_key_matching(
+        &self,
+        target: &EncryptionPublicKey,
+    ) -> Result<Option<EncryptionPrivateKey>> {
+        let target_bytes = target.as_element().to_byte_array();
+        for guard in self.encryption_keys.iter() {
+            let value = guard.value()?;
+            let byte_array: [u8; 32] = (&*value).try_into().map_err(|_| {
+                fjall::Error::Io(std::io::Error::new(
+                    std::io::ErrorKind::InvalidData,
+                    "invalid point",
+                ))
+            })?;
+            let scalar = match RistrettoScalar::from_byte_array(&byte_array) {
+                Ok(s) => s,
+                Err(_) => continue,
+            };
+            let candidate = EncryptionPrivateKey::from(scalar);
+            let candidate_pub = EncryptionPublicKey::from_private_key(&candidate);
+            if candidate_pub.as_element().to_byte_array() == target_bytes {
+                return Ok(Some(candidate));
+            }
+        }
+        Ok(None)
     }
 
     pub fn store_dealer_message(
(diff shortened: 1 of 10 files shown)


benr-ml (commented):
we also need the same flows for the signing key, should we add them in this pr?

crates/hashi/src/db.rs
```diff
@@ -379,9 +406,13 @@ impl Database {
         self.nonce_messages.remove(key)
     }
 
-    /// Prune all MPC keyspaces, deleting entries with `epoch < cutoff_epoch`.
+    /// Prune all MPC keyspaces.
     pub fn prune_messages_below(&self, cutoff_epoch: u64) -> Result<()> {
-        prune_keyspace(&self.encryption_keys, cutoff_epoch)?;
+        // TODO: Drop this extra retention once `start_reconfig` rejects
+        // stale-tagged registrations and the per-epoch stake threshold is scoped
```
bmwill: what does stale-tagged registration mean here?

crates/hashi/src/sui_tx_executor.rs
```diff
@@ -1412,35 +1485,7 @@ pub async fn build_register_or_update_validator_tx(
...
-                hashi_ids.package_id,
-                Identifier::from_static("validator"),
-                Identifier::from_static("update_next_epoch_encryption_public_key"),
-            ),
-            vec![hashi_arg, validator_address_arg, encryption_key_arg],
-        );
-        has_calls = true;
-    }
-
-    // 4. Update endpoint URL if available and changed.
```
bmwill, line 1492: its probably best to keep this here and instead of reading from the config, provide a way to lookup the "new" value that we want to update

bmwill: this way we don't need a whole new type of "update metadata" txn builder like you've added above and we can do the same thing for the BLS key update.

crates/hashi/src/mpc/service.rs
```diff
@@ -220,6 +220,17 @@ impl MpcService {
...
             }
+            match self.inner.next_reconfig_epoch().await {
+                Ok(target) => {
+                    if let Err(e) = self.inner.prepare_and_register_encryption_key(target).await {
+                        debug!(
+                            "Encryption key registration for epoch {target} failed: {e}; \
+                             will retry on next genesis_reconfig iteration"
+                        );
+                    }
+                }
```
bmwill, line 231: I think maybe we should split out the key rotation into its own task. It can monitor epoch change and make a decision on when/how to do the rotation. it doesn't need to be done inline with the mpc logic

crates/hashi/src/db.rs
zhouwfang: "Stale-tagged" is forward-looking — we will need to add an epoch tag to each on-chain registration so start_reconfig can reject ones whose tag doesn't match the upcoming epoch. Today the field is bare bytes with no tag, so start_reconfig can't tell a fresh registration from a stale one and copies whatever's there into the new committee. The +1 retention here keeps the previous epoch's private key in DB so find_encryption_key_matching can still resolve when a stale pub key gets captured.

crates/hashi/src/mpc/service.rs
zhouwfang: The genesis-loop and pre-start_reconfig registrations are intentionally inline — they gate execute_start_reconfig (95% on-chain threshold), so this validator's retry loop is what unblocks committee formation. Splitting them out creates a race where start_reconfig fires before this validator registers.                                                                            
                                                                                                                                   
The post-end_reconfig trigger is non-gating and could move into a separate monitor task — but it's already covered by the pre-reconfig retry on the next epoch, so splitting just one of the three adds a task for marginal benefit.

zhouwfang:
> we also need the same flows for the signing key, should we add them in this pr?

I created an issue for it and will do in a follow-up PR.

crates/hashi/src/db.rs
bmwill: So there are two approaches to this: 
- Explicit enforcement that you much rotate your keys each epoch otherwise you get kicked out
- Graceful rotating. The service handles the rotation but if a key wasn't rotated for some reason then the old encryption key will be used.

This have impacts on the actual reliability and security of our system and not necessarily in the ways you'd expect. I would personally prefer the graceful approach vs the hard explicit requirement of rotation. Main reason is that if an operator doesn't have Sui to do the rotation then we loose security/liveness of that node leading to concentration of stake (and key shares) in fewer nodes.

crates/e2e-tests/src/sui_network.rs
```diff
@@ -36,7 +36,7 @@ use tokio::time::sleep;
 
 const DEFAULT_NUM_VALIDATORS: usize = 4;
 const DEFAULT_EPOCH_DURATION_MS: u64 = 86_400_000; // 24 hours; tests that need epoch changes should set a shorter duration
-const NETWORK_STARTUP_TIMEOUT_SECS: u64 = 60;
+const NETWORK_STARTUP_TIMEOUT_SECS: u64 = 120;
```
bmwill: This is problematic, we should try to not increase the timeout and instead tighten up our startup process so it doesn't take as long

crates/hashi/src/db.rs
zhouwfang: SG, updated the comments.

crates/e2e-tests/src/sui_network.rs
zhouwfang: Reverted. Will dig into it later.

crates/hashi/src/lib.rs
```diff
@@ -449,14 +553,28 @@ impl Hashi {
...
+                    tracing::warn!(
+                        "Initial encryption key registration for epoch {next_epoch} \
+                         failed: {e}; will retry before next start_reconfig"
+                    );
+                }
+            }
+            Err(e) => {
+                tracing::warn!("Failed to compute next reconfig epoch: {e}");
+            }
+        }
```
bmwill: Can we do this calculation before the "execute_register_or_update" call? in particular the way this is written it requires two transactions instead of one.

bmwill (approved):

PR #504, opened 2026-04-30
deposit: introduce a deposit time delay

bmwill:


crates/hashi/src/leader/mod.rs
```diff
@@ -145,6 +151,10 @@ impl LeaderService {
                     self.process_signed_withdrawal_txns();
                     self.check_delete_proposals(checkpoint_timestamp_ms);
 
+                    // Reload from on-chain state so approved deposits are
+                    // picked up for the second-phase `confirm_deposit`
+                    // submission as soon as the time-delay window elapses.
+                    self.reload_pending_deposit_requests();
```
dlukegordon: I think this will change deposit retry cadence, now errors classified as `RetryOnNextBlock` will be retried on the next Sui checkpoint instead of waiting until the next Bitcoin block. Which will bring back the spammy erroring and retrying behavior.

packages/hashi/sources/btc/btc_config.move
```diff
@@ -105,6 +114,7 @@ public(package) fun set_withdrawal_cancellation_cooldown_ms(self: &mut Config, c
 /// Initialize BTC-specific config defaults. Called after config::create().
 public(package) fun init_defaults(config: &mut Config) {
     config.upsert(b"bitcoin_deposit_minimum", config_value::new_u64(30_000));
+    config.upsert(b"bitcoin_deposit_time_delay_ms", config_value::new_u64(1_000)); // TODO set this to a sane value
```
dlukegordon: Change to 10 mins?

dlukegordon (approved):

PR #505, opened 2026-04-30
Add per-epoch BLS signing key rotation with on-chain registration

zhouwfang:
#95 #163

### DIFF
--- a/crates/e2e-tests/src/sui_network.rs
+++ b/crates/e2e-tests/src/sui_network.rs
@@ -67,7 +67,7 @@ async fn wait_for_ready(client: &mut Client) -> Result<()> {
             .ledger_client()
             .get_service_info(sui_rpc::proto::sui::rpc::v2::GetServiceInfoRequest::default())
             .await
-            && resp.into_inner().checkpoint_height() > 5
+            && resp.into_inner().checkpoint_height() > 1
         {
             return Ok(());
         }
--- a/crates/hashi/src/db.rs
+++ b/crates/hashi/src/db.rs
@@ -12,6 +12,8 @@ use fjall::KeyspaceCreateOptions;
 use fjall::Result;
 use sui_sdk_types::Address;
 
+use hashi_types::committee::BLS12381PublicKey;
+use hashi_types::committee::Bls12381PrivateKey;
 use hashi_types::committee::EncryptionPrivateKey;
 use hashi_types::committee::EncryptionPublicKey;
 
@@ -28,6 +30,12 @@ pub struct Database {
     // key: big endian u64 for the epoch the key is used for
     // value: 32-byte RistrettoScalar
     encryption_keys: Keyspace,
+
+    // Column Family used to store BLS signing (protocol) keys.
+    //
+    // key: big endian u64 for the epoch the key is used for
+    // value: BCS-encoded `Bls12381PrivateKey`
+    signing_keys: Keyspace,
 
     // Column Family used to store dealer messages for DKG.
     //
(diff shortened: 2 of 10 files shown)


crates/hashi/src/lib.rs
```diff
@@ -238,14 +239,24 @@ impl Hashi {
...
-            .execute_register_or_update_validator(&self.config, None, Some(&public_key))
+            .execute_register_or_update_validator(
+                &self.config,
+                None,
+                Some(&enc_pub),
+                Some(sui_tx_executor::SigningKeyUpdate {
+                    public_key: &sig_pub,
+                    pop: &pop,
+                }),
+            )
```
bmwill: I would prefer we pass in the private key here so that this function can just use it to do the pop/epoch tests that it was already doing.

crates/hashi/src/sui_tx_executor.rs
```diff
@@ -1382,38 +1391,30 @@ pub async fn build_register_or_update_validator_tx(
         has_calls = true;
     }
 
-    // 2. Update BLS public key if available and changed.
-    if let Some(protocol_key) = config.protocol_private_key()
-        && onchain_member
+    // 2. Update next-epoch BLS public key if provided and changed.
+    if let Some(update) = next_epoch_signing_key {
```
bmwill: then this whole branch stays exactly the same

crates/e2e-tests/src/sui_network.rs
```diff
@@ -67,7 +67,7 @@ async fn wait_for_ready(client: &mut Client) -> Result<()> {
             .ledger_client()
             .get_service_info(sui_rpc::proto::sui::rpc::v2::GetServiceInfoRequest::default())
             .await
-            && resp.into_inner().checkpoint_height() > 5
+            && resp.into_inner().checkpoint_height() > 1
```
bmwill, line 70: why this change?

crates/hashi/src/db.rs
```diff
@@ -357,10 +427,11 @@ impl Database {
...
     /// Prune all MPC keyspaces.
     pub fn prune_messages_below(&self, cutoff_epoch: u64) -> Result<()> {
-        // Encryption keys retain one extra epoch to support graceful rotation: the
-        // committee may capture a previous epoch's pub key, and the matching
-        // private key must still be in DB.
+        // Encryption and signing keys retain one extra epoch to support
+        // graceful rotation: the committee may capture a previous epoch's
+        // pub key, and the matching private key must still be in DB.
         prune_keyspace(&self.encryption_keys, cutoff_epoch.saturating_sub(1))?;
+        prune_keyspace(&self.signing_keys, cutoff_epoch.saturating_sub(1))?;
```
bmwill: i'm actually wonder if it makes sense to not prune encryption keys and signing keys for a bit longer, just to be safe?

Also what value do we usually use for pruning in relation to the current epoch?

bmwill (approved):

crates/e2e-tests/src/sui_network.rs
zhouwfang: With `>5`, on slow CI it pushes the 4-validator default past the 60s startup timeout, flaking tests. `>1` still verifies checkpoints are progressing; downstream calls already block on their own checkpoints.

crates/hashi/src/db.rs
zhouwfang: We prune after submitting end reconfig with cutoff_epoch = target_epoch — the just-completed reconfig epoch. With < cutoff - 1 retention, we keep keys for current and current - 1 epochs. How much longer were you thinking?

bmwill: Just a few more epochs, say a weeks worth?

zhouwfang: SG, bumped to 7 extra epochs.

PR #506, opened 2026-05-01
Make emergency pause thresholds configurable via governance

Bridgerz:
## Summary
- Store emergency pause/unpause quorum thresholds in the on-chain config map so they can be tuned via `UpdateConfig` governance proposals without a package upgrade
- Replace the per-domain config validation functions (`is_valid_core_config_entry`, `btc_config::is_valid_config_entry`, `mpc_config::is_valid_config_entry`) with a single `is_valid_config_update` that checks the key exists and the new value's type matches the existing entry. This is safe because `init_defaults` establishes the schema — unknown keys are rejected since they don't exist in the map, and type mismatches are caught by `same_variant`. Per-field range checks are removed since the 2/3 committee vote is the real safety boundary, not validation logic. Adding a new config key now only requires a line in `init_defaults`.
- Add `same_variant` helper on `Value` enum for type-level comparison

## Test plan
- [x] All 84 Move tests pass
- [x] Rust crate builds cleanly
- [ ] Verify emergency pause proposal creation reads threshold from config
- [ ] Verify threshold can be changed via UpdateConfig proposal

### DIFF
--- a/packages/hashi/sources/btc/btc_config.move
+++ b/packages/hashi/sources/btc/btc_config.move
@@ -7,36 +7,9 @@ module hashi::btc_config;
 
 use hashi::{config::Config, config_value};
 
-// ======== Bitcoin Network Constants ========
-
 /// Minimum value (satoshis) for a Bitcoin output to be relayed (dust threshold).
 /// Uses the highest threshold (P2PKH 546 sats) as a conservative floor.
 const DUST_RELAY_MIN_VALUE: u64 = 546;
-
-// ======== Config Validation ========
-
-/// Returns true when `key` is a recognised BTC config key and `value`
-/// carries the type that key expects.
-#[allow(implicit_const_copy)]
-public(package) fun is_valid_config_entry(
-    key: &std::string::String,
-    value: &config_value::Value,
-): bool {
-    let k = key.as_bytes();
-    if (k == &b"bitcoin_deposit_minimum") {
-        value.is_u64()
-    } else if (k == &b"bitcoin_deposit_time_delay_ms") {
-        value.is_u64()
-    } else if (k == &b"bitcoin_withdrawal_minimum") {
-        value.is_u64()
-    } else if (k == &b"bitcoin_confirmation_threshold") {
-        value.is_u64()
-    } else if (k == &b"withdrawal_cancellation_cooldown_ms") {
-        value.is_u64()
-    } else {
-        false
-    }
-}
 
 // ======== Accessors ========
 
--- a/packages/hashi/sources/core/config/config.move
+++ b/packages/hashi/sources/core/config/config.move
@@ -25,7 +25,8 @@ const PAUSED_KEY: vector<u8> = b"paused";
 const GUARDIAN_URL_KEY: vector<u8> = b"guardian_url";
 const GUARDIAN_PUBLIC_KEY_KEY: vector<u8> = b"guardian_public_key";
 const EMERGENCY_PAUSE_THRESHOLD_BPS_KEY: vector<u8> = b"governance_emergency_pause_threshold_bps";
-const EMERGENCY_UNPAUSE_THRESHOLD_BPS_KEY: vector<u8> = b"governance_emergency_unpause_threshold_bps";
+const EMERGENCY_UNPAUSE_THRESHOLD_BPS_KEY: vector<u8> =
+    b"governance_emergency_unpause_threshold_bps";
 
 public struct Config has store {
     config: VecMap<String, Value>,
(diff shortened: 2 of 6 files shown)


packages/hashi/sources/btc/btc_config.move
```diff
@@ -7,37 +7,10 @@ module hashi::btc_config;
...
-        value.is_u64()
-    } else if (k == &b"bitcoin_withdrawal_minimum") {
-        value.is_u64()
-    } else if (k == &b"bitcoin_confirmation_threshold") {
-        value.is_u64()
-    } else if (k == &b"withdrawal_cancellation_cooldown_ms") {
-        value.is_u64()
-    } else {
-        false
-    }
```
bmwill, line 38: one risk, maybe its acceptable though, is that some values we explicitly omitted from here. specifically the chain id for bitcoin. With this PR we can now technically change it, which is maybe a good idea in the event of a bitcoin fork?

packages/hashi/sources/core/config/config.move
```diff
@@ -56,16 +59,11 @@ public(package) fun upsert(self: &mut Config, key: vector<u8>, value: Value) {
...
-    let k = key.as_bytes();
-    if (k == &PAUSED_KEY) {
-        value.is_bool()
-    } else {
-        false
-    }
+/// Returns true when `key` exists in the config and `value` has the
+/// same type as the existing entry.
+public(package) fun is_valid_config_update(self: &Config, key: &String, value: &Value): bool {
+    if (!self.config.contains(key)) return false;
```
bmwill, line 67: If we ever add new config values in the future this wouldn't work, although maybe we could just have a new "insert_new_config" proposal or something that would be able to init those new fields.

packages/hashi/sources/btc/btc_config.move
Bridgerz: Ya I thought about this too. the bitcoin chain id is really the only config value that should never change.... tho it does require 6667 stake and at that point we're open to a lot more. I think we should be comfortable with this.

packages/hashi/sources/core/config/config.move
Bridgerz: I think a new config value would most likely mean a change to the code and therefore an upgrade so I think we can cross that bridge when we get there.

packages/hashi/sources/core/mpc_config.move
```diff
@@ -7,29 +7,10 @@ use hashi::{config::Config, config_value};
...
-    key: &std::string::String,
-    value: &config_value::Value,
-): bool {
-    let k = key.as_bytes();
-    if (k == &b"mpc_threshold_in_basis_points") {
-        value.is_u64() && (*value).as_u64() > 0 && (*value).as_u64() <= MAX_BPS
-    } else if (k == &b"mpc_weight_reduction_allowed_delta") {
-        value.is_u64() && (*value).as_u64() <= MAX_BPS
-    } else if (k == &b"mpc_max_faulty_in_basis_points") {
-        value.is_u64() && (*value).as_u64() <= MAX_BPS
```
bmwill, line 27: We're also loosing some of the checks on the acceptable ranges for these.

packages/hashi/sources/core/config/config.move
```diff
@@ -22,6 +22,9 @@ const EVersionDisabled: vector<u8> = b"Version disabled";
 const EDisableCurrentVersion: vector<u8> = b"Cannot disable current version";
 
 const PAUSED_KEY: vector<u8> = b"paused";
+const EMERGENCY_PAUSE_THRESHOLD_BPS_KEY: vector<u8> = b"governance_emergency_pause_threshold_bps";
+const EMERGENCY_UNPAUSE_THRESHOLD_BPS_KEY: vector<u8> =
+    b"governance_emergency_unpause_threshold_bps";
```
bmwill, line 29: I wonder if it would make sense/be more extendable, to have the bps thresholds be stored in their own map like structure then we can use type-keys to fetch them... we can ensure they're in a valid range (like the mpc ones).

Not sure yet so maybe we sit and think about it for a bit.

Bridgerz: I think 6667 quorum is the real guard here. Yes this would enable preventing "accidental" out of boundary BPS values, but I think the complexity trade off may not be worth it.

bmwill (approved):

packages/hashi/sources/core/mpc_config.move
zhouwfang: These are useful to have, and this removal prevents us from adding additional check for nonce generation protocol config param in #706.

zhouwfang: Restored in #710.

PR #508, opened 2026-05-01
Add on chain Guardian management

Bridgerz:
## Summary
- New `UpdateGuardian` proposal type that atomically stores a guardian URL and public key in the on-chain core config
- Guardian config accessors (`guardian_url`, `guardian_public_key`, `set_guardian`) added to `config.move` using `try_get` (guardian is optional, not initialized at genesis)
- Rust side reads guardian endpoint from on-chain config first, falling back to node config file
- Full CLI support: `hashi proposal create update-guardian --url <url> --public-key <hex>`

## Test plan
- [ ] Move package builds cleanly
- [ ] Rust `cargo check` passes
- [ ] Verify `update_guardian::propose` and `update_guardian::execute` work in e2e tests
- [ ] Verify guardian client initialization reads from on-chain config when available
- [ ] Verify fallback to node config when no on-chain guardian config exists

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -464,6 +464,13 @@ pub struct AbortReconfig {
     pub epoch: u64,
 }
 
+/// Rust version of the Move hashi::update_guardian::UpdateGuardian type.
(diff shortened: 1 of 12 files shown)


bmwill (approved):
thanks!

PR #513, opened 2026-05-05
Increase withdrawal batch capacity

Bridgerz:
## Summary

- Decouple `confirm_withdrawal` from UTXO count by switching from `confirm_spent` (2 objects/input) to `mark_spent` (event-only, 0 objects/input). Actual record cleanup is deferred to a new `cleanup_spent_utxos` background GC task.
- Raise batch limits to 70 requests / 700 inputs (922 commit objects, 7.8% margin below Sui's 1000-object hard cap).
- Add orphan discovery at startup so crashed nodes recover pending cleanups.

## How it works

The withdrawal confirm path was the binding constraint — each input touched 2 dynamic field objects for `confirm_spent`. With the two-phase approach:

1. **Confirm** calls `mark_spent` which emits a `UtxoSpentEvent` but touches zero objects per input. The watcher updates local state from the event.
2. **Cleanup** runs as a background GC task (like deposit/proposal GC), calling `cleanup_spent_utxos` which moves records from `utxo_records` to `spent_utxos`. Chunked at 400 inputs/tx.
3. **Orphan discovery** scans `utxo_records` for UTXOs locked by withdrawals not in the active set, catching any cleanups lost to crashes.

## Commit object model

```
runtime_objects = 1 * inputs + 3 * requests + 12
```

At max capacity: `1 × 700 + 3 × 70 + 12 = 922` (7.8% below 1000).

Confirm is no longer input-bound: `requests + ~43` fixed overhead.

## Test plan

- [x] All 84 Move tests pass
- [x] All 296 Rust unit tests pass
- [x] Withdrawal budget tests updated for new constants (70 req / 700 inputs / 922 budget)
- [x] E2E stress test asserts commit, sign, and confirm stay under all Sui limits (tx size, effects size, runtime objects)
- [x] E2E stress test runs `cleanup_spent_utxos` after confirm

### DIFF
--- a/crates/hashi/src/sui_tx_executor.rs
+++ b/crates/hashi/src/sui_tx_executor.rs
@@ -92,14 +92,46 @@ fn build_chunked_vec_vec_u8_arg(
         let chunk_arg = builder.pure(&chunk);
         builder.move_call(
             Function::new(
-                Address::new([
-                    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-                    0, 0, 0, 0, 0, 1,
-                ]),
+                MOVE_STDLIB_ADDRESS,
                 Identifier::from_static("vector"),
                 Identifier::from_static("append"),
             )
             .with_type_args(vec![vec_u8_type.clone()]),
+            vec![combined, chunk_arg],
+        );
+    }
+
+    combined
+}
+
+/// Maximum number of arguments to pass to a single `make_move_vec` command.
+///
+/// Sui enforces a 512-argument limit per PTB command. Keep this well below the
+/// ceiling so large withdrawal commits can assemble vectors of custom Move
+/// structs without making one command consume hundreds of arguments.
+const MAX_MOVE_VEC_ARGS_PER_CHUNK: usize = 250;
+
+/// Build a `vector<T>` PTB argument from existing element arguments, chunking
+/// the `make_move_vec` calls so no command gets close to Sui's 512-argument
+/// ceiling. Later chunks are appended into the first one with
+/// `0x1::vector::append<T>`.
+fn build_chunked_move_vec_arg(
+    builder: &mut TransactionBuilder,
+    elements: Vec<sui_transaction_builder::Argument>,
+    element_type: TypeTag,
+) -> sui_transaction_builder::Argument {
+    let mut iter = elements.chunks(MAX_MOVE_VEC_ARGS_PER_CHUNK);
+    let first_chunk = iter.next().unwrap_or_default().to_vec();
+    let combined = builder.make_move_vec(Some(element_type.clone()), first_chunk);
+    for chunk in iter {
+        let chunk_arg = builder.make_move_vec(Some(element_type.clone()), chunk.to_vec());
+        builder.move_call(
+            Function::new(
+                MOVE_STDLIB_ADDRESS,
+                Identifier::from_static("vector"),
+                Identifier::from_static("append"),
+            )
+            .with_type_args(vec![element_type.clone()]),
             vec![combined, chunk_arg],
         );
     }
@@ -183,9 +215,13 @@ use crate::onchain;
 use crate::onchain::OnchainState;
 use crate::onchain::types::DepositConfirmationMessage;
 use crate::onchain::types::DepositRequest;
+use crate::onchain::types::UtxoId;
 use crate::withdrawals::WithdrawalTxCommitment;
 
 const DEFAULT_TIMEOUT_SECS: u64 = 10;
+
+/// Well-known Move stdlib package address (0x1)
+const MOVE_STDLIB_ADDRESS: Address = Address::from_static("0x1");
 
 /// Well-known Sui Clock object address (0x6)
 pub const SUI_CLOCK_OBJECT_ID: Address = Address::from_static("0x6");
@@ -1113,7 +1149,8 @@ impl SuiTxExecutor {
                 )
             })
             .collect();
-        let selected_utxos_arg = builder.make_move_vec(Some(utxo_id_type.into()), utxo_elements);
+        let selected_utxos_arg =
+            build_chunked_move_vec_arg(&mut builder, utxo_elements, utxo_id_type.into());
 
         let output_utxo_type = StructTag::new(
             self.hashi_ids.package_id,
@@ -1137,7 +1174,8 @@ impl SuiTxExecutor {
                 )
             })
             .collect();
-        let outputs_arg = builder.make_move_vec(Some(output_utxo_type.into()), output_elements);
+        let outputs_arg =
+            build_chunked_move_vec_arg(&mut builder, output_elements, output_utxo_type.into());
 
         let txid_arg = builder.pure(&approval.txid);
         let cert_arg = build_committee_signature_arg(&mut builder, self.hashi_ids.package_id, cert);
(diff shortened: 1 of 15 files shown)


crates/hashi/src/utxo_pool/mod.rs
```diff
@@ -348,7 +348,13 @@ impl CoinSelectionParams {
...
     /// constraint. Larger batches amortize the fixed transaction
     /// overhead across more requests, reducing per-request fees.
-    pub const DEFAULT_MAX_WITHDRAWAL_REQUESTS: usize = 50;
+    pub const DEFAULT_MAX_WITHDRAWAL_REQUESTS: usize = Self::ABSOLUTE_MAX_WITHDRAWAL_REQUESTS;
+
+    /// Absolute maximum number of withdrawal requests per batch.
+    ///
+    /// Larger configured values are clamped because Sui runtime object limits
+    /// make request count and selected input count a coupled constraint.
+    pub const ABSOLUTE_MAX_WITHDRAWAL_REQUESTS: usize = 130;
```
bmwill: Can we adjust this down with the assumption that every 1 request can be expected to account for 10 inputs? Thats mostly how i had expected the transaction construction logic to work wrt limits. We can change that of course if we want a different ratio.

crates/hashi/src/sui_tx_executor.rs
```diff
@@ -105,6 +105,74 @@ fn build_chunked_vec_vec_u8_arg(
...
+    let combined = builder.pure(&first_chunk);
+
+    for chunk in chunks {
+        let chunk_arg = builder.pure(&chunk.to_vec());
+        builder.move_call(
+            Function::new(
+                Address::new([
+                    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 1,
+                ]),
```
bmwill: Can we use Address::ZERO?

crates/hashi/src/sui_tx_executor.rs
```diff
@@ -105,6 +105,74 @@ fn build_chunked_vec_vec_u8_arg(
...
+    let first_chunk = iter.next().unwrap_or_default().to_vec();
+    let combined = builder.make_move_vec(Some(element_type.clone()), first_chunk);
+    for chunk in iter {
+        let chunk_arg = builder.make_move_vec(Some(element_type.clone()), chunk.to_vec());
+        builder.move_call(
+            Function::new(
+                Address::new([
+                    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                    0, 0, 0, 0, 0, 1,
+                ]),
```
bmwill: same

crates/hashi/src/withdrawals.rs
```diff
@@ -51,6 +51,16 @@ const WITHDRAWAL_SIGNING_TIMEOUT: Duration = Duration::from_secs(5);
...
 
+/// Conservative runtime-object budget for withdrawal commit transactions.
+///
+/// Sui's hard object-runtime cache limit is 1000 objects. Empirical
+/// measurements put withdrawal commit transactions at roughly
+/// `selected_utxos + 3 * requests + fixed_overhead` runtime objects, so target
+/// 800 to leave headroom for metadata/runtime drift.
+const WITHDRAWAL_COMMIT_RUNTIME_OBJECT_BUDGET: usize = 800;
+const WITHDRAWAL_COMMIT_FIXED_RUNTIME_OBJECTS: usize = 12;
+const WITHDRAWAL_COMMIT_RUNTIME_OBJECTS_PER_REQUEST: usize = 3;
```
bmwill, line 72: Did we do an analysis on how finalizing the withdrawal would work? I mean we have that extra onchain indexing that we also do and i wonder how all of that plays into this. I *think* we should have more headroom during confirm because we do less operations but still worth double checking the limits there and ensuring they fit into the limits we define here.

crates/hashi/src/withdrawals.rs
```diff
@@ -742,16 +760,9 @@ impl Hashi {
...
-        // max_fee_per_request to match the Move contract's worst-case cap,
-        // and max_withdrawal_requests to honour the leader's configured
-        // batch cap.
-        let params = CoinSelectionParams {
-            max_fee_per_request: self.onchain_state().worst_case_network_fee(),
-            max_withdrawal_requests: self.config.withdrawal_max_batch_size(),
-            max_mempool_chain_depth: self.config.max_mempool_chain_depth(),
-            ..CoinSelectionParams::new(change_address.clone())
-        };
+        let configured_max_inputs = CoinSelectionParams::DEFAULT_MAX_INPUTS;
```
bmwill, line 822: I think this is just 500 for now but if we increase requests from 50 to 130 then this will need to increase as well.

In generally the limiting factor i'm mostly worried about is the # of inputs since we want aggressively smash utxos if we can. so just raising the request count isn't interesting, mostly because it could result in those withdrawals contributing less to pool management than they otherwise would.

crates/hashi/src/utxo_pool/mod.rs
Bridgerz: Ah got it. How conservative is 10 inputs per 1 request? If we wanted to stay at 10, we could move up from 50 to 60: 
12 + 3R + 10R <= 800 (80% of 1000 runtime objects) 
R <= ~60

crates/hashi/src/withdrawals.rs
Bridgerz: Good point... ok if we raise from 50 requests to 60 requests (for the 10 inputs per request) we can raise the inputs from 500 to 600. Not the doubling of request processing I was expecting, but a 20% increase none the less.... I'll take it.

crates/hashi/src/withdrawals.rs
Bridgerz: wrench thrown. Going back to the drawing board.

crates/hashi/src/utxo_pool/mod.rs
bmwill: do we need to reduce this from 130 since we upped inputs to 600?

packages/hashi/sources/btc/utxo_pool.move
```diff
@@ -103,6 +103,26 @@ public(package) fun confirm_spent(self: &mut UtxoPool, utxo_id: UtxoId, epoch: u
     sui::event::emit(UtxoSpentEvent { utxo_id, spent_epoch: epoch });
 }
```
bmwill: Is this still used after this PR? Or can we call this from "cleanup_spent"?

packages/hashi/sources/btc/utxo_pool.move
```diff
@@ -103,6 +103,26 @@ public(package) fun confirm_spent(self: &mut UtxoPool, utxo_id: UtxoId, epoch: u
...
+
+/// Deferred bookkeeping for a spent UTXO: remove from `utxo_records` and
+/// record in `spent_utxos`. Safe to call in a separate transaction after
+/// `mark_spent`. No-ops if the record has already been cleaned up.
+public(package) fun cleanup_spent(self: &mut UtxoPool, utxo_id: UtxoId, epoch: u64) {
+    if (self.utxo_records.contains(utxo_id)) {
+        let UtxoRecord { utxo, produced_by: _, locked_by: _ } = self.utxo_records.remove(utxo_id);
+        utxo.delete();
+        self.spent_utxos.add(utxo_id, epoch);
+        sui::event::emit(UtxoSpentEvent { utxo_id, spent_epoch: epoch });
```
bmwill: so now we're emitting the event twice, once in `mark_spent` (which just emits an event) and then again here when we clean it up. We probably want to remove one of these emits?

Bridgerz:
This is ready for another round when you've got a chance @bmwill

packages/hashi/sources/btc/withdraw.move
```diff
@@ -302,6 +304,25 @@ entry fun confirm_withdrawal(hashi: &mut Hashi, withdrawal_id: address, cert: Co
...
+    let input_ids = hashi.bitcoin().withdrawal_queue().confirmed_txn_input_ids(withdrawal_id);
+    let epoch = hashi.committee_set().epoch();
+    let len = input_ids.length();
+    let end = if (start + count < len) { start + count } else { len };
+    let mut i = start;
+    while (i < end) {
+        hashi.bitcoin_mut().utxo_pool_mut().cleanup_spent(input_ids[i], epoch);
+        i = i + 1;
+    };
+}
```
bmwill: what are your thoughts on having this be a bit more flexible, that is when we confirm a withdrawal we actually mark the utxos and then this can take a marked utxo and clean it up instead of needing to grab a withdrawal and then walk the utxos it used

Bridgerz: Marking the UTXOs themselves would cost runtime objects during confirm, which is what we're trying to avoid, but this would be a "cleaner cleanup" approach (more stateless). Good idea.

bmwill: yes it would touch more objects, but not more than is required when we do txn construction i believe?

Bridgerz: I think you're right.

packages/hashi/sources/btc/utxo_pool.move
```diff
@@ -94,15 +97,26 @@ public(package) fun get_utxo(self: &UtxoPool, utxo_id: UtxoId): hashi::utxo::Utx
...
+/// been cleaned up.
+public(package) fun cleanup_spent(self: &mut UtxoPool, utxo_id: UtxoId) {
+    if (self.utxo_records.contains(utxo_id)) {
+        let UtxoRecord { utxo, produced_by: _, locked_by: _, spent_epoch } = self
+            .utxo_records
+            .remove(utxo_id);
+        utxo.delete();
+        self.spent_utxos.add(utxo_id, spent_epoch.destroy_or!(0));
+    };
+}
```
bmwill, line 118: There isn't any check that spent_epoch is some since you're doing destroy_or. We only want to allow cleaning up spent utxos iif they've been confirmed spent.

Bridgerz: good catch

bmwill (approved):

PR #522, opened 2026-05-07
[5/n][guardian-integration] change WithdrawalID type from u64 to Address

0xsiddharthks:
## Context

A withdrawal already has a unique id on Sui (32 byte tx object id). However, the guardian is currently using a separate `wid` ( added in #140 and #189 ) that is of type u64.

We want to use `wid` as the cache key later on in the guardian proxy cache.

This PR unifies the two types, and updates the guardian `WithdrawalID` to `sui_sdk_types::Address`.

## Changes
- `WithdrawalID` is now `sui_sdk_types::Address` (32-byte BCS layout, `0x`-prefixed Display).
- Proto wire flips from `uint64` to `bytes`.
- Audit/RPC tracing and the withdrawal log path use `Address`'s Display directly — no more manual `hex::encode` shims, and `hex` is dropped from `hashi-guardian`.


crates/hashi-guardian/src/withdraw.rs
```diff
@@ -164,37 +167,37 @@ pub fn verify_hashi_cert<T: Serialize>(
 
 async fn log_withdrawal_success(
     enclave: &Enclave,
-    wid: u64,
+    wid_hex: &str,
```
bmwill: the wid is an `Address`, can we just use that type directly? It already handles the hex display and everything

crates/hashi-types/src/guardian/mod.rs
```diff
@@ -280,9 +280,9 @@ pub enum WithdrawalLogMessage {
 //      Helper types & structs
 // ---------------------------------
 
-/// Unique identifier for a withdrawal request
-/// It is used to correlate events across sui & guardian.
-pub type WithdrawalID = u64;
+/// 32-byte UID of the on-chain `WithdrawalTransaction` Sui object.
+/// Used to correlate events across Sui, hashi nodes, and the guardian.
+pub type WithdrawalID = [u8; 32];
```
bmwill: so just make this change
```suggestion
/// 32-byte UID of the on-chain `WithdrawalTransaction` Sui object.
/// Used to correlate events across Sui, hashi nodes, and the guardian.
pub type WithdrawalID = Address;
```
and then you can remove all the hex encoding logic added here, simplifying a bit

bmwill (changes requested):
Only thing to change is making WithdrawalId an `Address`

0xsiddharthks:
@bmwill , thanks! i switched `WithdrawalID` to `sui_sdk_types::Address`

bmwill (approved):

PR #523, opened 2026-05-07
[6/n][guardian-integration] anchor local limiter to on-chain WithdrawalSignedEvent

0xsiddharthks:
## Summary

In #495 , we started confirming with the `LocalLimiter` whenever we did MPC signing per node. 

However, we need to update the `LocalLimiter` whenever the transaction is submitted onchain. For that, we are using the `WithdrawalSignedEvent` as the anchor. Every node observes it from the same Sui checkpoint stream, so leader and follower advance in lockstep.

## Note
@bmwill , we are not currently doing a 'soft reserve' vs 'hard reserve' on the local limiter. As of now, we are validating with the LocalLimiter at every MPC signing event, and then updating the LocalLimiter on every WithdrawalSignedEvent .

This should work well, since we are using MPC concurrency cap of 1 when the guardian is configured. 

TODO: Later on I'll be adding a soft reserve per `WithdrawalID`.

## Changes
- New `start_guardian_limiter_observer` on every node. This listens for the new `Notification::WithdrawalSigned` and calls `apply_consume`.
- Watcher emits `WithdrawalSigned` with the limiter inputs (amount, timestamp) so the observer doesn't re-read the mirror.
- withdrawal selection is capacity-aware: leader truncates the approved batch to the longest timestamp-sorted prefix that fits the limiter.
- MPC concurrency capped at 1 when the limiter is in play.
- `wait_until_signed_committed` blocks the leader against itself so the next checkpoint can't respawn before the prior withdrawal is observed.
- e2e tests: `wait_for_local_limiter` helper so tests don't race observer bootstrap.

### DIFF
--- a/crates/e2e-tests/src/hashi_network.rs
+++ b/crates/e2e-tests/src/hashi_network.rs
@@ -162,10 +162,7 @@ impl HashiNodeHandle {
         }
     }
 
-    pub(crate) async fn wait_for_local_limiter(
-        &self,
-        timeout: std::time::Duration,
-    ) -> Result<()> {
+    pub(crate) async fn wait_for_local_limiter(&self, timeout: std::time::Duration) -> Result<()> {
         tokio::time::timeout(timeout, self.wait_for_local_limiter_inner())
             .await
             .map_err(|_| anyhow::anyhow!("local limiter bootstrap timed out after {:?}", timeout))
--- a/crates/e2e-tests/src/hashi_network.rs
+++ b/crates/e2e-tests/src/hashi_network.rs
@@ -162,6 +162,21 @@ impl HashiNodeHandle {
         }
     }
 
+    pub async fn wait_for_local_limiter(&self, timeout: std::time::Duration) -> Result<()> {
+        tokio::time::timeout(timeout, self.wait_for_local_limiter_inner())
+            .await
+            .map_err(|_| anyhow::anyhow!("local limiter bootstrap timed out after {:?}", timeout))
+    }
+
+    async fn wait_for_local_limiter_inner(&self) {
+        loop {
+            if self.hashi().local_limiter().is_some() {
+                return;
+            }
+            tokio::time::sleep(POLL_INTERVAL).await;
+        }
+    }
+
     pub fn current_epoch(&self) -> Option<u64> {
         self.hashi()
             .onchain_state_opt()
--- a/crates/e2e-tests/src/hashi_network.rs
+++ b/crates/e2e-tests/src/hashi_network.rs
@@ -162,7 +162,10 @@ impl HashiNodeHandle {
         }
     }
 
-    pub async fn wait_for_local_limiter(&self, timeout: std::time::Duration) -> Result<()> {
+    pub(crate) async fn wait_for_local_limiter(
+        &self,
+        timeout: std::time::Duration,
+    ) -> Result<()> {
         tokio::time::timeout(timeout, self.wait_for_local_limiter_inner())
             .await
             .map_err(|_| anyhow::anyhow!("local limiter bootstrap timed out after {:?}", timeout))
--- a/crates/hashi-types/src/guardian/bitcoin_utils.rs
+++ b/crates/hashi-types/src/guardian/bitcoin_utils.rs
@@ -344,6 +344,23 @@ impl TxUTXOs {
             .sum()
     }
 
+    /// BTC that leaves the pool when this txn broadcasts: `inputs - change`,
+    /// equivalent to `external_out_amount + miner_fee`. The amount that
+    /// consumes the rate-limiter (miner fee leaves the pool too; change
+    /// flows back).
+    pub fn gross_outflow_amount(&self) -> Amount {
+        let inputs: Amount = self.inputs.iter().map(|i| i.amount).sum();
+        let internal: Amount = self
+            .outputs
+            .iter()
+            .filter_map(|utxo| match utxo {
+                OutputUTXO::Internal(x) => Some(x.amount),
+                OutputUTXO::External(_) => None,
+            })
+            .sum();
+        inputs - internal
+    }
+
     /// Constructs an unsigned Bitcoin transaction for these UTXOs.
     fn unsigned_tx(
         &self,
--- a/crates/hashi/src/leader/mod.rs
+++ b/crates/hashi/src/leader/mod.rs
@@ -1446,7 +1446,9 @@ impl LeaderService {
         const VISIBILITY_TIMEOUT: Duration = Duration::from_secs(30);
         if tokio::time::timeout(
             VISIBILITY_TIMEOUT,
-            inner.onchain_state().wait_until_checkpoint(signed_checkpoint),
+            inner
+                .onchain_state()
+                .wait_until_checkpoint(signed_checkpoint),
         )
         .await
         .is_err()
--- a/crates/hashi/src/leader/mod.rs
+++ b/crates/hashi/src/leader/mod.rs
@@ -998,7 +998,7 @@ impl LeaderService {
         // so we don't sit on excess demand for the full batching window.
         let (batch, at_capacity) = if let Some(limiter) = self.inner.local_limiter() {
             let timestamp_secs = checkpoint_timestamp_ms / 1000;
-            let capacity = limiter.capacity_at(timestamp_secs).await;
+            let capacity = limiter.capacity_at(timestamp_secs);
             let mut cumulative = 0u64;
             let mut take_count = 0;
             for req in &approved {
@@ -1306,11 +1306,8 @@ impl LeaderService {
         let expected_limiter_seq = if let Some(limiter) = inner.local_limiter() {
             let amount_sats: u64 = txn.withdrawal_outputs.iter().map(|o| o.amount).sum();
             let timestamp_secs = txn.timestamp_ms / 1000;
-            let next_seq = limiter.next_seq().await;
-            if let Err(e) = limiter
-                .validate_consume(next_seq, timestamp_secs, amount_sats)
-                .await
-            {
+            let next_seq = limiter.next_seq();
+            if let Err(e) = limiter.validate_consume(next_seq, timestamp_secs, amount_sats) {
                 warn!(
                     withdrawal_txn_id = %txn.id,
                     "Leader local limiter rejected withdrawal; will retry on next checkpoint: {e}"
@@ -2010,7 +2007,7 @@ impl LeaderService {
         let wait = async {
             loop {
                 let committed = match (inner.local_limiter(), leader_seq_used) {
-                    (Some(limiter), Some(seq)) => limiter.next_seq().await > seq,
+                    (Some(limiter), Some(seq)) => limiter.next_seq() > seq,
                     _ => inner
                         .onchain_state()
                         .withdrawal_txn(txn_id)
--- a/crates/hashi/src/leader/mod.rs
+++ b/crates/hashi/src/leader/mod.rs
@@ -148,8 +148,7 @@ impl LeaderService {
                     }
 
                     self.process_unapproved_withdrawal_requests(checkpoint_timestamp_ms);
-                    self.process_approved_withdrawal_requests(checkpoint_timestamp_ms)
-                        .await;
+                    self.process_approved_withdrawal_requests(checkpoint_timestamp_ms);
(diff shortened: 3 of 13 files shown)


crates/hashi/src/onchain/watcher.rs
```diff
@@ -437,14 +437,29 @@ async fn handle_events(client: &mut Client, state: &OnchainState, events: &[Hash
...
+                    let mut state = state.state_mut();
+                    state
+                        .hashi
+                        .withdrawal_queue
+                        .withdrawal_txns
+                        .get_mut(&event.withdrawal_txn_id)
+                        .map(|txn| {
+                            txn.signatures = Some(event.signatures.clone());
+                            let amount_sats: u64 =
+                                txn.withdrawal_outputs.iter().map(|o| o.amount).sum();
```
bmwill: This probably isn't the correct way to calculate this, and we probably need to also double check that we do this calculation properly on the validating side when we commit to a txn.

The amount leaving the pool is the total gross request amount, not the final net (minus miner fees) that actually ends up in the output. In particular we also need to properly handle that some of the outputs are change that flow back into the pool and we can't count that against the limit. So probably worth having a single function that does this calculation that we can reuse here as well as in the txn construction/validation section.

crates/e2e-tests/src/hashi_network.rs
```diff
@@ -162,6 +162,21 @@ impl HashiNodeHandle {
...
+    }
+
+    async fn wait_for_local_limiter_inner(&self) {
+        loop {
+            if self.hashi().local_limiter().is_some() {
+                return;
+            }
+            tokio::time::sleep(POLL_INTERVAL).await;
+        }
+    }
```
bmwill: We do we need to wait for this to be spawned?

crates/hashi/src/leader/mod.rs
```diff
@@ -936,7 +939,7 @@ impl LeaderService {
     // Step 2: Construct withdrawal tx for approved requests
     // ========================================================================
 
-    fn process_approved_withdrawal_requests(&mut self, checkpoint_timestamp_ms: u64) {
+    async fn process_approved_withdrawal_requests(&mut self, checkpoint_timestamp_ms: u64) {
```
bmwill: This function cannot be make async, and it cannot do any major async work in order to avoid the leader event loop stalling for any reason. For that purpose we may actually be better off wrapping the limiter in a RwLock and which would allow us to read the value here properly without needing to resort to async.

crates/hashi/src/leader/mod.rs
```diff
@@ -986,10 +989,50 @@ impl LeaderService {
...
+                let stuck = &approved[0];
+                if self.stuck_withdrawal_warned.insert(stuck.id) {
+                    warn!(
+                        request_id = %stuck.id,
+                        btc_amount = stuck.btc_amount,
+                        available = capacity,
+                        "Withdrawal request exceeds local limiter capacity; staying in queue"
+                    );
+                }
+                return;
```
bmwill: Lets add a TODO (or maybe even just implement it) to handle the exceptional case to skip a request which has an amount that is larger than the max bucket size and increment a metric that we can use to alert on (in addition to the log). This way a single, very large output that would never be able to flow out, gets skipped so other requests in the queue get to drain.

unsticking a large one would either require the user canceling and doing a few smaller txns, or us upping the limit.

crates/hashi/src/leader/mod.rs
```diff
@@ -986,10 +989,50 @@ impl LeaderService {
...
             );
 
-        if batch.is_empty() {
+        if approved.is_empty() {
             return;
         }
 
+        // Truncate to the longest timestamp-sorted prefix that fits the local
+        // limiter's available capacity; the dropped tail flips `at_capacity`
+        // so we don't sit on excess demand for the full batching window.
```
bmwill: This is great, effectively we only let through the requests that would fit in the existing limit.

crates/hashi/src/leader/mod.rs
```diff
@@ -1196,7 +1239,14 @@ impl LeaderService {
...
 
-        let max_concurrent = self.inner.config.max_concurrent_leader_job_tasks();
+        // Cap to 1 when the limiter is in play: concurrent MPC tasks would
+        // race on `expected_limiter_seq` since the observer bumps it on every
+        // signed event.
+        let max_concurrent = if self.inner.guardian_client().is_some() {
+            1
+        } else {
+            self.inner.config.max_concurrent_leader_job_tasks()
+        };
```
bmwill, line 1264: I wonder if we should also add a check to the txn construction flow to only build a new one (as well as approve it) if there doesn't exist another unsigned withdrawal txn.

crates/hashi/src/leader/mod.rs
```diff
@@ -1362,6 +1412,11 @@ impl LeaderService {
                 .inc();
         })?;
 
+        // Block before returning so the next checkpoint can't respawn this
+        // task with a stale `next_seq` (guardian) or double-submit
+        // `sign_withdrawal` (no-guardian).
+        Self::wait_until_signed_committed(&inner, &txn.id, expected_limiter_seq).await;
```
bmwill: I think our txn executor waits for checkpoint inclusion, if it doesn't maybe we should add an option for that, and then this shouldn't really be needed. or rather then we could take the checkpoint # we get from the txn submission (if it was successful of course) and then just wait till our local checkpoint view catches up. This would eliminate most of this waiting around function i think.

crates/hashi/src/lib.rs
```diff
@@ -741,6 +743,67 @@ impl Hashi {
         })
     }
 
+    /// Sole mutator of `LocalLimiter` after bootstrap: advances on every
+    /// observed `WithdrawalSignedEvent`. Symmetric across leader and
+    /// follower since every node consumes the same Sui checkpoint stream.
+    fn start_guardian_limiter_observer(self: Arc<Self>) -> Service {
```
bmwill: What if we just had the watcher update the confirmed limit directly, it would be one less event loop to deal with possibly, thoughts?

crates/hashi/src/onchain/watcher.rs
0xsiddharthks: gotcha. yeah I was previously matching the implementation in `crates/hashi-guardian/src/withdraw.rs`.

Just fixed both this and the guardian code.

I added a new `withdrawal_limiter_consumption_amount` helper for this.
Now both this anchor side and the MPC signing validate path in `mpc_sign_withdrawal_tx` go through this helper.

crates/e2e-tests/src/hashi_network.rs
0xsiddharthks: The bootstrap is an async background task per node (it makes a `GetGuardianInfo` RPC to the guardian and seeds `LocalLimiter`), so there's a window between node start and the limiter being set. Tests that issue withdrawals immediately would race that window.

But I did a refactor just now and moved it into a `finalize_guardian_harness` in the e2e harness so it runs once during `with_guardian()` setup.

crates/hashi/src/leader/mod.rs
0xsiddharthks: got it.
yeah, i just made this change.

crates/hashi/src/leader/mod.rs
0xsiddharthks: gotcha. just implemented this.

crates/hashi/src/leader/mod.rs
0xsiddharthks: Yeah, the executor already waits for checkpoint inclusion in `execute_transaction_and_wait_for_checkpoint` . 
Just refactored the flow.
- `execute_sign_withdrawal` now returns `Result<u64>` (the checkpoint seq).
- `submit_sign_withdrawal` propagates it.
- The `wait_until_signed_committed` function is now replaced with one call to a new `OnchainState::wait_until_checkpoint(seq)` which loops on `subscribe_checkpoint().changed()` until checkpoint is reached.

crates/hashi-types/src/guardian/bitcoin_utils.rs
```diff
@@ -344,6 +344,23 @@ impl TxUTXOs {
             .sum()
     }
 
+    /// BTC that leaves the pool when this txn broadcasts: `inputs - change`,
+    /// equivalent to `external_out_amount + miner_fee`. The amount that
+    /// consumes the rate-limiter (miner fee leaves the pool too; change
+    /// flows back).
+    pub fn gross_outflow_amount(&self) -> Amount {
```
bmwill, line 351: i wonder if we should cross check the request amounts equal this outflow, maybe we already have that though?

crates/hashi/src/leader/mod.rs
```diff
@@ -1339,15 +1401,15 @@ impl LeaderService {
 
         // 4. Submit sign_withdrawal to Sui (writes signatures on-chain).
         // Broadcast + confirm happens via process_signed_withdrawal_txns on the next tick.
-        Self::submit_sign_withdrawal(
+        let signed_checkpoint = Self::submit_sign_withdrawal(
```
bmwill: nit: this is the checkpoint that the txn is included in, not really a "signed_checkpoint"

crates/hashi/src/onchain/mod.rs
```diff
@@ -87,6 +87,8 @@ struct Inner {
     tls_private_key: Option<ed25519_dalek::SigningKey>,
     grpc_max_decoding_message_size: Option<usize>,
     metrics: Option<Arc<crate::metrics::Metrics>>,
+    /// Set once after guardian bootstrap; advanced by the watcher.
+    local_limiter: RwLock<Option<Arc<crate::guardian_limiter::LocalLimiter>>>,
```
bmwill: this is an interesting type signature...

crates/hashi/src/onchain/watcher.rs
```diff
@@ -437,14 +438,41 @@ async fn handle_events(client: &mut Client, state: &OnchainState, events: &[Hash
...
             HashiEvent::WithdrawalSignedEvent(event) => {
                 tracing::info!(withdrawal_txn_id = %event.withdrawal_txn_id, "Withdrawal signatures stored on-chain");
-                let mut state = state.state_mut();
-                if let Some(txn) = state
-                    .hashi
-                    .withdrawal_queue
-                    .withdrawal_txns
-                    .get_mut(&event.withdrawal_txn_id)
+                // Watcher is the sole mutator of the local limiter
+                // post-bootstrap; advance it inline with the mirror update.
```
bmwill, line 442: ok sorry for the slight whiplash but i take it we also are expected to have a a reconciliation against the guardian's state itself right? I ask because atm we can technically miss events due to checkpoint subscription breaking so when that happens we'll want to restart the subscription and then fetch the real value from the guardian to make sure we're in-sync. 

Eventually, end of next month most likely, i'll rewrite all of this with the new txn filtering logic we'll have to make this a bit nicer though.

crates/hashi/src/onchain/mod.rs
bmwill: Theres already a rwlock in the limiter itself, can we squash these a bit so this ends up just being Option<Arc<Limiter>>

crates/hashi-types/src/guardian/bitcoin_utils.rs
0xsiddharthks: Yeah, thats right, we're doing the check one level up at the validator commit step (in `validate_withdrawal_tx_commitment`), so it never reaches the guardian as an explicit check (the BLS cert subsumes it).

The guardian then `verify_hashi_cert`s that commitment, so it transitively trusts the equality.

crates/hashi/src/onchain/mod.rs
0xsiddharthks: Yeah, this was getting too complicated for me 😅 , so i was pretty unsure about this one as well. Makes sense , I think we can fix that. Let me update the PR for that.

crates/hashi/src/onchain/watcher.rs
bmwill: ok so only thing remaining is re-syncing the limiter state from the guardian correct?

0xsiddharthks: Thats right! I'm implementing that right now, as a separate PR on top of the stack. (since it is independent from this PR).

bmwill (approved):

PR #532, opened 2026-05-11
[mpc] handle missing previous committee at genesis

0xsiddharthks:
Fixes a #502 regression. On a fresh devnet, `create_mpc_manager` errors with `no committee for previous epoch 0` and the initial DKG never starts — it demands a previous-epoch committee that doesn't exist yet. `MpcManager::new` already handles the genesis case, the outer wrapper just never reached it.

Moved the selection into `CommitteeSet::previous_committee_for_target` and added unit tests for the genesis / rotation / recovery cases.

### DIFF
--- a/crates/hashi/src/lib.rs
+++ b/crates/hashi/src/lib.rs
@@ -360,8 +360,6 @@ impl Hashi {
             validator_address,
             epoch,
         )?;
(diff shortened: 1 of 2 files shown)


bmwill:
how do any unit tests work without this?

0xsiddharthks:
closed in favour of #533

PR #535, opened 2026-05-12
[leader] reload deposit worklist after the deposit time-delay

0xsiddharthks:
## Issue

We only reload the deposit worklist when a new Bitcoin block arrives. But a deposit's Confirm phase is gated on the Sui clock (`bitcoin_deposit_time_delay_ms`), and a successful Approve task drops the deposit from the worklist — so once approved, a deposit's confirmation waits for the *next Bitcoin block* to reload the worklist, even though the delay it's waiting on is a Sui-clock quantity. With sparse blocks, approved deposits sit idle between blocks.

## Changes

Re-arm a one-shot timer on each Bitcoin block that fires `bitcoin_deposit_time_delay_ms` later and reloads the worklist once. Once a deposit is back in the worklist the existing per-checkpoint re-evaluation handles the precise Confirm timing.

Reloads stay block-paced (at most ~2 per block) so transient-failure retries remain rate-limited per #480 — unlike a fixed-interval poll, which would retry failures unboundedly when the queue can't drain in time and revive the log spam #480 fixed.

### DIFF
--- a/crates/hashi/src/leader/mod.rs
+++ b/crates/hashi/src/leader/mod.rs
@@ -62,6 +62,7 @@ use x509_parser::nom::AsBytes;
 
 const NUM_CONSECUTIVE_LEADER_CHECKPOINTS: u64 = 100;
 const LEADER_TASK_TIMEOUT: Duration = Duration::from_secs(60);
+const DEPOSIT_WORKLIST_REFRESH_INTERVAL_MS: u64 = 30_000;
 
 #[derive(Clone, Copy, Debug)]
 enum DepositPhase {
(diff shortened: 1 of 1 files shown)


dlukegordon (changes requested):
This is not going to work, currently we don't do any backoff of failed deposit requests, so we could easily get into a situation where we overload the queue and can't process everything within 30 seconds, and will bring back the massive log spam.

How about instead we check one additional time after each bitcoin block, at the block time + bitcoin_deposit_time_delay_ms? Or we even don't check after each block, and instead we only check after each block time + bitcoin_deposit_time_delay_ms. So when the block hits, we just schedule a job which will execute in bitcoin_deposit_time_delay_ms.

crates/hashi/src/leader/mod.rs
```diff
@@ -214,6 +226,16 @@ impl LeaderService {
                     self.check_delete_expired_deposit_requests(checkpoint_timestamp_ms);
                     self.process_deposit_requests();
                 }
+                () = &mut deferred_deposit_reload, if deferred_deposit_reload_armed => {
```
dlukegordon, line 229: I just refactored the leader code to go into multiple smaller files, so please consider if some of this code should now go in leader/deposits.rs

0xsiddharthks: oh cool cool. let me look into this and rebase.

dlukegordon:
Duplicate of https://github.com/MystenLabs/hashi/pull/630

PR #540, opened 2026-05-13
docs: add GitHub Pages PR-preview workflow

jon-mysten:
Mirrors the pages-preview workflow Seal uses:
https://github.com/MystenLabs/seal/blob/main/.github/workflows/pages-preview.yaml

On every PR that touches `design/**`:
- `build-with-linkcheck` job runs `npm run build` (catches broken links, MDX syntax errors, broken anchors before merge).
- `preview` job builds the site with `DOCUSAURUS_BASE_URL=/hashi/pr-preview/pr-<N>/` and publishes it to the gh-pages branch under `pr-preview/pr-<N>/` via `rossjrw/pr-preview-action`. The action posts a sticky comment on the PR with the preview URL and removes the directory when the PR is closed.

Config changes to support the preview build:
- `baseUrl` reads `DOCUSAURUS_BASE_URL` env var, falling back to the production path (`/hashi/design/`).
- Hardcoded `/hashi/design/...` paths in the kapa modal image, kapa navbar trigger img src, and the llms-txt directive link now use the computed `${baseUrl}` so they resolve under any deploy target.
- Plausible tracking is suppressed on previews (empty `domain:` makes the client bail out), so preview pageviews don't pollute prod analytics.

Verified both builds emit correct paths:
- prod: `data-modal-image="/hashi/design/img/logo.svg"` etc.
- preview: `data-modal-image="/hashi/pr-preview/pr-99/img/logo.svg"` etc.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -34,11 +34,36 @@ jobs:
       - name: Build Docusaurus site
         run: cd design && npm run build
 
-      - name: stage build artifacts
+      - name: Stage build artifacts
         run: |
-          mkdir site
+          mkdir -p site
           mv ./target/doc site/rustdoc
           mv design/build site/design
+
+      # We deploy with `force_orphan: true` so the gh-pages branch stays a
+      # single orphan commit (no history accumulation), but that wipes the
+      # branch contents — which would blow away the `pr-preview/` directory
+      # the .github/workflows/pages-preview.yaml workflow writes there.
+      #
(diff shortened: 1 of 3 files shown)


bmwill:
Wouldn't the publish job overwrite the preview if a publish happens after the preview was uploaded?

zhu-jon:
> Wouldn't the publish job overwrite the preview if a publish happens after the preview was uploaded?

good catch, pushed a fix

bmwill (approved):

PR #542, opened 2026-05-13
Backups 2/3: Automatically backup after epoch change

dlukegordon:


crates/hashi/src/mpc/service.rs
```diff
@@ -723,9 +723,15 @@ impl MpcService {
...
-                 will retry at next trigger"
+                  will retry at next trigger"
             );
         }
+        let hashi = self.inner.clone();
+        let _backup_task = tokio::task::spawn_blocking(move || {
+            if let Err(e) = hashi.backup_after_epoch_change(target_epoch) {
+                error!("Automatic backup after epoch {target_epoch} failed: {e:#}");
+            }
+        });
```
bmwill: Do we want to save the task so that we can properly wait for it to finish in the `Service` so that graceful shutdown works properly?

dlukegordon: Good point, small chance this happens but might as well prevent the possibility.

dlukegordon: Pushed a commit adding a backup service

bmwill (approved):

PR #549, opened 2026-05-15
Allow guardian configuration at publish time

Bridgerz:
## Summary
- Adds optional `guardian_url` and `guardian_public_key` parameters to `finish_publish` so the guardian can be configured at deploy time instead of requiring a governance proposal
- Adds `--guardian-url` and `--guardian-public-key` CLI flags to the publish command
- Both must be provided together or both omitted; e2e tests pass `None` (no guardian in localnet)

## Test plan
- [ ] `sui move build` and `sui move test` pass
- [ ] `cargo build` passes
- [ ] Publish to devnet with guardian params and verify config is set
- [ ] Publish to devnet without guardian params and verify it still works


0xsiddharthks (approved):

0xsiddharthks:
🫶

PR #553, opened 2026-05-15
[cli][internal tools] add epoch override during key-recovery

0xsiddharthks:
as of now, the `key-recovery` auto-detects the source epoch from the DB's latest encryption key.

However, the DB only retains rotation messages for the most recent rotation, and that rotation typically completes *before* the on-chain committee for the new epoch is published. 

So the auto-detected epoch lands in a window where `MpcManager::new` can't find a committee for `reconstruction_epoch` and there's no on-chain epoch where rotation messages are still in the DB.

So this PR basically adds an `--epoch` flag to pin the source epoch.

### DIFF
--- a/crates/internal-tools/src/key_recovery.rs
+++ b/crates/internal-tools/src/key_recovery.rs
@@ -37,6 +37,12 @@ pub struct Args {
     /// Expected public key in hex (for verification).
     #[arg(long)]
     expected_pubkey: Option<String>,
+
+    /// Override the source epoch (the one whose KeyRotation output we want
+    /// to reconstruct). Defaults to the latest encryption-key epoch found
+    /// in the first DB.
+    #[arg(long)]
+    epoch: Option<u64>,
 }
 
 pub async fn run(args: Args, onchain_state: &OnchainState, chain_id: &str) -> anyhow::Result<()> {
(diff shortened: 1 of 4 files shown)


crates/internal-tools/src/key_recovery.rs
```diff
@@ -37,28 +37,109 @@ pub struct Args {
...
             })?
             .clone()
     };
 
+    // `MpcManager::new` requires a committee for `reconstruction_epoch`;
+    // graft `previous_epoch`'s in if the cluster hasn't published it yet.
+    {
+        let mut state = onchain_state.state_mut_for_recovery();
+        let committees = state.hashi_mut().committees.committees_mut();
+        if !committees.contains_key(&reconstruction_epoch)
```
dlukegordon: Should we check if it's missing, or mismatches as well?

crates/internal-tools/src/key_recovery.rs
```diff
@@ -37,28 +37,109 @@ pub struct Args {
...
             .ok_or_else(|| {
-                anyhow!("no committee found for epoch {reconstruction_epoch} or {previous_epoch}")
+                anyhow!("no committee found for epoch {previous_epoch} or {reconstruction_epoch}")
             })?
             .clone()
     };
 
+    // `MpcManager::new` requires a committee for `reconstruction_epoch`;
+    // graft `previous_epoch`'s in if the cluster hasn't published it yet.
+    {
```
dlukegordon: We're mutating the shared onchain state here. I wonder if it could be overwritten by the watcher if there's a rescape, leading to race conditions?

Perhaps we could remove `state_mut_for_recovery()` and `hashi_mut()`, and instead clone and patch and pass into mpc manager?

0xsiddharthks: yeah good catch! Just fixed this.

dlukegordon (approved):

crates/internal-tools/src/key_recovery.rs
0xsiddharthks: gotcha! yeah, just fix this.

the overlay is now unconditional, and both "missing" and "mismatched" cases collapse to the same code path. If an on-chain committees[reconstruction_epoch] already exists and differs from committees[previous_epoch], the tool prints a warning so we don't silently ignore a real new committee.

PR #554, opened 2026-05-15
[guardian][cli][internal-tools] Add cli for full OperatorInit + ProvisionerInit guardian bootstrap

0xsiddharthks:
Add script to fully initialize a freshly deployed guardian against the real onchain committee

Paired with MystenLabs/sui-operations#7955.

## Changes
- `internal-tools bootstrap-guardian`: 
    - scrapes the live `HashiCommittee` via `OnchainState`
    - generates BTC master + Shamir shares in memory
    - runs OperatorInit → GetGuardianInfo → ProvisionerInit × THRESHOLD over gRPC
        - Threshold derived from `certificate_threshold(committee.total_weight())`. 
- Containerfile: copy `/etc/ssl/cert.pem` so the AWS SDK's rustls-native-certs verifier can validate the S3 endpoint chain.


crates/internal-tools/Cargo.toml
```diff
@@ -27,3 +28,7 @@ reqwest.workspace = true
 futures.workspace = true
 csv = "1"
 url = "2"
+
+# bootstrap-guardian only
```
dlukegordon, line 32: Should these go in the root Cargo.toml?

crates/internal-tools/src/bootstrap_guardian.rs
```diff
@@ -0,0 +1,234 @@
...
+
+    let info_pb = client
+        .get_guardian_info(pb::GetGuardianInfoRequest {})
+        .await
+        .context("GetGuardianInfo RPC failed")?
+        .into_inner();
+    let info = GetGuardianInfoResponse::try_from(info_pb)
+        .map_err(|e| anyhow!("decode GetGuardianInfoResponse: {e:?}"))?;
+    let enc_pubkey = EncPubKey::from_bytes(&info.signed_info.data.encryption_pubkey)
+        .map_err(|e| anyhow!("decode guardian encryption pubkey: {e:?}"))?;
```
dlukegordon: The tool decodes GetGuardianInfoResponse and directly uses info.signed_info.data.encryption_pubkey. It never calls info.signed_info.verify(&info.signing_pub_key), never checks the returned guardian info matches the just-submitted share commitments/bucket, and does not pin/check a session id like hashi-monitor does in crates/hashi-monitor/src/kp/mod.rs:141-155.

0xsiddharthks: gotcha. yeah fixed this.

crates/internal-tools/Cargo.toml
0xsiddharthks: Yeah. these two dependencies were redeclared in a bunch of crates. moved to root Cargo.toml and added `workspace = true` everywhere else.

crates/internal-tools/src/bootstrap_guardian.rs
dlukegordon: Codex:
> The update now verifies signed_info with resp.signing_pub_key and checks echoed bucket/commitments. That fixes the specific bug where an endpoint could return an unsigned/tampered GuardianInfo from the real guardian response shape.
It does not authenticate that resp.signing_pub_key belongs to the expected enclave/session. A malicious endpoint can still accept OperatorInit, echo the submitted bucket/commitments, sign its own GuardianInfo, return its own encryption key, and receive THRESHOLD encrypted shares. Since verify_enclave_attestation is currently a TODO in hashi-types, there is no real remote attestation gate here.

0xsiddharthks: Yeah this is by design. 
I am planning on adding this once we move to a nitro enclave. currently the guardian is running in a kub cluster.

Adding a TODO for now.

dlukegordon (approved):

PR #555, opened 2026-05-16
[guardian] in-process StandardWithdrawal response cache

0xsiddharthks:
## Summary

- Throwaway branch — not for merge. Used to deploy a guardian that re-serves cached `StandardWithdrawal` responses for the K8s dev cluster.
- Cache keyed by `wid` (withdrawal ID): any retry for the same withdrawal is a cache hit, so the guardian consumes each withdrawal's limiter capacity exactly once regardless of seq churn (hashi restart, MPC retries). A new `wid` forwards upstream.

## Test

- `cargo nextest run -p hashi-guardian` (cache hit/miss covered by unit tests).
- `make fmt && make clippy` clean.

## Follow-ups

- The Nitro design replaces this with an out-of-enclave proxy with the same wire shape.

### DIFF
--- a/crates/hashi-guardian/src/cache.rs
+++ b/crates/hashi-guardian/src/cache.rs
@@ -1,0 +1,337 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Temporary in-process response cache for the guardian's `StandardWithdrawal`
+//! RPC. Lives on the throwaway `siddharth/guardian-response-cache` branch and is
+//! not intended for merge — the Nitro design pulls this layer out into an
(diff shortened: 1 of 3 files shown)


0xsiddharthks:
@mskd12 I think we can probably merge this in for the private-testnet run this week.
Next week, we'll be migrating to the nitro enclave with a managed caching solution built in the sidecar.

So Could we merge it right now (so that we can deploy PTN from main), and I will revert it once we have a separate caching layer.

The change is primarily isolated to the new file `crates/hashi-guardian/src/cache.rs` so it should be very straightforward to revert.

mskd12 (approved):

PR #556, opened 2026-05-16
[guardian] bootstrap with on-chain committee

0xsiddharthks:
## Summary
- Pull the live `HashiCommittee` from the on-chain `Hashi` object via a one-shot `OnchainState` scrape (watcher aborts on drop) instead of the single-member `mock_dev_committee`.
- Derive `committee_threshold` from `certificate_threshold(committee.total_weight())` so it matches what the leader uses to certify withdrawal requests.
- Add new env vars to `bootstrap_guardian_init`: `SUI_RPC_URL`, `HASHI_PACKAGE_ID`, `HASHI_OBJECT_ID`. Drop `HASHI_COMMITTEE_THRESHOLD` (derived now).
- `hashi` is an optional dep gated by `non-enclave-dev` — no impact on the enclave binary or the workspace's prod build matrix.

## Why
With the mock committee, every `StandardWithdrawal` RPC from real hashi nodes failed at `verify_hashi_cert` (`crates/hashi-guardian/src/withdraw.rs:79`) before reaching the rate-limit check — so the guardian's rate-limiter never actually exercised during dev-cluster testing.

## Caveat
Re-run this bootstrap whenever the on-chain committee rotates; the guardian holds an in-memory snapshot keyed to the epoch we submit.

## Test plan
- [ ] `cargo build -p hashi-guardian --features non-enclave-dev --example bootstrap_guardian_init` (done locally)
- [ ] Drive a fresh hashi-guardian-dev pod end-to-end via `scripts/hashi/bootstrap-guardian.sh` (stacked sui-operations PR provides the env-var plumbing) and confirm `GetGuardianInfo` reflects the on-chain committee + 10-BTC bucket.
- [ ] Run a deposit + withdrawal sweep on hashi-devnet that crosses the rate-limit; confirm 100 succeed and the next attempts are rejected by the guardian's `consume_from_limiter`.


0xsiddharthks:
Folded into #554 — that PR now contains the on-chain committee bootstrap, the new `hashi-guardian-bootstrap` crate, and the Containerfile + CryptoProvider fixes. Rebased onto main.

PR #558, opened 2026-05-17
[mpc] recover current epoch from stored rotation messages

0xsiddharthks:
## Issue

After `end_reconfig` for epoch N, every node's `current_output` is in-memory only and messages from the rotation that produced epoch N-1's shares have been pruned. A coordinated restart at that point leaves `run_key_rotation` with neither recovery source — the cluster stops signing and stays stuck until an externally driven reconfig to N+1 gives it new rotation messages to reconstruct from. This PR closes that gap.

## Changes

- What's still on disk after `end_reconfig` for N: the rotation messages from the just-completed `N-1 → N` run, stored under epoch N. Decrypting them with this node's epoch-N encryption key reproduces this node's shares — same result as `run_key_rotation_as_party`.
- Add `reconstruct_current_from_stored_rotation`, sharing the decrypt/combine body with `reconstruct_from_rotation_certificates` via a new `RotationReconstructionView` helper.
- `recover_mpc_state` calls it as a fast path before falling back to `run_key_rotation`. After reconstruction we cross-check the recovered public key against the on-chain `mpc_public_key`; on mismatch we fall back to the live protocol.
- Tests: unit test for the new reconstruction path, plus an e2e test that restarts every node after a rotation and asserts the cluster recovers without advancing the epoch.

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -2921,23 +2921,9 @@ impl MpcManager {
         })
     }
 
-    /// Reconstruct **this node's** `MpcOutput` for `self.mpc_config.epoch` from
-    /// the rotation messages on disk under that same epoch.
-    ///
-    /// `reconstruct_from_rotation_certificates` reconstructs the *previous*
-    /// epoch's output (the input to a not-yet-completed rotation). That path
-    /// can't be taken after a coordinated restart: `prune_messages_below(N)`
-    /// runs at the end of epoch `N`'s reconfig and wipes the previous epoch's
-    /// rotation messages, while `previous_output`/`current_output` are
-    /// in-memory only. There is no source of truth left to redo the
-    /// `N-1 -> N` rotation from.
-    ///
-    /// What *is* still on disk is each peer's rotation messages from the
-    /// just-completed `N-1 -> N` run, stored under epoch `N`. This method
-    /// uses those, decrypted with the current epoch's encryption key, to
-    /// re-derive this node's key shares for epoch `N` directly — the same
-    /// computation that `run_key_rotation_as_party` performs at the end of
-    /// the live protocol.
+    /// Re-derive this node's `MpcOutput` for `self.mpc_config.epoch` from the
+    /// rotation messages on disk under that same epoch — the data is exactly
+    /// what `run_key_rotation_as_party` would have produced live.
     pub fn reconstruct_current_from_stored_rotation(
         &mut self,
         certificates: &[CertificateV1],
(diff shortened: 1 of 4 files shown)


crates/hashi/src/mpc/service.rs
```diff
@@ -293,6 +300,59 @@ impl MpcService {
...
+        let mpc_manager = self
+            .inner
+            .mpc_manager()
+            .ok_or_else(|| anyhow::anyhow!("MpcManager not initialized for fast-path recovery"))?;
+        let result = {
+            let mut mgr = mpc_manager.write().unwrap();
+            mgr.reconstruct_current_from_stored_rotation(&certificates)
+        };
+        match result {
+            Ok(output) => Ok(Some(output)),
```
dlukegordon: Should we check that the recovered key matches what we expect?

```
Ok(output) => {
    let recovered_key = bcs::to_bytes(&output.public_key)
        .expect("public key serialization should succeed");
    if recovered_key != onchain_state.mpc_public_key() {
        warn!(
            "try_recover_from_stored_rotation: reconstructed key does not match on-chain key; falling back to live protocol"
        );
        return Ok(None);
    }
    Ok(Some(output))
}
```

crates/hashi/src/mpc/service.rs
```diff
@@ -293,6 +300,59 @@ impl MpcService {
         Ok(output)
     }
 
+    /// Returns `Ok(Some)` only when the current epoch's rotation messages are
+    /// on disk and reconstruct cleanly; otherwise `Ok(None)` so the caller
+    /// runs the live protocol.
+    async fn try_recover_from_stored_rotation(
```
dlukegordon, line 306: Should we add a test?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2921,6 +2921,134 @@ impl MpcManager {
         })
     }
 
+    /// Re-derive this node's `MpcOutput` for `self.mpc_config.epoch` from the
+    /// rotation messages on disk under that same epoch — the data is exactly
+    /// what `run_key_rotation_as_party` would have produced live.
+    pub fn reconstruct_current_from_stored_rotation(
```
zhouwfang, line 2949: We should reduce the significant duplication with `reconstruct_from_rotation_certificates`.

zhouwfang (commented):
Good catch. But without this PR, the nodes will be automatically recovered when the next epoch starts. So "Coordinated restarts in an already-completed key-rotation epoch couldn't recover" in the PR description is not accurate.

Please add an e2e test.

crates/hashi/src/mpc/mpc_except_signing.rs
0xsiddharthks: gotcha. Yeah, moved the shared part into `reconstruct_rotation_in_view`.

0xsiddharthks:
@zhouwfang , @dlukegordon  thanks for the review. updated the description and added the e2e test: `test_coordinated_restart_after_rotation_recovers_without_new_rotation`

crates/hashi/src/mpc/service.rs
0xsiddharthks: Thanks! yeah, added this snippet. on mismatch, we fallthrough to `run_key_rotation`.

crates/hashi/src/mpc/service.rs
0xsiddharthks: Yeah! 
Just added some:

unit test: `test_reconstruct_current_from_stored_rotation`
e2e test: `test_coordinated_restart_after_rotation_recovers_without_new_rotation`

dlukegordon (approved):

PR #568, opened 2026-05-19
[leader] drive guardian committee handoff

0xsiddharthks:
builds on top of #569. The leader drives the guardian's committee forward to match the on-chain epoch each tick.

## Changes

- `SignCommitteeTransition` RPC:
  - each peer rebuilds the transition payload from on-chain state at `from_epoch + 1` and signs with the historical `from_epoch` BLS key.
  - uses `Hashi::sign_message_proto_at_epoch` for signing with a historical epoch's key rather than the current one.
- leader loop:
  - Each leader tick spawns a bounded reconcile task to check / update the guardian epoch.
- add a metric `hashi_guardian_current_committee_epoch` for the guardian's reported epoch.


crates/hashi-types/proto/sui/hashi/v1alpha/bridge_service.proto
```diff
@@ -186,3 +188,13 @@ message SignGuardianWithdrawalRequestRequest {
 message SignGuardianWithdrawalRequestResponse {
   MemberSignature member_signature = 1;
 }
+
+// Peers rebuild the transition from on-chain state; no committee bytes on the wire.
+message SignCommitteeTransitionRequest {
+  // Outgoing committee epoch (new_epoch = from_epoch + 1).
```
bmwill: note that we can skip epochs so the invariant you're expecting new_epoch = from_epoch +1 does not hold.

crates/hashi/src/leader/mod.rs
```diff
@@ -2305,6 +2331,205 @@ impl LeaderService {
...
+            if guardian_epoch >= hashi_epoch {
+                return Ok(());
+            }
+
+            let from_epoch = guardian_epoch;
+            info!(
+                from_epoch,
+                hashi_epoch, "Driving guardian committee handoff"
+            );
+            let signed = Self::collect_committee_transition_signatures(inner, from_epoch).await?;
```
bmwill, line 2405: So one challenge with this is that we're expecting a previous committee to ratchet the guardian up to a new epoch. Assuming this happens quickly after epoch change this should be fine, and we can always recover by having the provisioners explicitly do the ratcheting themselves. But you need to worry about the "leader" maybe not even being a part of that previous committee.

crates/hashi/src/leader/mod.rs
```diff
@@ -169,6 +174,7 @@ impl LeaderService {
                         continue;
                     }
 
+                    self.maybe_reconcile_guardian_committee();
```
dlukegordon: Can we rename to `check_reconcile_guardian_committee` to match the other functions?

crates/hashi/src/leader/mod.rs
```diff
@@ -2305,6 +2331,205 @@ impl LeaderService {
...
+        let outcome = if result.is_ok() {
+            crate::metrics::GUARDIAN_RPC_OUTCOME_OK
+        } else {
+            crate::metrics::GUARDIAN_RPC_OUTCOME_UNAVAILABLE
+        };
+        inner
+            .metrics
+            .record_guardian_rpc(method, outcome, start.elapsed().as_secs_f64());
+        result
+    }
```
bmwill: We don't want to add explicit wrappers around rpc apis to do metrics collection. We already setup metrics collection for validator-validator communication using middleware we should do the same for the validator-guardian interactions too. that will properly handle timing and matrics related to all routes and status'

crates/hashi/src/leader/mod.rs
```diff
@@ -2305,6 +2331,205 @@ impl LeaderService {
...
+
+    async fn collect_committee_transition_signatures(
+        inner: &Arc<Hashi>,
+        from_epoch: u64,
+    ) -> anyhow::Result<SignedMessage<CommitteeTransition>> {
+        // TODO: hashi committee epochs are sparse — find the next entry in
+        // `committees` instead of assuming `+1`.
+        let to_epoch = from_epoch + 1;
```
bmwill: oh you already have a todo here, lets just fix this instead of having a TODO because this could brick us since we have seen this cause issues already in the past.

bmwill (dismissed):

crates/hashi/src/leader/mod.rs
dlukegordon: Also, we check this every checkpoint (every few seconds), but we should only need to call this on an epoch change right? Not a huge deal since we just call guardian info and exit if the epoch hasn't changed, but maybe we can introduce a delay, or only trigger the job on an epoch change.

crates/hashi/src/leader/mod.rs
```diff
@@ -2305,6 +2331,205 @@ impl LeaderService {
...
+                // ProvisionerInit hasn't run yet; the bootstrap CLI seeds it.
+                return Ok(());
+            };
+            inner
+                .metrics
+                .guardian_current_committee_epoch
+                .set(guardian_epoch as i64);
+
+            let hashi_epoch = inner.onchain_state().epoch();
+            if guardian_epoch >= hashi_epoch {
```
dlukegordon: Would the guardian epoch every be greater than the hashi epoch? Perhaps in this case we should log a warning?

crates/hashi/src/leader/mod.rs
```diff
@@ -2305,6 +2331,205 @@ impl LeaderService {
             })
             .ok()
     }
+
+    // ========================================================================
+    // Guardian: committee handoff (post-rotation)
+    // ========================================================================
+
+    /// Walk the guardian forward one epoch at a time until it matches the chain.
```
dlukegordon, line 2357: Why do we have to walk the guardian forward one epoch at a time, instead of just advancing directly to the current epoch?

bmwill: chain of trust issue, you need the past committee to sign off on the next committee. Ideally this would actually be solved in one of two other ways:
-  have past committees provide a hand-off signature as a part of closing off hashi epoch change. That signature is stored onchain so that the process of catching up the guardian is easy as providing that chain and anyone can do it
- The guardian does its own trust ratcheting and follows the onchain sui state itself, possibly using checkpoint object inclusion proofs to attest to epoch change txns

The existing mechanism requires past committees to be online which can be a problem. I think option 1 is probably the easiest change to make

crates/hashi/src/leader/mod.rs
```diff
@@ -238,11 +244,31 @@ impl LeaderService {
...
+        if self
+            .guardian_committee_reconcile_task
+            .as_ref()
+            .is_some_and(|t| !t.is_finished())
+        {
+            return;
+        }
+        let inner = self.inner.clone();
+        let handle = tokio::spawn(async move { Self::reconcile_guardian_committee(&inner).await });
+        self.guardian_committee_reconcile_task = Some(AbortOnDropHandle::new(handle));
```
dlukegordon, line 287: If the old reconcile task has finished but the select! completion branch has not consumed it yet, this overwrites the handle, and the result/error is lost. Perhaps we should also handle the completed task here, or only handle it here.

crates/hashi/src/leader/mod.rs
```diff
@@ -2305,6 +2331,205 @@ impl LeaderService {
...
+                inner,
+                crate::metrics::GUARDIAN_RPC_METHOD_GET_GUARDIAN_INFO,
+                guardian.get_guardian_info(),
+            )
+            .await
+            .map_err(|e| anyhow::anyhow!("GetGuardianInfo failed: {e}"))?;
+            let info: hashi_types::guardian::GetGuardianInfoResponse = info_pb
+                .try_into()
+                .map_err(|e| anyhow::anyhow!("parse GetGuardianInfoResponse: {e:?}"))?;
+            let Some(guardian_epoch) = info.current_committee_epoch else {
```
dlukegordon: I think we could reuse the current_committee_epoch from UpdateCommittee here again if this is not the first iteration of the loop.

crates/hashi/src/leader/mod.rs
```diff
@@ -2305,6 +2331,205 @@ impl LeaderService {
...
+            let proto = signed_committee_transition_to_pb(&signed);
+
+            let resp = Self::time_guardian_rpc(
+                inner,
+                crate::metrics::GUARDIAN_RPC_METHOD_UPDATE_COMMITTEE,
+                guardian.update_committee(proto),
+            )
+            .await
+            .map_err(|status| anyhow::anyhow!("UpdateCommittee failed: {}", status.message()))?;
+            let new_guardian_epoch = resp.current_committee_epoch.unwrap_or(from_epoch);
```
dlukegordon: Should we log a warning or bail if resp doesn't contain current_committee_epoch? Seems like that would be a rare and erroneous condition.

crates/hashi/src/leader/mod.rs
dlukegordon: https://linear.app/mysten-labs/issue/IOP-421/advance-guardian-multiple-epochs-at-a-time

crates/hashi/src/withdrawals.rs
```diff
@@ -583,6 +591,47 @@ impl Hashi {
...
 
+    // --- Guardian: validate and BLS-sign a `CommitteeTransition` ---
+
+    /// Rebuild the transition from on-chain state and sign with the historical key.
+    #[tracing::instrument(level = "info", skip_all, fields(from_epoch))]
+    pub fn validate_and_sign_committee_transition(
+        &self,
+        from_epoch: u64,
+    ) -> anyhow::Result<hashi_types::proto::MemberSignature> {
+        // TODO: hashi committee epochs are sparse — find the next entry in
```
dlukegordon: Connected to todo above, we should also probably fix in this PR

crates/hashi/src/leader/mod.rs
0xsiddharthks: done both. 
Added `last_guardian_reconcile_epoch: Option<u64>` to `LeaderService` and gated `check_reconcile_guardian_committee` on `hashi_epoch != last_guardian_reconcile_epoch`. The first leader tick still triggers a reconcile (`None != Some(epoch)`), and after that we only spawn when the hashi epoch advances.

crates/hashi/src/leader/mod.rs
0xsiddharthks: Done!

Replaced `to_epoch = from_epoch + 1` with `committees_map.range((from_epoch + 1)..).next()` in both `collect_committee_transition_signatures` (leader side) and `validate_and_sign_committee_transition` (follower side).

Both sides read on-chain state independently and derive the same `to_epoch` deterministically.

crates/hashi/src/leader/mod.rs
0xsiddharthks: I don't think this can happen tbh, since the guardian only advances via certs signed by the hashi committees.

crates/hashi/src/leader/mod.rs
0xsiddharthks: oh good catch! I had completely missed that.

switched the in-flight check from `is_some_and(|t| !t.is_finished())` to plain `is_some()`.

The select! arm is the sole owner of the completion path now, so a finished-but-unconsumed result can never be overwritten and we just wait one loop iteration for it to drain.

crates/hashi/src/leader/mod.rs
0xsiddharthks: gotcha, added an anyhow clause so a missing field bails the reconcile with the error.

crates/hashi/src/withdrawals.rs
0xsiddharthks: Yes! Fixed this, on both the leader and follower side.

crates/hashi/src/leader/mod.rs
0xsiddharthks: @bmwill wait i think the leader doesn't have to be part of the outgoing commmittee, since it only fans out the `SignCommitteeTransition` over peer RPCs to the outgoing committee's members and aggregates a threshold cert.

Each follower signs with the historical epoch key it kept in `db.signing_keys`, so a hashi-server that has rotated out of the active committee can still sign for an epoch it was in.

so as long as enough former committee members are online i think we should be fine, right?

crates/hashi/src/leader/mod.rs
0xsiddharthks: Done!
Pulled the initial `GetGuardianInfo` out of the loop and hoisted `guardian_epoch` into a mutable seed

crates/hashi/src/leader/mod.rs
dlukegordon: Awesome

dlukegordon (approved):

PR #569, opened 2026-05-19
[guardian] add UpdateCommittee RPC for committee handoff

0xsiddharthks:
## Context

The guardian's committee is only set once during `ProvisionerInit`.

So this, adds a new RPC to the guardian for updating the committee at every epoch change.

## Changes

- `UpdateCommittee(SignedCommitteeTransition)` RPC: 
  - verifies the outgoing committee's threshold signature
  - swaps the in-memory committee
  - logs to S3
  - Idempotent on already-applied / older transitions

## Deployment

This has already been deployed to devnet since last week.


crates/hashi-guardian/src/committee_update.rs
```diff
@@ -0,0 +1,248 @@
...
+    let msg = CommitteeUpdateLogMessage::Success {
+        from_epoch,
+        to_epoch: new_committee.epoch,
+        new_committee: new_committee.clone(),
+        request_sign: signed.committee_signature().clone(),
+    };
+    enclave.log_committee_update(msg).await
+}
+
+async fn log_failure(
```
dlukegordon, line 99: Should we make this work like failure logging in withdrawals, and return an error?

```
async fn log_failure(...) -> GuardianResult<()> {
    if let Err(log_err) = enclave.log_committee_update(msg).await {
        error!(...);
        return Err(GuardianError::InternalError(format!(
            "Failed to log committee update error {err} due to S3 logging error {log_err}"
        )));
    }
    Ok(())
}
```

crates/hashi-guardian/src/committee_update.rs
```diff
@@ -0,0 +1,248 @@
...
+use hashi_types::guardian::GuardianResult;
+use hashi_types::guardian::HashiCommittee;
+use hashi_types::guardian::HashiSigned;
+use std::sync::Arc;
+use tracing::error;
+use tracing::info;
+
+/// Advance the committee one epoch with a cert from the outgoing committee.
+/// Idempotent on already-applied or older transitions.
+pub async fn update_committee(
```
dlukegordon, line 25: From 5.5:
> Concurrent UpdateCommittee calls can roll the committee epoch backwards. update_committee reads current once, logs success, then blindly calls replace_committee with no expected-epoch check or lock spanning the read/log/replace sequence. A real interleaving is: request A validates 5 -> 6 and stalls after logging; request B applies 5 -> 6; request C applies 6 -> 7; request A resumes and replaces the committee with epoch 6 again. That leaves the guardian behind after it already accepted epoch 7. The race is at crates/hashi-guardian/src/committee_update.rs:29 and crates/hashi-guardian/src/committee_update.rs:69-71; replace_committee at crates/hashi-guardian/src/enclave.rs:296-308 does not verify the old epoch.

We could fix this my making adding a lock for the committee updates. Maybe we could also add an expected_epoch field to replace_committee and check it.

crates/hashi-types/src/guardian/mod.rs
```diff
@@ -63,13 +63,15 @@ pub const S3_OBJECT_LOCK_DURATION_INIT: Duration = ONE_WEEK;
 pub const S3_OBJECT_LOCK_DURATION_WITHDRAW: Duration = ONE_WEEK;
 pub const S3_OBJECT_LOCK_DURATION_HEARTBEAT: Duration = ONE_WEEK;
 pub const S3_OBJECT_LOCK_DURATION_SECRET_SHARING: Duration = ONE_WEEK;
+pub const S3_OBJECT_LOCK_DURATION_COMMITTEE_UPDATE: Duration = ONE_WEEK;
 
 /// S3 sub-prefixes used for guardian log streams.
 /// See `crates/hashi-guardian/README.md` for canonical key layout.
```
mskd12, line 69: Nite: Update `crates/hashi-guardian/README.md` re the layout

crates/hashi-types/src/guardian/mod.rs
```diff
@@ -304,6 +317,24 @@ pub enum WithdrawalLogMessage {
     },
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub enum CommitteeUpdateLogMessage {
+    /// Committee handoff success
+    Success {
+        from_epoch: u64,
+        to_epoch: u64,
```
mskd12: Isn't `to_epoch` always `from_epoch + 1`? If so, shall we just have one field? I was wondering if you were trying to do batch updates given the message shape. Same comment also applies to Failure side.

crates/hashi-types/src/guardian/mod.rs
```diff
@@ -625,6 +656,28 @@ impl WithdrawalLogMessage {
     }
 }
 
+impl CommitteeUpdateLogMessage {
```
mskd12, line 662: I would lean towards a simpler format to start with. How about a single S3 directory (no hour scoping) where key formats are as follows:
- Success: `{new_epoch}-{session_id}.json`
- Failures: `failure-{new_epoch}-{session_id}-{random_suffix}.json`

crates/hashi-guardian/src/committee_update.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    if proposed_epoch != current_epoch + 1 {
+        let err = InvalidInputs(format!(
+            "non-sequential committee transition: current {current_epoch} -> proposed {proposed_epoch}"
+        ));
+        log_failure(&enclave, current_epoch, &signed, &err).await;
+        return Err(err);
+    }
+
+    // `WithdrawalConfig.committee_threshold` is genesis-only; derive fresh.
```
mskd12: The purpose of `WithdrawalConfig.committee_threshold` is to allow us to assign a higher threshold in the guardian than in hashi, e.g., a normal hashi cert requires 66% but guardian expects 80%. So I think we should use `WithdrawalConfig.committee_threshold` here.

crates/hashi-guardian/src/committee_update.rs
```diff
@@ -0,0 +1,248 @@
...
+            "new committee epoch ({}) does not match transition epoch ({proposed_epoch})",
+            new_committee.epoch()
+        ));
+        log_failure(&enclave, current_epoch, &signed, &err).await;
+        return Err(err);
+    }
+
+    // Log before the in-memory swap so failed S3 writes don't advance the committee.
+    log_success(&enclave, current_epoch, &signed).await?;
+    enclave.state.replace_committee(new_committee)?;
```
mskd12: This can be converted to an expect because `is_fully_initialized` already checks that committee is initialized.

crates/hashi-guardian/src/committee_update.rs
0xsiddharthks: yeah this makes sense.

I added a mutex on the `committee_update_lock`. and added the `expected_current_epoch` param.

crates/hashi-guardian/src/committee_update.rs
0xsiddharthks: yeah! done. we are now following a similar pattern to the `log_withdrawal_failure` in `withdraw.rs`.

crates/hashi-guardian/src/committee_update.rs
0xsiddharthks: ~oO i see. i didn't know this. fixed it :)~
reverted the fix. @bmwill's comment makes sense.

crates/hashi-types/src/guardian/mod.rs
bmwill: no it is not always +1, we have lots of cases where it isn't. unless you're talking about sui's epoch and not hashi's.

crates/hashi-guardian/src/committee_update.rs
bmwill: This likely won't really work and will need to be kept in lockstep with what hashi's committee expects because we bail early in sig collection once we have a sufficient number of sigs in some cases.

dlukegordon (approved):

crates/hashi-guardian/src/committee_update.rs
mskd12: Hmm, got it. cc'ing @benr-ml who may have thoughts on this. I think we could table that feature for later - in that case, I can get rid of `WithdrawalConfig.committee_threshold` altogether (in a later PR).

mskd12 (approved):

crates/hashi-guardian/src/committee_update.rs
bmwill: FWIW we are requiring 66% for signing in hashi which is more than the 33% needed to actually sign with mpc

benr-ml: Yeah, we did talk about having a higher threshold for this cert. Let's have at least a ticket to consider that again later since that extra security comes almost "for free" (in the protocol level).

PR #572, opened 2026-05-19
[Guardian] Provisioner init: Recover limiter state from withdraw logs

mskd12:
## Summary

Fixes [IOP-252](https://linear.app/mysten-labs/issue/IOP-252/compute-withdrawalstate-from-s3-logs). Two stacked commits on this branch (in dependency order):

1. **`[hashi-monitor] rename kp -> provisioner-init; refresh README`** — rename the `kp`/`kp-init` module and CLI subcommand to `provisioner`/`provisioner-init` (matches the existing `ProvisionerInitRequest` / gRPC verb). Also refreshes the README which was stale.

2. **`[guardian][hashi-monitor] embed limiter post_state in withdraw logs; recover at provisioner-init`** — replaces the mock `LimiterState` in provisioner-init with one recovered from S3.

### Schema changes (hashi-types / hashi-guardian)

- Add `post_state: LimiterState` to `WithdrawalLogMessage::Success`. Populated in `log_withdrawal_success` from the held limiter guard.
- New S3 key format for withdraw logs:
  - `withdraw/{yyyy}/{mm}/{dd}/{hh}/success-{seq:020}-{session_id}-wid{wid}.json`
  - `withdraw/{yyyy}/{mm}/{dd}/{hh}/failure-{session_id}-wid{wid}-{rand8}.json`

  Seq-prefix on Success makes lex order = seq order within an hour bucket → max-seq is the lex-last key, no body parsing required.
- New `LimiterState::genesis(&WithdrawalConfig)` helper; used at provisioner-init, `bootstrap_guardian`, and test sites.

### Recovery flow (hashi-monitor)

- `detect_deployment_mode` inspects `init/` for any session other than the heartbeat-audit-selected live session. ≥1 prior session ⇒ `Rotation`; only the current session present ⇒ `Genesis`. No operator-supplied "is this genesis?" flag needed.
- On `Rotation`: walk hour buckets back from `now`, take max-seq Success across the first non-empty bucket and one peek-back (clock-skew safety across hour boundaries), return that log's `post_state`. Cap to current config's `max_bucket_capacity_sats`. Error if no Success found within 7 days.
- On `Genesis`: `LimiterState::genesis(&cfg.withdrawal_config)`.
- Deliberately skips `is_readable`'s 10-min write-completion gate inside the recovery loop — `heartbeat_audit` + S3 read-after-write consistency already guarantee visibility; gating would cause an enclave that died late in an hour to have its final-hour bucket skipped.

### Notes for review

- Recovery's per-bucket cost is currently O(n) S3 object reads (fetches+verifies every body in the bucket). A NOTE in `limiter_recovery.rs` flags the natural O(1) optimization using the new seq-prefixed keys — list keys, pick lex-last `success-*`, fetch only that one. Deferred.
- `bootstrap_guardian` is still the path for fresh deployments end-to-end (it also runs `operator_init` + `setup_new_key`); this PR doesn't change that.

## Test plan

- [x] `cargo build -p hashi-monitor -p hashi-guardian -p hashi-types -p internal-tools` clean.
- [x] `cargo test -p hashi-types -p hashi-guardian -p hashi-monitor` — all unit tests pass (the one failure under `cargo test` is `rpc::btc::tests::lookup_btc_confirmation_with_local_regtest`, an environmental test that spawns `bitcoind`; unrelated).
- [ ] Run provisioner-init against a staging S3 bucket with a real prior session's logs; verify recovered state matches the prior enclave's last Success.
- [ ] Run provisioner-init against a fresh bucket (no init logs); verify it takes the Genesis branch.
- [ ] Verify rotation with a prior session that has zero Success logs errors out cleanly (no silent genesis downgrade).

### DIFF
--- a/crates/hashi-monitor/README.md
+++ b/crates/hashi-monitor/README.md
@@ -32,22 +32,22 @@ Audits the cross-system bridge flow on two parallel tracks.
 
 ### Batch audit
 ```bash
-cargo run -p hashi-monitor -- batch --config config.sample.yaml --start 1700000000 --end 1700003600
+cargo run -p hashi-monitor -- batch --config audit.sample.yaml --start 1700000000 --end 1700003600
 ```
 `--end` defaults to the current time if omitted.
 
 ### Continuous monitoring
 ```bash
-cargo run -p hashi-monitor -- continuous --config config.sample.yaml --start 1700000000
+cargo run -p hashi-monitor -- continuous --config audit.sample.yaml --start 1700000000
 ```
 
 ### Provisioner-init
 ```bash
-cargo run -p hashi-monitor -- provisioner-init --config provisioner-init.yaml
+cargo run -p hashi-monitor -- provisioner-init --config provisioner-init.sample.yaml
 ```
 
 ## Config
-See `config.sample.yaml` for a complete batch/continuous example:
+See `audit.sample.yaml` for a complete batch/continuous example:
 
 ```yaml
 # Liveness delay bounds (seconds)
@@ -70,7 +70,7 @@ btc:
     type: none
 ```
 
-The `provisioner-init` subcommand takes a separate YAML (`ProvisionerConfig`) with the key-provisioner share, expected share commitments, S3 config, Hashi committee, withdrawal config, and Hashi BTC master pubkey; if `guardian_endpoint` is set, the request is submitted after local checks pass.
+The `provisioner-init` subcommand takes a separate YAML (`ProvisionerConfig`, see `provisioner-init.sample.yaml`) with the key-provisioner share, expected share commitments, S3 config, Hashi committee, withdrawal config, and Hashi BTC master pubkey; if `guardian_endpoint` is set, the request is submitted after local checks pass.
 
 ## Status
 - Implemented:
--- a/crates/hashi-monitor/README.md
+++ b/crates/hashi-monitor/README.md
@@ -20,7 +20,7 @@ Audits the cross-system bridge flow on two parallel tracks.
 ### Modes
 1. **Batch**: one-time audit over a guardian time range `[start, end]`.
 2. **Continuous**: long-running monitor that polls Sui, Guardian S3, and BTC RPC on fixed intervals and reports findings as they appear.
-3. **Provisioner-init**: one-shot provisioner-init flow run by the key provisioner — audits heartbeats for the latest enclave session, verifies attestation and expected config, builds a `ProvisionerInitRequest`, and optionally submits it to a guardian endpoint.
+3. **Provisioner-init**: one-shot provisioner-init flow run by the key provisioner — audits heartbeats for the latest enclave session, verifies attestation and expected config, auto-detects whether this is a fresh deployment or a rotation (based on whether any prior session's init logs exist in S3) and sources the initial `LimiterState` accordingly (genesis defaults vs. recovered from the prior enclave's withdrawal logs), then builds a `ProvisionerInitRequest` and optionally submits it to a guardian endpoint.
 
 ### Timeline semantics (withdrawals)
 - User-provided `start` / `end` are interpreted on the **guardian (E2)** timeline.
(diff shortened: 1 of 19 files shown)


bmwill (approved):

crates/hashi-monitor/README.md
```diff
@@ -56,14 +70,15 @@ btc:
...
-  - `batch` and `continuous` are wired through the CLI but currently stop at deferred `todo!` implementation points.
+- Implemented:
+  - Domain model and withdrawal / deposit state-machine checks.
+  - Batch and continuous auditor loops (cursor advancement, BTC fetch, violation detection, GC, progress watermarks).
+  - Guardian S3 withdrawal log polling with attestation and signature verification.
+  - BTC confirmation lookup via Bitcoin Core RPC.
+  - Provisioner-init flow (heartbeat audit, attestation/config check, optional submission).
+- Not yet implemented:
+  - Sui event polling — `AuditorCore::poll_sui` is a stub that returns `CursorUnmoved`, so E1 (withdrawal) and the deposit pipeline currently see no Sui input.
+  - Limiter state recovery from S3 logs in `provisioner-init` (currently uses a mock).
```
dlukegordon: This is implemented now

crates/hashi-monitor/src/provisioner/mod.rs
```diff
@@ -38,17 +41,30 @@ pub async fn run(cfg: ProvisionerConfig) -> anyhow::Result<()> {
...
+    let limiter_state = match mode {
+        DeploymentMode::Rotation => {
+            let mut recovered = limiter_recovery::recover_limiter_state(s3_client.clone()).await?;
+            // Cap to the current config's bucket capacity in case max capacity
+            // was lowered across the rotation. (Raising is fine — refill will
+            // fill it.)
+            recovered.num_tokens_available = recovered
+                .num_tokens_available
+                .min(cfg.withdrawal_config.max_bucket_capacity_sats);
+            assert!(recovered.last_updated_at < now_unix_seconds()); // sanity check
```
dlukegordon: We allow withdrawal timestamps up to 5 mins in the future (withdraw.rs:93-100), so I think we should either allow that here in the assert, or change that in withdraw.rs.

crates/hashi-monitor/src/provisioner/limiter_recovery.rs
```diff
@@ -0,0 +1,92 @@
...
+        poller.retreat_cursor();
+    }
+
+    let state = best.ok_or_else(|| {
+        anyhow::anyhow!(
+            "no successful withdrawal logs found within last {} hours; \
+             cannot recover limiter state",
+            MAX_WALK_BACK_HOURS
+        )
+    })?;
```
dlukegordon: This makes rotation impossible if the previous enclave had no successful withdrawals in the last week. Do we want to fail in this case? Perhaps instead reinitialize from genesis/config?

crates/hashi-types/src/guardian/mod.rs
dlukegordon: Starts with `success-{seq}` now

crates/hashi-monitor/src/provisioner/mod.rs
mskd12: hmm, so key provisioners already wait for a small delay (`OTHER_SESSION_QUIET_PERIOD`) since the end of last enclave session (you can find it in `heartbeat_checks.rs`). 

the above check happens after that; so I think it probably works as is. having said that, i am leaning towards just deleting it given the mental gymnastics we needed to make sure it does not break something + its low value as a useful assert..

dlukegordon: SGTM

crates/hashi-monitor/src/provisioner/limiter_recovery.rs
mskd12: hmm, fair enough, i don't see an issue with using the genesis state as the default for this corner case.

dlukegordon (approved):

PR #582, opened 2026-05-20
[Guardian] hashi-monitor: replace 7d walk-back with S3 tree-walk for limiter recovery

mskd12:
## Summary

Follow-up to #572. Replaces the arbitrary `MAX_WALK_BACK_HOURS = 7d` time window in `recover_limiter_state` with an S3 tree-walk that finds the latest non-empty hour bucket by descending `withdraw/YYYY/MM/DD/HH/` via `CommonPrefixes`. Bounded by actual data on disk, not by a wall-clock cliff.

### Behavior changes
- No 7-day cliff. If the prior enclave's last withdrawal is older than 7 days (idle period before rotation), we still find it.
- No warn-then-fallback-to-genesis on missed walk-back. `None` now means "no Success log exists anywhere under `withdraw/`" (caller still genesis-inits in that case).
- Same clock-skew defense: after finding the latest hour bucket with `success-*`, we also read the prior hour bucket and take max-seq across both.

### Implementation
- `hashi-guardian/src/s3_logger.rs`: new `list_common_prefixes` (`list_objects_v2` with `delimiter='/'`).
- `hashi-types/src/guardian/s3_utils.rs`: new `S3HourScopedDirectory::from_path` as the inverse of `Display`; factored shared `parse_calendar` helper between `from_path` and `to_unix_seconds`.
- `hashi-monitor/src/provisioner/limiter_recovery.rs`: 4-level tree descent at each level taking the lex-greatest subdir; at each candidate hour bucket, check for `success-*` keys; on first hit, read it + one bucket back and return max-seq `post_state`.

### Tests
- 5 new mock-S3 tests for `find_latest_success_bucket` covering: empty bucket, single hour with success, latest hour empty of success backs out to earlier hour, multi-year tree picks lex-greatest leaf, intra-day backtrack.
- Wired via a new `hashi_guardian::test_utils::mock_logger_with_layout` helper that synthesizes `list_objects_v2` / `list_object_versions` responses from an in-memory key set. hashi-monitor pulls hashi-guardian's `test-utils` feature for dev — no second copy of `aws-smithy-mocks`.

## Test plan

- [x] `make check-fmt && make clippy` clean.
- [x] `cargo nextest run -p hashi-monitor provisioner::limiter_recovery` — 9 pass (4 existing + 5 new).
- [ ] Manual: run provisioner-init against a staging bucket where the prior enclave has been idle for >7 days and verify recovery finds the older Success log (would have failed pre-PR).
- [ ] Manual: run provisioner-init against a fresh bucket (no `withdraw/` keys) and verify it returns `None` quickly without scanning a 168-hour window.

### DIFF
--- a/crates/hashi-guardian/src/s3_logger.rs
+++ b/crates/hashi-guardian/src/s3_logger.rs
@@ -309,10 +309,16 @@ impl S3Logger {
                     out.insert(p.to_string());
                 }
             }
-            match response.next_continuation_token() {
-                Some(token) => continuation_token = Some(token.to_string()),
-                None => break,
+            if response.is_truncated() != Some(true) {
+                break;
             }
+            let Some(token) = response.next_continuation_token() else {
+                return Err(S3Error(format!(
+                    "Truncated response but no next_continuation_token for prefix {}",
+                    prefix
+                )));
+            };
+            continuation_token = Some(token.to_string());
         }
         Ok(out.into_iter().collect())
     }
--- a/crates/hashi-guardian/src/s3_logger.rs
+++ b/crates/hashi-guardian/src/s3_logger.rs
@@ -279,6 +279,44 @@ impl S3Logger {
         self.ensure_no_duplicates_or_deletions(prefix).await
     }
 
+    /// Lists immediate subdirectories under `prefix` (S3 `CommonPrefixes`,
+    /// returned by `list_objects_v2` with `delimiter='/'`). Used to tree-walk
+    /// the hour-partitioned withdraw layout (`withdraw/YYYY/MM/DD/HH/`)
+    /// without paginating every object key. Returned prefixes are unique and
+    /// sorted lexicographically.
+    pub async fn list_common_prefixes(&self, prefix: &str) -> GuardianResult<Vec<String>> {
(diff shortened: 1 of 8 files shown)


crates/hashi-guardian/src/s3_logger.rs
```diff
@@ -279,6 +279,44 @@ impl S3Logger {
...
+                    prefix,
+                    DisplayErrorContext(&e)
+                ))
+            })?;
+            for cp in response.common_prefixes() {
+                if let Some(p) = cp.prefix() {
+                    out.insert(p.to_string());
+                }
+            }
+            match response.next_continuation_token() {
```
dlukegordon: Should we return an error in the case that s3 is truncated, like we do [below](https://github.com/MystenLabs/hashi/pull/582/changes#diff-cd9044431f6f8bc2e54f76ee19951b7ff09ed695e2ccc8e24c93a9a6ec56da63R383-R395)?

dlukegordon (approved):

crates/hashi-guardian/src/s3_logger.rs
mskd12: thanks, edited it to follow the same defensive pattern.

PR #583, opened 2026-05-20
[guardian-hardening][limiter] reconcile to guardian on stall

0xsiddharthks:
If the local-limiter and guardian sequence numbers diverge for over `STALL_RECONCILE_TICKS` (set to 5 min wait for now) , then we reset the local limiter back to that of the guardian.

Also add a metric `guardian_limiter_reconciled_total` for tracking this.

### DIFF
--- a/crates/hashi/src/guardian_limiter.rs
+++ b/crates/hashi/src/guardian_limiter.rs
@@ -123,6 +123,16 @@ impl LocalLimiter {
         guard.next_seq += 1;
         Ok(())
     }
+
+    /// Overwrite local state with the guardian's authoritative `state`. The
+    /// guardian is the source of truth; this is the recovery path for a mirror
+    /// that has drifted and that the watcher's event stream cannot re-sync on
+    /// its own — e.g. a `WithdrawalSignedEvent` dropped across a
+    /// checkpoint-subscription reconnect, then skipped on redelivery via
+    /// `signatures.is_some()`. Config is immutable and left unchanged.
+    pub fn reconcile_to(&self, state: LimiterState) {
+        *self.state.write().unwrap() = state;
+    }
 }
 
 fn project_capacity(config: &LimiterConfig, state: &LimiterState, timestamp_secs: u64) -> u64 {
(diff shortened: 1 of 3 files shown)


crates/hashi/src/lib.rs
```diff
@@ -829,8 +806,69 @@ impl Hashi {
...
+            }
+        }
+    }
+
+    /// Snap the local limiter to the guardian once `tracker` confirms the
+    /// drift has persisted past ordinary in-flight lag.
+    async fn reconcile_guardian_limiter(
+        &self,
+        tracker: &mut guardian_limiter::LimiterStallTracker,
+    ) {
```
bmwill, line 842: When we do reconciliation do we still reserve any already consumed capacity that would have come from a pending but not yet signed transaction? or is that handled in a slightly different way?

0xsiddharthks: @bmwill i think this should be handled automatically.

lets say the guardian is at sequence `n` and currently am MPC signing for sequence `n+1` is going on while the node snaps to the guardian, then this is totally fine, since the sui-watcher will be receiving the WithdrawalSigned event for `n+1` after it's submitted onchain, at which point the new node with sync.

So any inflight request will be accounted for later once the onchain event is detected.

bmwill (approved):

crates/hashi/src/guardian_limiter.rs
```diff
@@ -144,6 +149,42 @@ pub(crate) fn should_defer_guardian_finalize(
...
+    frozen_ticks: u32,
+}
+
+impl LimiterStallTracker {
+    pub(crate) fn observe(&mut self, local_seq: u64, guardian_seq: u64) -> bool {
+        if local_seq == guardian_seq {
+            self.reset();
+            return false;
+        }
+        if self.frozen_at_seq == Some(local_seq) {
```
dlukegordon: I think this only catches "stuck behind", not "still moving but behind". If we miss one WithdrawalSignedEvent during reconnect, local_seq can keep advancing on later events while staying one behind guardian_seq forever. Since frozen_at_seq changes every tick, the tracker never fires, but the leader can still later send a stale seq to the guardian and get rejected. Should this track persistent drift instead of requiring local_seq to be frozen?

0xsiddharthks: @dlukegordon Yeah this totally makes sense. I changed my conditions a bit.

so persistent drift would kinda also not be the best here since both update at different moments. the guardian updates the moment it receives a new request, whereas the hashi node updates once it sees the onchain `WithdrawalSignedEvent`. 

instead, what I'm doing now is that instead of waiting for a frozen `local_seq`, we now reconcile if mirror hasn't caught up to where the guardian was ~5 min ago.

so this too isn't ideally the best solution, since this will effectively not fix in case an event is lost in a few hashi nodes, but the system keeps processing. however, it will work if the quorum can no longer be met because most of the nodes have fallen behind. in this case it will snap all the nodes to the current sequence.

I think this is a decent workaround for now. and once we have a reliable grpc events API, i'll be migrating to that instead, which should remove the need for this altogether.

dlukegordon (approved):

PR #584, opened 2026-05-20
Make MPC config updates atomic

zhouwfang:
#368

### DIFF
--- a/packages/hashi/sources/core/proposal/types/update_config.move
+++ b/packages/hashi/sources/core/proposal/types/update_config.move
@@ -8,30 +8,34 @@ use std::string::String;
 use sui::{clock::Clock, vec_map::VecMap};
 
 #[error]
-const EInvalidConfigEntry: vector<u8> = b"Unknown config key or wrong value type";
+const EInvalidConfigEntry: vector<u8> = b"Unknown config key or wrong value type in proposed entry";
+
+#[error]
+const ENoEntriesProvided: vector<u8> = b"UpdateConfig proposal must contain at least one entry";
 
 const THRESHOLD_BPS: u64 = 6667;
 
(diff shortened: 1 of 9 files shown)


packages/hashi/sources/core/proposal/types/update_config.move
```diff
@@ -26,11 +30,13 @@ public fun propose(
     ctx: &mut TxContext,
 ): ID {
     hashi.config().assert_version_enabled();
+    assert!(!mpc_config::is_mpc_config_key(&key), EMpcKeyMustUseDedicated);
```
Bridgerz: I'm wondering if there is a better approach for any future atomic configs (not ideal to require an assertion added to this function) 

Feels more natural to have something like this?
```
public struct UpdateConfig has copy, drop, store {
    entries: VecMap<String, Value>,
}
```

packages/hashi/sources/core/proposal/types/update_config.move
```diff
@@ -8,30 +8,34 @@ use std::string::String;
...
 }
 
 public fun execute(hashi: &mut Hashi, proposal_id: ID, clock: &Clock) {
-    let UpdateConfig { key, value } = proposal::execute(hashi, proposal_id, clock);
-    assert!(hashi.config().is_valid_config_update(&key, &value), EInvalidConfigEntry);
-    let bytes = *key.as_bytes();
-    hashi.config_mut().upsert(bytes, value);
+    let UpdateConfig { entries } = proposal::execute(hashi, proposal_id, clock);
+    let (keys, values) = entries.into_keys_values();
+    keys.zip_do!(values, |key, value| {
```
Bridgerz, line 37: Zippity doo dah!

Bridgerz (approved):

PR #585, opened 2026-05-21
[guardian-hardening][limiter] eagerly reconcile after a watcher rescrape

0xsiddharthks:
the sui checkpoint stream can disconnect, and then reconnect on the latest block, which can cause us to lose some blocks and lead to the local-limiter sequence number being out of sync from that of the guardian.

so , whenever there is a stream disconnect/reconnect, we are resyncing the local-limiter state from the latest state on the guardian.

### DIFF
--- a/crates/hashi/src/lib.rs
+++ b/crates/hashi/src/lib.rs
@@ -834,36 +834,68 @@ impl Hashi {
         }
     }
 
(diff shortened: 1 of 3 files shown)


bmwill (approved):

bmwill:
This will eventually not be necessary but it probably wont be solved till after we're in testnet

PR #586, opened 2026-05-21
Add regression test for stale deposit UTXO reuse and fix

dlukegordon:


Bridgerz (approved):

packages/hashi/sources/btc/deposit.move
bmwill, line 145: the same assert needs to be in here as well

PR #591, opened 2026-05-22
[Guardian] KP-rotation prep: N/T runtime + SecretSharingConfig

mskd12:
## Summary

Prep work that paves the way for KP rotation and aligns with the IOP-225 init-design direction. No new user-visible feature lands here — this is plumbing + cleanup.

- **`key_state/` S3 stream**: new `LogMessage::KeyState(CurrentKeyState)` written to `key_state/{seq:020}-{session_id}.json`. `setup_new_key` writes `seq=0`; future `rotate_kps` will append `seq=prev+1`. Drops the old `InitLogMessage::SetupNewKeySuccess` (no consumers).
- **Shared `split_and_encrypt_for_kps` helper**: extracts the genesis split/encrypt/commit path so `rotate_kps` can reuse it.
- **N/T runtime**: drop the compile-time `THRESHOLD = 3` / `NUM_OF_SHARES = 5` constants. Add `MIN_THRESHOLD = 2` and `validate_share_params(n, t)` (`t >= 2`, `n >= t`).
- **`SecretSharingConfig` bundle**: `{ commitments, num_shares, threshold }` with a validating constructor. Replaces the loose triple in `OperatorInitRequest`, `CurrentKeyState`, and `GuardianInfo`.
- **N/T moved to OperatorInit (per IOP-225)**: `threshold` is no longer on `ProvisionerInitState`. The enclave stores the `SecretSharingConfig` at `operator_init` time; `finalize_init` reads `t` from there. The other three IOP-225 fields (`committee`, `limiter_state`, `hashi_btc_master_pubkey`) keep their current home and will move together in the IOP-225 PR proper.
- **Misc**:
  - Bump S3 object lock duration from 5min to 7 days (TODO left for per-stream tuning).
  - Truncate `session_id_from_signing_pubkey` to 16 hex chars (full pubkey verification still happens via the signed log payload).
  - Fix inaccurate `main.rs` route-gate comment.
  - `bootstrap_guardian` gets `--num-shares` / `--threshold` flags (defaults to the prior 5, 3).
  - `hashi-monitor`: `GuardianConfig` now reads commitments from `secret_sharing_config.commitments()`; TODO left to also verify N/T against KP-side expected values.

## Test plan

- [ ] CI passes (`make fmt`, `make clippy`, full nextest)
- [ ] Manual: `hashi-guardian` `setup_new_key` → S3 has a `key_state/00000000000000000000-{session_id}.json` entry with the new `CurrentKeyState` payload
- [ ] Manual: `operator_init` → `provisioner_init` → enclave reaches fully-initialized state (covered by existing unit tests; e2e harness builds clean)
- [ ] `bootstrap_guardian` defaults to (n=5, t=3) and the existing `signet` deployment cuts over without YAML changes


crates/hashi-types/src/guardian/crypto.rs
```diff
@@ -50,17 +50,32 @@ pub struct Share {
...
-// TODO: How to rotate committee / change the below config?
-pub const THRESHOLD: usize = 3;
-pub const NUM_OF_SHARES: usize = 5;
+/// Minimum reconstruction threshold (`t > 1`). A `t == 1` setup gives any
+/// single KP unilateral control of the secret, defeating the point of secret
+/// sharing.
+pub const MIN_THRESHOLD: usize = 2;
+
+/// Validate a `(n, t)` secret-sharing parameter pair.
+pub fn validate_share_params(n: usize, t: usize) -> GuardianResult<()> {
```
dlukegordon, line 59: Nit: we should check that n is <= u16::MAX, or else someone could potentially cause the system to panic with an abnormally large number for n.

crates/hashi-guardian/src/init.rs
```diff
@@ -202,7 +205,11 @@ async fn finalize_init(
     incoming_state: ProvisionerInitState,
 ) {
     info!("Threshold reached, combining shares.");
-    let enclave_btc_keypair = combine_shares(shares).expect("Unable to combine shares");
+    let threshold = enclave
+        .secret_sharing_config()
```
dlukegordon, line 209: The caller already has ssc in scope, should we just pass it in as an arg?

crates/hashi-types/src/guardian/crypto.rs
```diff
@@ -393,6 +476,9 @@ mod tests {
         );
     }
 
+    const TEST_N: usize = 5;
+    const TEST_T: usize = 3;
```
dlukegordon: Should we also test for different values of n and t?

Also maybe we should add a test for the failing conditions of validate_share_params?

crates/hashi-types/src/guardian/test_utils.rs
```diff
@@ -25,6 +25,10 @@ use super::StandardWithdrawalRequest;
 use super::StandardWithdrawalResponse;
 use super::WithdrawalConfig;
 use super::WithdrawalID;
+
+// Default secret-sharing params used by mock_for_testing helpers.
+const TEST_N: usize = 5;
+const TEST_T: usize = 3;
```
dlukegordon, line 31: Let's move these lines below all the use statements

crates/hashi-types/src/guardian/crypto.rs
mskd12: added!

crates/hashi-types/src/guardian/crypto.rs
mskd12: done!

crates/hashi-guardian/src/init.rs
mskd12: makes sense - simplified!

crates/hashi-types/src/guardian/crypto.rs
mskd12: added!

crates/hashi-types/proto/sui/hashi/v1alpha/guardian_service.proto
```diff
@@ -54,8 +54,8 @@ message SignedGuardianInfo {
 
 // Information about the guardian enclave.
 message GuardianInfoData {
-  // Share commitments (if set). Used by key provisioners to check that right key will be used.
-  optional GuardianShareCommitments share_commitments = 1;
+  // Secret-sharing config (if set).
+  optional SecretSharingConfig secret_sharing_config = 1;
```
dlukegordon, line 58: Why are we using optional here? I see we do this in other places in the rest of the codebase but not sure why.

mskd12: Think it is a gRPC convention. IIRC without an "optional" mark, receiving clients cannot distinguish between whether the field was unset or the field was initialized to its default value. @bmwill may know better..

dlukegordon (approved):

PR #592, opened 2026-05-22
[Guardian] Add rotate_kps: ceremony-mode KP rotation

mskd12:
## Summary

Adds the ceremony-mode `rotate_kps` gRPC handler that rotates the KP set holding the existing BTC key, **without changing the key itself**.

Rotation is **operator-driven**. Each current KP produces a single HPKE-encrypted old share (bound to the agreed new-KP-set state as AAD) and hands it to the operator, who gathers T of them and makes **one** `rotate_kps` call carrying `{ encrypted_old_shares: Vec<GuardianEncryptedShare>, state }`. The enclave decrypts and verifies each submission against its stored commitments, reconstructs the BTC key in memory, re-splits it with fresh randomness, encrypts the new shares to the new KPs' OpenPGP certs, writes a `CeremonyLogMessage` at `sharing_seq = prev + 1` to `ceremony/`, and returns the new shares in a **signed `RotateKpsResponse`**. The enclave never persists the reconstructed key. Asymmetric `(old_n, old_t)` → `(new_n, new_t)` is supported.

Built on the OpenPGP/YubiKey share model: new shares are ASCII-armored PGP messages, same as `setup_new_key`.

## Ciphertexts are no longer persisted in S3

The `ceremony/` log no longer carries the encrypted shares — a `CeremonyLogMessage` now holds **only the `SecretSharingInstance`** (commitments + N + T + `sharing_seq`). KPs obtain their ciphertexts from the signed `setup_new_key` / `rotate_kps` response, or from `get_guardian_info` (served from the enclave's in-memory copy of the latest ceremony's shares), and verify their decrypted share against the instance's commitments. The commitment already binds the share value, so no separate ciphertext hash is logged — see the `TODO` on `CeremonyLogMessage` for the one case (decryption-free delivery audit) that would justify adding one.

## Notable design choices

- **One ceremony per enclave instance.** `setup_new_key` (genesis) and `rotate_kps` (rotation) are the two ceremony-mode operations, and an enclave performs exactly one. Rotation finalizes in a single call; a shared `ceremony_complete` guard (generalized from `setup_new_key`'s lock) prevents the two from interleaving and rejects a second `rotate_kps` call after completion (no re-split at the same `sharing_seq`).
- **Agreement on the rotation target is cryptographic, not cross-call.** Every old-share submission binds `state.digest()` as HPKE AAD, so a submission only decrypts if that KP authorized the exact same `RotateKpsState`. `RotateKpsState` has private fields + a validating `new()` (cert count, valid `(N, T)`, distinct certs) and **sorts the cert set into a canonical order**, so the digest is independent of the order each KP lists certs in.
- **Current-side state is operator-supplied at `operator_init`** (commitments, N, T, `sharing_seq` via `SecretSharingInstance`). KPs don't resubmit it; the enclave reads its stored instance. KPs verify off-enclave (S3 + `GuardianInfo`) that the operator gave the right state — that pre-flight lives in the KP CLI (deferred, see below).
- **`combine_shares` returns `k256::SecretKey`**, with a `k256_sk_to_btc_keypair` helper for the BTC callers.
- **`ShareCommitments::verify_share`** consolidates the share-vs-commitment check shared with `provisioner_init`. (Rotation no longer needs cross-call state-hash matching, so `provisioner_init` keeps its own inline state-hash check — there is no shared `check_or_set_state_hash`.)
- **"Ceremony" naming.** The secret-sharing log is `CeremonyLogMessage` / `LogMessage::Ceremony`, written to `ceremony/`; "setup mode" is "ceremony mode", including the env var **`SETUP_MODE` → `CEREMONY_MODE`**. (`SecretSharingInstance`, the scheme, keeps its name.)

> ⚠️ **Deployment:** the `SETUP_MODE` → `CEREMONY_MODE` env-var rename must be applied to any infra/runbook that sets it, or a ceremony enclave silently boots in normal mode.

## Testing

- `rotate.rs` flow tests (against a mock S3 logger that captures every PutObject): happy path (symmetric + asymmetric N/T), duplicate share id within the batch, below threshold, a submission bound to a mismatched state (graceful rejection — the decrypt fails on AAD; the old cross-call state-hash panic is gone), share not matching commitments, before `operator_init`, and rejection of a second call after completion. Happy-path assertions check the captured `ceremony/` log (`sharing_seq=1`, new N/T, **no ciphertexts**) and that the signed response returns one PGP-armored share per new KP.
- `setup_new_key`'s test additionally asserts its ceremony log carries no ciphertexts.
- `RotateKpsState::new` unit tests (wrong cert count, duplicate certs, digest order-independence) live in `hashi-types`.
- A `signed_rotate_kps_response_round_trip` proto round-trip test.
- **TODO:** strengthen the happy-path tests (and `setup_new_key`'s) to decrypt the armored shares with the new KPs' PGP keys and verify they reconstruct the original BTC key — needs a PGP-decrypt test helper.

## Follow-ups

- KP CLI + operator tooling that build/relay the rotation (off-enclave pre-flight checks; OOB new-KP-set verification). [IOP-409]

## Test plan

- [x] `hashi-guardian` + `hashi-types` unit tests pass locally
- [ ] CI green


crates/hashi-types/src/guardian/mod.rs
```diff
@@ -188,6 +188,24 @@ pub struct ProvisionerInitState {
...
+}
+
+/// The portion of a `RotateKpsRequest` that must agree across all T old KPs
+/// (digest-matched in-enclave). Asymmetry between old and new (`n`, `t`) is
+/// allowed.
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct RotateKpsState {
+    /// HPKE public keys for the new KP set. Length equals `new_num_shares`.
+    new_kp_pubkeys: Vec<EncPubKeyBytes>,
+    new_num_shares: usize,
```
benr-ml: why is this needed, isn't that new_kp_pubkeys.len()?

crates/hashi-types/src/guardian/mod.rs
```diff
@@ -530,6 +548,94 @@ impl ProvisionerInitRequest {
     }
 }
 
+impl RotateKpsState {
+    pub fn new(
+        new_kp_pubkeys: Vec<EncPubKeyBytes>,
```
benr-ml: nit - sort new_kp_pubkeys as the canonical rep of this set

crates/hashi-guardian/src/rotate.rs
```diff
@@ -0,0 +1,373 @@
...
+use tracing::info;
+
+/// Setup-mode rotation flow. Each *current* KP submits one of these. The
+/// enclave digest-matches `state` across the current threshold submissions
+/// (HPKE AAD on the encrypted old share is the same digest), verifies each
+/// old share against the enclave's stored commitments, and on threshold:
+///   1. reconstructs the BTC key in memory,
+///   2. re-splits it with fresh randomness for `state.new_kp_pubkeys`,
+///   3. writes `SecretSharingLogMessage` with `sharing_seq = prev + 1`.
+///      New KPs fetch their encrypted shares from there.
```
benr-ml: I'm not a big fan of putting those encryptions in public files, also for setup (maybe let's add a todo to improve that later)

crates/hashi-guardian/src/rotate.rs
```diff
@@ -0,0 +1,373 @@
...
+    if received_shares.iter().any(|s| s.id == old_share.id) {
+        return Err(InvalidInputs("Duplicate share ID".into()));
+    }
+    received_shares.push(old_share);
+    let count = received_shares.len();
+    let old_t = ssc.threshold();
+    info!("Total shares received: {count}/{old_t}.");
+
+    // 5) On threshold: reconstruct, re-split, emit new SecretSharingLogMessage
+    if count >= old_t {
```
benr-ml: isn't this can be invoked more than once? both in the happy and unhappy flows:
- (old_num_shares - old_t + 1) times - when all old parties call it
- replay attack old messages

crates/hashi-guardian/src/rotate.rs
mskd12: hmm, two options:
1. have a separate log message with just the encryptions and a shorter object lock duration (say a month) + auto delete turned on
2. rely on the operator to call `rotate_kps` & `setup_new_key` and distribute the encryptions back to kp's. kp's check if these decrypted shares match the commitments on S3 and complain if not.

crates/hashi-types/src/guardian/mod.rs
mskd12: it is technically redundant. i prefer to have it tho for ease of understanding.

crates/hashi-guardian/src/rotate.rs
mskd12: not sure what you mean by "replay attack old messages". the first is a problem that i fixed with a "finish" flag.

crates/hashi-types/src/guardian/mod.rs
benr-ml: ok, pls document that reason

crates/hashi-guardian/src/rotate.rs
benr-ml: option 2 sounds better (maybe log to S3 the hash of each encryption)

crates/hashi-guardian/src/rotate.rs
benr-ml: is there a lock/mutex on this function?
can it be that:
- n requests reach line 81 and then all of them enter the if, invoking finalize_rotation n-t-1 times

mskd12: i added a `ceremony_complete` lock on line 34 yesterday

crates/hashi-guardian/src/rotate.rs
mskd12: Hmm, a related thing we could do is have a getter endpoint in guardian that returns the ciphertexts. Even if the operator loses the ciphertexts, we can still retrieve them.

crates/hashi-types/src/guardian/mod.rs
mskd12: with the introduction of `SecretSharingParams`, the `num_shares` field has gone a level deeper. do you still want me to add some doc?

crates/hashi-types/src/guardian/mod.rs
mskd12: done!

crates/hashi-guardian/src/rotate.rs
mskd12: i made the above change in this PR itself. i did not log encryption hashes for now because share commitments serve the same purpose, no?

benr-ml: if the encryptions never reach to their destinations, can previous KPs still provision an enclave?
i'd assume that previous KPs can provision until there is a log that says that the new KPs provisioned the enclave, no?

crates/hashi-types/src/guardian/mod.rs
benr-ml: yes, for future readers (me :) )

benr-ml (approved):
left a question about new KPs, either to resolve here or in a follow up pr

crates/hashi-guardian/src/rotate.rs
mskd12: hmm, i have not yet defined that flow precisely, so it is still open how we want to do it. but i was internally leaning towards the most restrictive option: once the CeremonyLogMessage lands, all upcoming guardian-init's must use the latest shares. do you see any problem with that?

re "encryptions never reach to their destinations": that means the operator must've acted wrongly, no? why do we want to account for that?

perhaps an operator should plan it as follows:
- ensure no init's are needed for the next day to account for ceremony; if needed do an init once before ceremony starts.
- spawn guardian in ceremony mode alongside the normal guardian so withdrawals can happen like usual 
- do the rotate_kp's ceremony: collect encryptions from old kp's & share it to new kp's in ~1 day

benr-ml: > re "encryptions never reach to their destinations": that means the operator must've acted wrongly, no? why do we want to account for that?

careless operator, enclave restarted...

mskd12: Hmm, I see. Your idea seems more robust then. But we should have some policy in place to retire old KP's.. Maybe we say that old KP's can still provision the enclave for (say) ~1 week after a new CeremonyLogMessage appears? Then, we can have a "good practice norm" to do a initialization flow with new KP's immediately after `rotate_kps` finishes -- this would give us a week's buffer to resolve any issues.

benr-ml: sure, that can be configurable on the KP's script

PR #595, opened 2026-05-22
Implement kp share private keys on yubikeys with sequoia

dlukegordon:


crates/hashi-types/proto/sui/hashi/v1alpha/guardian_service.proto
```diff
@@ -110,12 +110,18 @@ message GuardianShareID {
   optional uint32 id = 1;
 }
 
-// Encrypted secret share sent to a key provisioner.
+// HPKE-encrypted secret share sent to the guardian.
 message GuardianShareEncrypted {
```
mskd12: Nit: I think we should either rename to `HpkeEncryptedShare` and `PgpEncryptedShare` (describing the scheme) or go for `GuardianEncryptedShare` and `KPEncryptedShare` (describing who it is encrypted to). This comment applies both to proto and rust struct names.

I am leaning towards the latter.. but open to either. Once we decide, let us use the same across rust & proto.

crates/hashi-guardian/src/setup.rs
```diff
@@ -64,49 +67,40 @@ pub async fn setup_new_key(
...
         let enclave = Enclave::create_operator_initialized().await;
         let verification_key = &enclave.signing_pubkey();
-        let (request, kp_private_keys) = mock_setup_new_key_request();
+        let request = mock_setup_new_key_request();
         let resp = setup_new_key(enclave.clone(), request).await.unwrap();
         let validated_resp = resp.verify(verification_key).unwrap();
         assert_eq!(validated_resp.encrypted_shares.len(), TEST_N);
+        assert_eq!(validated_resp.share_commitments.len(), TEST_N);
 
-        for (enc_share, sk) in validated_resp
```
mskd12, line 100: Is it hard to recreate the full test pipeline with Sequoia? I also did not see any tests in `crypto.rs` for the new sequoia-pgp encryption code.

crates/hashi-types/src/guardian/crypto.rs
```diff
@@ -213,6 +232,17 @@ impl GuardianEncKeyPair {
     }
 }
 
+impl PgpPublicCert {
+    pub fn new(armored: String) -> GuardianResult<Self> {
```
mskd12, line 237: Nit: You could record the recipients from `validate` and use that later to avoid double parsing, right?

crates/hashi-types/src/guardian/crypto.rs
```diff
@@ -263,6 +293,56 @@ pub fn decrypt(
...
+        .supported()
+        .alive()
+        .revoked(false)
+        .for_transport_encryption()
+        .next()
+        .ok_or_else(|| InvalidInputs("OpenPGP certificate has no usable encryption key".into()))?;
+    Ok(())
+}
+
+fn pgp_encrypt_armored(plaintext: &[u8], cert: &PgpPublicCert) -> GuardianResult<String> {
```
mskd12, line 312: Let's add a simple encrypt/decrypt test for this function?

mskd12 (commented):
LGTM! Just left some small comments.

crates/hashi-guardian/src/setup.rs
dlukegordon: I will add!

0xsiddharthks (approved):

Cargo.toml
```diff
@@ -55,6 +55,9 @@ tracing-subscriber = { version = "0.3.19", features = ["env-filter", "json"] }
 # Upstream issues: https://github.com/str4d/rage/issues/568
 #                  https://github.com/kellpossible/cargo-i18n/issues/150
 age = { git = "https://github.com/str4d/rage.git", rev = "92f437bc2a061312fa6633bb70c91eef91142fd4", features = ["cli-common", "plugin"] }
```
bmwill, line 58: If we're using sequoia for guardian ops, do we want to revisit using it for backups instead of age?

dlukegordon: Yes we do. Now that I know about oct, it may be worth having the validators just use pgp as well, and drop the age dep.

mskd12 (approved):

PR #597, opened 2026-05-22
[Guardian] cleanups

mskd12:
## Summary

Bundle of `hashi-guardian` / `hashi-types` cleanups motivated by recent #591/#592 design discussions. No production behavior change beyond the setup-mode invariants.

- **`SecretSharingConfig` → `SecretSharingInstance`** across the proto, regenerated bindings, all Rust source, and README. The type represents a versioned commitment set (one per `sharing_seq`), not a "config."
- **`SecretSharingParams { num_shares, threshold }`** as the validated `(n, t)` parameter type. Both `SecretSharingInstance` and `SetupNewKeyRequest` store `params: SecretSharingParams` internally; `params()` accessor exposes it. Existing `num_shares()` / `threshold()` getters forward through `params`.
- **Drop `Result` from `split_secret`** — its only failure was the internal `validate_share_params?`, which duplicates the validation already enforced at the request-construction boundary.
- **`split_and_encrypt_for_kps` keeps `Result`** for the pubkey-count check (genuinely a precondition between two args). `setup.rs` `.expect`s with a comment naming the `SetupNewKeyRequest::new` invariant. Same shape applied to `SecretSharingInstance::new` in `setup.rs` after the change.
- **`setup_complete` one-shot gate**: `Scratchpad.setup_complete: OnceLock<()>` + `Enclave::is_setup_complete()`. `setup_new_key` gates on it at entry and sets it on success — single ceremony per enclave instance.
- **`rpc.rs` cleanup**: `require_setup_mode` / `require_normal_mode` helpers; the three main-resident handlers (`setup_new_key`, `provisioner_init`, `standard_withdrawal`) converted to one-line gates.
- **`provisioner_init`**: clear `received_shares` after threshold-reach so the scratchpad doesn't retain share material post-init.
- **TODO** in `main.rs` for skipping the heartbeat writer in setup mode (separate follow-up; setup-mode heartbeats confuse the normal-mode `heartbeat_audit`).

## Test plan

- [ ] `make fmt` and `make clippy` clean
- [ ] `cargo nextest run -p hashi-types -p hashi-guardian` passes (covers the crypto signature changes + new `setup_complete` gate)
- [ ] Manual smoke: `setup_new_key` rejects a second call on the same enclave instance


crates/hashi-types/src/guardian/crypto.rs
```diff
@@ -50,25 +50,48 @@ pub struct Share {
...
+pub const MAX_NUM_SHARES: usize = u16::MAX as usize;
 
-/// Validate a `(n, t)` secret-sharing parameter pair.
-pub fn validate_share_params(n: usize, t: usize) -> GuardianResult<()> {
-    if t < MIN_THRESHOLD {
-        return Err(InvalidInputs(format!(
-            "threshold {t} below minimum {MIN_THRESHOLD}"
-        )));
+/// Validated `(n, t)` secret-sharing parameters.
+#[derive(Serialize, Deserialize, Copy, Clone, Debug, PartialEq, Eq)]
```
dlukegordon, line 67: Can we remove `Deserialize` from `SecretSharingParams`, `SecretSharingInstance`, `LogRecord`, etc? Doesn't seem like it's needed.

crates/hashi-guardian/src/enclave.rs
```diff
@@ -63,15 +63,15 @@ pub struct EnclaveState {
     rate_limiter: OnceLock<Arc<tokio::sync::Mutex<RateLimiter>>>,
 }
 
-/// Scratchpad used only during initialization.
-/// Note: We don't clear it post-init because it does not have a lot of data.
+/// Scratchpad used only during initialization. `shares` is cleared once the
+/// init flow that gathered it completes; the OnceLock flags retain their state.
 #[derive(Default)]
 pub struct Scratchpad {
```
dlukegordon, line 69: I think theoretically if there were concurrent calls to setup_new_key we could have a race condition:

> The code checks is_setup_complete() before generating the key and writing the secret_sharing log, but only sets setup_complete after the S3 write. Two concurrent requests can both pass the check, generate different BTC keys, and both attempt to write secret_sharing/00000000000000000000-{session_id}.json. One request may return setup_complete already set, but the losing request has already emitted a conflicting secret-sharing log/version. If object versioning is enabled, this creates multiple immutable versions for the same logical genesis entry; if consumers rely on “latest”, that is dangerous.

We could add a mutex to scratchpad to prevent this:

```
pub async fn setup_new_key(
    enclave: Arc<Enclave>,
    request: SetupNewKeyRequest,
) -> GuardianResult<GuardianSigned<SetupNewKeyResponse>> {
    let _setup_guard = enclave.scratchpad.setup_new_key_lock.lock().await;
    info!("/setup_new_key - Received request.");
```

crates/hashi-types/src/guardian/crypto.rs
```diff
@@ -50,25 +50,48 @@ pub struct Share {
...
+            )));
+        }
+        if num_shares < threshold {
+            return Err(InvalidInputs(format!(
+                "num_shares {num_shares} below threshold {threshold}"
+            )));
+        }
+        if num_shares > MAX_NUM_SHARES {
+            return Err(InvalidInputs(format!(
+                "{num_shares} must be less than u16::MAX"
```
dlukegordon: `must be at most`

crates/hashi-types/src/guardian/crypto.rs
```diff
@@ -593,24 +633,24 @@ mod tests {
...
             );
         }
         // num_shares < threshold.
-        assert!(validate_share_params(2, 3).is_err());
-        assert!(validate_share_params(5, 7).is_err());
-        // num_shares > u16::MAX.
-        assert!(validate_share_params(u16::MAX as usize + 1, 3).is_err());
+        assert!(SecretSharingParams::new(2, 3).is_err());
+        assert!(SecretSharingParams::new(5, 7).is_err());
+        // num_shares > threshold.
```
dlukegordon: `// num_shares > MAX_NUM_SHARES`

crates/hashi-guardian/src/enclave.rs
```diff
@@ -476,16 +484,16 @@ impl Enclave {
...
+    pub fn secret_sharing_instance(&self) -> GuardianResult<&SecretSharingInstance> {
         self.scratchpad
-            .secret_sharing_config
+            .secret_sharing_instance
             .get()
             .ok_or(InvalidInputs("Secret sharing config not set".into()))
     }
 
-    pub fn set_secret_sharing_config(&self, cfg: SecretSharingConfig) -> GuardianResult<()> {
+    pub fn set_secret_sharing_instance(&self, cfg: SecretSharingInstance) -> GuardianResult<()> {
```
dlukegordon: Rename cfg to instance?

crates/hashi-guardian/src/enclave.rs
```diff
@@ -476,16 +484,16 @@ impl Enclave {
         &self.scratchpad.shares
     }
 
-    pub fn secret_sharing_config(&self) -> GuardianResult<&SecretSharingConfig> {
+    pub fn secret_sharing_instance(&self) -> GuardianResult<&SecretSharingInstance> {
         self.scratchpad
-            .secret_sharing_config
+            .secret_sharing_instance
             .get()
             .ok_or(InvalidInputs("Secret sharing config not set".into()))
```
dlukegordon: Rename config to instance?

crates/hashi-guardian/src/enclave.rs
```diff
@@ -476,16 +484,16 @@ impl Enclave {
...
     }
 
-    pub fn set_secret_sharing_config(&self, cfg: SecretSharingConfig) -> GuardianResult<()> {
+    pub fn set_secret_sharing_instance(&self, cfg: SecretSharingInstance) -> GuardianResult<()> {
         self.scratchpad
-            .secret_sharing_config
+            .secret_sharing_instance
             .set(cfg)
             .map_err(|_| InvalidInputs("Secret sharing config already set".into()))
```
dlukegordon: Rename config to instance?

crates/hashi-guardian/src/init.rs
```diff
@@ -53,18 +53,18 @@ pub async fn operator_init(
 
     info!(
         "Storing secret-sharing config: n={}, t={}, {} commitments.",
```
dlukegordon: Rename config to instance? There are a few other places below as well.

crates/hashi-guardian/src/enclave.rs
```diff
@@ -63,15 +63,15 @@ pub struct EnclaveState {
     rate_limiter: OnceLock<Arc<tokio::sync::Mutex<RateLimiter>>>,
 }
 
-/// Scratchpad used only during initialization.
-/// Note: We don't clear it post-init because it does not have a lot of data.
+/// Scratchpad used only during initialization. `shares` is cleared once the
+/// init flow that gathered it completes; the OnceLock flags retain their state.
```
dlukegordon: `init flow that gathered it (provisioner_init)`

crates/hashi-guardian/src/enclave.rs
dlukegordon: We could also add a test that checks that we can't run two simultaneously, or one after another.

mskd12: Hmm, good idea! I collapsed the lock and the existing flag into a Mutex<bool>

crates/hashi-types/src/guardian/crypto.rs
mskd12: Deserialize helps clients (e.g., hashi-monitor) read the LogRecord's from S3.

dlukegordon: Fine with keeping it but fyi I was able to delete the deserialize derivations and it didn't cause any compilation errors so I think it's not currently being used

dlukegordon (approved):

PR #602, opened 2026-05-27
[guardian][onchain] verify guardian /info pubkey against on-chain config

0xsiddharthks:
builds on top of #549, which allows publishing the guardian configuration onchain.

### hashi
- At hashi startup, compare the live guardian `signing_pub_key` against on-chain `guardian_public_key` on both the bootstrap and the periodic-reconcile paths in `Hashi::start`. 
- on mismatch, the retry loop keeps running so a corrective `UpdateGuardian` proposal heals the node without a restart.

### observability
- add new metric `guardian_bootstrap_outcomes_total{outcome="key_mismatch"}` to track failures

### publishing the config during deployments
- `set_guardian` asserts a 32-byte key on-chain and at the `hashi publish --guardian-public-key` flag boundary.
- `internal-tools fetch-guardian-info --endpoint <url>` prints the hex pubkey from a deployed guardian's `/info`
- MystenLabs/sui-operations#8055 uses this to pipe the `--guardian-public-key`

### DIFF
--- a/crates/hashi/src/lib.rs
+++ b/crates/hashi/src/lib.rs
@@ -781,6 +781,9 @@ impl Hashi {
                 return false;
             }
         };
+        if !self.verify_guardian_signing_pubkey(&info.signing_pub_key) {
+            return false;
+        }
         let _ = self.guardian_signing_pubkey.set(Some(info.signing_pub_key));
         let (Some(state), Some(config)) = (info.limiter_state, info.limiter_config) else {
             self.metrics.record_guardian_bootstrap_outcome(
@@ -835,7 +838,9 @@ impl Hashi {
     }
 
     /// Fetch the guardian's authoritative `LimiterState` and the live local
-    /// limiter handle. `None` if the limiter isn't seeded or the RPC fails.
+    /// limiter handle. `None` if the limiter isn't seeded, the RPC fails, or
+    /// the guardian's signing key doesn't match the on-chain expected key
+    /// (treated as fatal: a stale or wrong guardian must not steer the limiter).
     async fn guardian_limiter_and_state(
         &self,
     ) -> Option<(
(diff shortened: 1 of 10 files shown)


crates/hashi/src/lib.rs
```diff
@@ -844,10 +847,28 @@ impl Hashi {
...
+        &self,
+        signing_pub_key: &hashi_types::guardian::GuardianPubKey,
+    ) -> bool {
+        let expected = self
+            .onchain_state()
+            .state()
+            .hashi()
+            .config
+            .guardian_public_key()
+            .map(<[u8]>::to_vec);
```
dlukegordon: You could add a helper method for this in `crates/hashi/src/onchain/mod.rs` around lines 465-497, we have a lot of similar helpers,

crates/hashi/src/lib.rs
```diff
@@ -780,6 +780,9 @@ impl Hashi {
                 return false;
             }
         };
+        if !self.verify_guardian_signing_pubkey(&info.signing_pub_key) {
+            return false;
+        }
```
dlukegordon, line 786: > Guardian key rotation breaks withdrawals until restart  
> crates/hashi/src/lib.rs:783-786 sets guardian_signing_pubkey only during initial bootstrap, and the field is a OnceLock. Later reconciliation verifies the current on-chain key at crates/hashi/src/lib.rs:850-853, but it does not update the cached signing key. Withdrawal responses are still verified with the stale cached key at crates/hashi/src/leader/mod.rs:2208-2223. If UpdateGuardian rotates the guardian key at the same endpoint, limiter reconciliation can accept the new key, but guardian withdrawal response verification keeps using the old key and fails until the node restarts.

crates/hashi/src/lib.rs
```diff
@@ -844,10 +847,28 @@ impl Hashi {
...
         let info_pb = self.fetch_guardian_info().await?;
         let info = hashi_types::guardian::GetGuardianInfoResponse::try_from(info_pb).ok()?;
+        if !self.verify_guardian_signing_pubkey(&info.signing_pub_key) {
+            return None;
+        }
         let state = info.limiter_state?;
         Some((limiter, state))
     }
 
+    /// Re-read on each call so `UpdateGuardian` heals without a restart.
```
dlukegordon: > UpdateGuardian URL changes are still not picked up without restart  
> The guardian client is built once from on-chain config during startup at crates/hashi/src/lib.rs:653-683. The new verification re-reads the on-chain key each time at crates/hashi/src/lib.rs:857-869, but it still fetches /info through the old cached GuardianClient. If governance updates both URL and key, running nodes keep contacting the old guardian, hit key_mismatch, and stop seeding/reconciling until restart. The comment at crates/hashi/src/lib.rs:857 says UpdateGuardian heals without restart, but that is only true for the key lookup, not the endpoint.

packages/hashi/sources/core/config/config.move
```diff
@@ -93,6 +96,7 @@ public(package) fun guardian_public_key(self: &Config): Option<vector<u8>> {
 }
 
 public(package) fun set_guardian(self: &mut Config, url: String, public_key: vector<u8>) {
+    assert!(public_key.length() == GUARDIAN_PUBLIC_KEY_LEN, EBadGuardianPublicKeyLength);
```
dlukegordon: This is called only during execution at packages/hashi/sources/core/proposal/types/update_guardian.move:36-38. update_guardian::propose accepts any vector<u8> at packages/hashi/sources/core/proposal/types/update_guardian.move:17-29, and the CLI proposal path decodes hex without checking length at crates/hashi/src/cli/commands/proposal.rs:680-695. So a malformed proposal can be created, voted on, and only fail when executed. This should probably reject at proposal creation too, matching the publish CLI validation.

crates/hashi/src/lib.rs
0xsiddharthks: Yeah, I think this is fine. we won't be swapping guardian's without doing a redeploy of hashi itself.

crates/hashi/src/lib.rs
0xsiddharthks: i think this should be fine. we don't be swapping the guardian without a hashi redeploy.

dlukegordon (approved):

PR #603, opened 2026-05-27
[guardian][onchain] publish guardian BTC pubkey at deploy time

0xsiddharthks:
### deployment and publishing config
- New `Config::guardian_btc_public_key` (32-byte x-only) + `set_guardian_btc_public_key` setter on `Config`. Immutable once set — re-setting must use the same key, so a buggy reconfig can't invalidate every 2-of-2 deposit address.
- `GuardianInfoData.enclave_btc_pubkey` populated after `provisioner_init` (consumed by the on-chain pin check added in #602).
- `finish_publish` takes `guardian_btc_public_key: Option<vector<u8>>` and pins it when present. Same lifecycle as `guardian_url` / `guardian_public_key`: set once at publish time, no later mutation.

The companion deploy workflow (sui-operations#8057) needs to fetch the guardian's BTC pubkey before `hashi publish` and pipe it through `--guardian-btc-public-key`. Follow-up: rework bootstrap-guardian so master-key generation happens pre-publish.

### utils
- add 2-of-2 taproot helpers
  - `two_of_two_taproot_script_path_address`
  - `_spend_artifacts`

## Test plan
- [x] `sui move test --path packages/hashi --build-env testnet`
- [x] `cargo nextest run -p hashi --lib`
- [ ] Devnet rollout once sui-operations#8057 is reworked


packages/hashi/sources/core/config/config.move
```diff
@@ -95,12 +105,40 @@ public(package) fun guardian_public_key(self: &Config): Option<vector<u8>> {
...
+public(package) fun deployer(self: &Config): Option<address> {
+    self.try_get(DEPLOYER_KEY).map!(|v| v.as_address())
+}
+
+/// Pinned at publish time to the address that called `finish_publish`.
+/// One-time set; subsequent calls abort.
+public(package) fun set_deployer(self: &mut Config, deployer: address) {
+    assert!(self.deployer().is_none(), EDeployerAlreadySet);
+    self.upsert(DEPLOYER_KEY, config_value::new_address(deployer));
+}
```
bmwill: I would prefer if we don't have a deployer and instead require all params set upfront.

packages/hashi/sources/hashi.move
```diff
@@ -153,6 +163,19 @@ entry fun finish_publish(
...
+entry fun set_guardian_btc_public_key(
+    self: &mut Hashi,
+    btc_public_key: vector<u8>,
+    ctx: &TxContext,
+) {
+    self.config.assert_version_enabled();
+    let deployer = self.config().deployer().destroy_or!(abort ENotDeployer);
+    assert!(ctx.sender() == deployer, ENotDeployer);
+    self.config_mut().set_guardian_btc_public_key(btc_public_key);
+}
```
bmwill: As stated above i would prefer if we don't allow the deployer this extra latitude, lets just require that they set the btc pubkey at the same time in finish_publish

bmwill (dismissed):

packages/hashi/sources/core/config/config.move
0xsiddharthks: @bmwill gotcha.

The BTC pubkey now flows through `finish_publish` only, same shape as `guardian_url`/`guardian_public_key`.

#8057 generates it before `hashi publish` so it's passed in via `--guardian-btc-public-key`, instead of a second transaction.

packages/hashi/sources/hashi.move
0xsiddharthks: fixed. same as above :)

crates/internal-tools/src/bootstrap_guardian.rs
```diff
@@ -96,7 +103,12 @@ pub async fn run(args: Args, onchain_state: &OnchainState) -> Result<()> {
     let mut rng = thread_rng();
     let n = args.num_shares;
     let t = args.threshold;
-    let material = generate_share_material(n, t, &mut rng)?;
+    let existing_secret = args
+        .master_secret_hex
+        .as_deref()
+        .map(parse_master_secret)
+        .transpose()?;
+    let material = generate_share_material(n, t, &mut rng, existing_secret)?;
```
dlukegordon, line 111: Should we check here that it matches onchain_state.state().hashi().config.guardian_btc_public_key()?

crates/hashi/src/cli/mod.rs
```diff
@@ -908,6 +913,19 @@ pub async fn run_publish(opts: PublishOpts) -> anyhow::Result<()> {
     let mut client = sui_rpc::Client::new(&opts.sui_rpc_url)?;
 
     // Build optional guardian config
+    let guardian_btc_public_key = match opts.guardian_btc_public_key {
```
dlukegordon, line 916: Maybe instead of just checking the length, we should also check that it encodes a valid bitcoin pubkey?

hashi_types::guardian::BitcoinPubkey::from_slice(&bytes)

packages/hashi/sources/hashi.move
```diff
@@ -130,6 +131,12 @@ entry fun finish_publish(
         guardian_public_key.destroy_none();
     };
 
+    if (guardian_btc_public_key.is_some()) {
```
dlukegordon: Should we merge this if block into the above if block? That way we either set all or nothing.

crates/internal-tools/src/generate_master_key.rs
```diff
@@ -0,0 +1,57 @@
...
+    let secret_hex = hex::encode(secret_bytes);
+    let pubkey_hex = hex::encode(master_pubkey.serialize());
+
+    match args.format {
+        OutputFormat::Json => {
+            println!("{{\"secret_hex\":\"{secret_hex}\",\"pubkey_hex\":\"{pubkey_hex}\"}}");
+        }
+        OutputFormat::Env => {
+            println!("MASTER_SECRET_HEX={secret_hex}");
+            println!("MASTER_PUBKEY_HEX={pubkey_hex}");
```
dlukegordon: I think these need to be prefixed with `HASHI_` to match the env var names?

0xsiddharthks: Thanks! Just fixed this.

packages/hashi/sources/hashi.move
0xsiddharthks: Yeah. i've merged the two blocks

crates/internal-tools/src/bootstrap_guardian.rs
0xsiddharthks: gotcha! Yeah, added an ensure clause here to confirm the onchain state matches.

crates/internal-tools/src/bootstrap_guardian.rs
```diff
@@ -112,6 +112,26 @@ pub async fn run(args: Args, onchain_state: &OnchainState) -> Result<()> {
     tracing::info!(master_pubkey = %hex::encode(material.master_pubkey.serialize()),
         n, t, "generated share material");
 
+    // Derived key must match the pubkey pinned on-chain at publish; a wrong or
+    // missing --master-secret-hex would otherwise provision a key the chain rejects.
+    let onchain_btc_pubkey = onchain_state
+        .state()
+        .hashi()
+        .config
+        .guardian_btc_public_key()
```
dlukegordon: Don't we have a helper for this?

packages/hashi/sources/hashi.move
```diff
@@ -125,9 +126,15 @@ entry fun finish_publish(
                 guardian_url.destroy_some(),
                 guardian_public_key.destroy_some(),
             );
+        if (guardian_btc_public_key.is_some()) {
```
dlukegordon: Do we want to be able to set guardian_public_key and guardian_url without setting guardian_btc_public_key?

dlukegordon (approved):

PR #604, opened 2026-05-27
[guardian][onchain] 2-of-2 (mpc, guardian) taproot deposits + make guardian required

0xsiddharthks:
Cuts the bridge over to 2-of-2 deposit addresses and makes the guardian required end-to-end. The mechanical `guardian_signatures` field plumbing is split into #620 (below); this PR is the load-bearing crypto + policy.

## Review guide (by commit)
- **require guardian config in finish_publish** — drop the `Option<>` guardian args.
- **derive deposit addresses as 2-of-2** — `tr(NUMS, multi_a(2, guardian_btc_pubkey, derive(mpc, path)))`.
- **build 2-of-2 witness** — `[hashi_sig, guardian_sig, leaf_script, control_block]` per input (multi_a satisfier order).
- **capture guardian sigs + post both arrays** — leader gets per-input enclave sigs from the guardian; every follower Schnorr-verifies both arrays against the same sighash before BLS-signing, and the cert binds both (a malicious leader can't pair good MPC sigs with garbage guardian sigs).
- **make guardian required** — `Hashi::start` requires an endpoint; publish/CLI flags required; e2e always spins up a guardian.
- **MPC-sign change-UTXO at the zero path** — change outputs sign `derive(G,[0;32])` to match the 2-of-2 change leaf.

## Notes
- **Start-time enforcement**: `Hashi::start` fails fast only if the guardian *endpoint* is missing/unreachable. The BTC-pubkey match against on-chain runs in the non-fatal, retrying background bootstrap (`verify_and_pin_guardian_btc_pubkey`), not at start. (Corrects the earlier draft wording.)
- **Breaking on-chain change**: `WithdrawalTransaction` / events / messages gain `guardian_signatures` → upgrade-incompatible; needs a fresh publish + devnet wipe. All deposit addresses change (single-key → 2-of-2), so pre-existing UTXOs are stranded — fresh deploy only.
- Stacked on #620 (plumbing); the raw-`G` derivation fix is #609 above. #604 alone is red on the withdrawal e2e until #609.

## Test plan
- [x] `sui move test --build-env testnet`
- [x] `cargo nextest` with #609 on top (rotation + withdrawal e2e green; epoch-boundary `#[ignore]`d pending guardian wid-idempotency)

### DIFF
--- a/crates/e2e-tests/src/e2e_flow.rs
+++ b/crates/e2e-tests/src/e2e_flow.rs
@@ -351,42 +351,23 @@ mod tests {
 
     #[tokio::test]
     async fn test_bitcoin_withdrawal_e2e_flow() -> Result<()> {
-        run_bitcoin_withdrawal_e2e(false).await
-    }
-
-    #[tokio::test]
-    async fn test_bitcoin_withdrawal_with_guardian_e2e_flow() -> Result<()> {
-        run_bitcoin_withdrawal_e2e(true).await
-    }
-
-    async fn run_bitcoin_withdrawal_e2e(with_guardian: bool) -> Result<()> {
         init_test_logging();
-        let label = if with_guardian {
-            " (with guardian)"
-        } else {
-            ""
-        };
-        info!("=== Starting Bitcoin Withdrawal E2E Test{label} ===");
+        info!("=== Starting Bitcoin Withdrawal E2E Test ===");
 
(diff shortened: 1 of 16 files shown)


0xsiddharthks:
Consolidated into #609, which now carries the whole guardian 2-of-2 stack as one PR. Closing this in favour of that.

PR #605, opened 2026-05-27
[guardian] add cross-language 2-of-2 test vectors

0xsiddharthks:
Stacked on #609 (rebased off #604).

Pins the `(derived child, regtest/signet address, leaf-script hex, tap-leaf hash)` outputs of `two_of_two_taproot_script_path_address` / `derive_hashi_child_pubkey` for two regimes:

- **even-y master** — `(enclave=[1u8;32], hashi=[2u8;32])` across three derivation paths (zero, `[1u8;32]`, mixed-byte). The parent's y-parity is forced even via `hashi_master_g_from_xonly`, matching the SDK's `RUST_HASHI_MASTER_SEC1_EVEN_Y` constant.
- **odd-y master** — `hashi=[4u8;32]`, the first scalar in `3..=255` whose `s·G` lands on odd y, derived against the raw `G` with no parity forcing. This locks the production path #609 fixed (the leaf script formerly disagreed with the MPC signature for ~50% of DKG outputs).

The same exact bytes are asserted in `hashi-ts-sdk/packages/hashi/test/unit/bitcoin.test.ts` (MystenLabs/hashi-ts-sdk#32), so drift between the Rust impl and the SDK fails CI on both sides.

The deposit-address rollout is unforgiving: if the SDK ever produces an address different from the bridge's `validate_deposit_request_derivation_path`, user BTC is sent into a black hole. These vectors are the canary.

## Test plan
- [x] `cargo nextest run -p hashi-types cross_lang_2of2_test_vectors cross_lang_2of2_test_vectors_odd_y` — pass in CI
- [x] `make fmt && make clippy` — clean
- [x] Rebased onto #609; PR base updated


0xsiddharthks:
Consolidated into #609, which now carries the whole guardian 2-of-2 stack as one PR. Closing this in favour of that.

PR #607, opened 2026-05-27
[guardian][onchain] entry fn to publish BTC pubkey after init

0xsiddharthks:
Stacked on #603. Adds the deploy-time hook for landing the guardian BTC pubkey on chain when it isn't available at publish time.

- `hashi::set_guardian_btc_public_key` entry fn — gated to the address that called `finish_publish` (pinned in `Config` as a new `deployer` field). The underlying `public(package)` setter rejects rotation, so this stops adversarial front-runs from bricking deposits with a wrong key.
- `hashi set-guardian-btc-public-key` CLI subcommand — `--keypair`, `--package-id`, `--hashi-object-id`, `--btc-public-key`. Builds the PTB and executes it.
- Move + Rust tests covering the deployer pin, the wrong-sender path, rotation rejection, and idempotency.

The deploy workflow (sui-operations #8057) will fetch the BTC pubkey from the guardian after `provisioner_init` and pipe it through this CLI.

## Test plan
- [x] `sui move test --path packages/hashi --build-env testnet` — 113 passing
- [x] `cargo nextest run -p hashi --lib` — 359 passing
- [ ] Devnet: deploy via sui-operations#8057, verify `Config.guardian_btc_public_key` is populated after the post-init step

### DIFF
--- a/crates/hashi/src/cli/mod.rs
+++ b/crates/hashi/src/cli/mod.rs
@@ -541,6 +541,37 @@ pub struct PublishOpts {
     pub yes: bool,
 }
 
+/// Options for the `set-guardian-btc-public-key` subcommand.
+///
+/// Publishes the guardian's enclave BTC pubkey on the already-published
+/// Hashi object. Signed by the original deployer keypair.
+#[derive(Args)]
+pub struct SetGuardianBtcPublicKeyOpts {
+    /// Sui RPC endpoint URL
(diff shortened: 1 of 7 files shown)


0xsiddharthks:
Folded into #603 as a single follow-up commit — the two pieces (publish the BTC pubkey on-chain + the post-init entry fn that actually lets the deploy workflow do it) belong together.

PR #609, opened 2026-05-27
[guardian][onchain] mandatory 2-of-2 (mpc, guardian) taproot deposits + end-to-end guardian signatures

0xsiddharthks:
Cuts the bridge over to **mandatory 2-of-2 (MPC + guardian) taproot deposits** and threads guardian signatures through the withdrawal flow end-to-end. Consolidates the guardian-2of2 stack (#620, #604, #609, #617, #618, #605) into one PR — CI only goes green with all of it.

## What changed
- **2-of-2 deposits** — addresses derive as `tr(NUMS, multi_a(2, guardian_btc_pubkey, derive(mpc_G, path)))`. The child is derived from the **raw MPC `G`** (y-parity preserved) so the leaf script matches what the MPC signs — fixes the "Invalid Schnorr signature" failures that hit ~50% of odd-y DKG outputs.
- **Guardian required** — `finish_publish` drops the `Option<>` guardian args; `Hashi::start` fails fast without a guardian endpoint; e2e always spins up a guardian.
- **Guardian signatures threaded end-to-end** — `guardian_signatures` added to the Move `WithdrawalTransaction` / `WithdrawalSignedEvent` / `WithdrawalSignedMessage`, the Rust `move_types` mirror, the `SignWithdrawalTxSigningRequest` proto, the bridge-service parser, and the watcher. The leader captures per-input enclave sigs; every follower Schnorr-verifies both arrays against the same sighash before BLS-signing, and the cert binds both.
- **2-of-2 witness** — `[hashi_sig, guardian_sig, leaf_script, control_block]` per input; change outputs sign `derive(G, [0;32])` to match the change leaf.
- **Cross-language test vectors** pin `(derived child, address, leaf-script, tap-leaf hash)` for even-y and odd-y masters, asserted byte-for-byte against the hashi-ts-sdk (MystenLabs/hashi-ts-sdk#32).
- **Cleanup** — removed the dead x-only helpers (`get_hashi_pubkey` / `get_onchain_mpc_pubkey`) that would reintroduce the odd-y bug; fixed guardian-always-on regressions in the rotation e2e tests.

## Notes
- **Breaking on-chain change**: events/messages gain `guardian_signatures` and all deposit addresses change (single-key → 2-of-2) → upgrade-incompatible; fresh publish + devnet wipe only, pre-existing UTXOs are stranded.
- `test_withdrawal_signs_across_epoch_boundary` is `#[ignore]`d (real cross-epoch seq-replay; fix is guardian wid-idempotency, out of scope for this stack).
- Follow-up: `bootstrap_guardian` / `hashi-monitor` provisioner still load the master as 32-byte x-only and rebuild `G` via `with_even_y` (TODO in `provisioner/mod.rs`) — fine for even-y DKG outputs, needs a 33-byte config before production.

## Test plan
Full local CI green: `make check-fmt`, `make buf-lint`, `make clippy`, `make proto` (no diff), `cargo nextest run --workspace --all-features --profile ci`, `cargo test --workspace --all-features --doc`, `make test-move`, `make doc`. The 15 commits are kept intact for a review walk-through.


crates/hashi-types/src/guardian/bitcoin_utils.rs
```diff
@@ -28,8 +28,8 @@ use bitcoin::taproot::TaprootSpendInfo;
 use bitcoin::transaction::Version;
 use bitcoin::*;
 use fastcrypto::serde_helpers::ToFromByteArray;
-use fastcrypto_tbls::threshold_schnorr;
 use fastcrypto_tbls::threshold_schnorr::Address as SuiAddress;
+use fastcrypto_tbls::threshold_schnorr::G;
```
mskd12: nit: reuse `HashiMasterG` instead of `G`?

crates/hashi-types/src/guardian/bitcoin_utils.rs
```diff
@@ -893,4 +895,38 @@ mod bitcoin_tests {
...
     }
+
+    // Regression: an odd-y MPC master key used to silently disagree between
+    // the witness leaf script (which reconstructed an even-y parent from the
+    // x-only bytes) and the MPC signature (which signs against
+    // `derive_verifying_key(raw_g, path)`), giving "Invalid Schnorr
+    // signature" at Bitcoin. `derive_hashi_child_pubkey` consumes the raw G
+    // so the two derivations now agree regardless of y-parity.
+    #[test]
+    fn derive_hashi_child_pubkey_matches_mpc_for_odd_y_master() {
```
mskd12: Seems like a pointless test

crates/hashi/src/deposits.rs
mskd12: It seems a little dangerous to have this and the next function (`get_onchain_mpc_pubkey`) as they are also dropping the parity bit and child derivations made from the output of these functions will fail the same way. Remove them for defensive reasons or alter the return type?

mskd12 (approved):
One small comment. LGTM!

0xsiddharthks:
Thanks @mskd12 — addressed in 400c4aa50:

1. **`G` → `HashiMasterG`**: done — `bitcoin_utils` uses the re-exported alias throughout, dropped the second `G` import.
2. **"pointless test"**: agreed, it was tautological — it compared `derive_hashi_child_pubkey` against the `derive_verifying_key` it wraps, so it passed regardless of y-parity. Rewrote it to exercise the real leaf path: for an odd-y master it builds the 2-of-2 leaf via `two_of_two_taproot_script_path_spend_artifacts` and asserts the leaf embeds the MPC-signed child, and that the old even-y-forced child is **not** present (so reverting to even-y forcing would fail it).
3. **parity-dropping helpers**: `get_hashi_pubkey` / `get_onchain_mpc_pubkey` are unused after this fix; removed them in follow-up #618 to keep this PR scoped to the derivation fix.

PR #618, opened 2026-05-28
[guardian] remove dead x-only MPC pubkey helpers

0xsiddharthks:
`get_hashi_pubkey` / `get_onchain_mpc_pubkey` return the x-only projection of the MPC master key. After the raw-`G` derivation fix (#609) they have no callers, and deriving a child from their output would reintroduce the odd-y "Invalid Schnorr signature" bug — a reviewer asked to drop them. Removes both and the now-unused `SchnorrPublicKey` import.

Stacked on the regression-fix PR.


0xsiddharthks:
Folded into #609, which now carries the whole guardian 2-of-2 stack as one PR (GitHub auto-marked this 'merged' since its commit is now part of the #609 branch). No separate review needed — see #609.

PR #621, opened 2026-05-29
[Guardian] Move all init config to operator

mskd12:
## [Guardian] Move enclave init state into operator_init (IOP-406)

Moves the KP-verifiable session state out of `provisioner_init` and into `operator_init`, with the operator supplying it once at boot. The state's digest (`state_hash`) is the AAD every KP binds on its share submission, so a share only decrypts if the KP agreed on the operator-configured state.

### Highlights

- **`EnclaveMode` { Ceremony, Withdraw }** replaces the `ceremony_mode` bool (now on `EnclaveConfig`). RPC gates and the operator-init completeness check are mode-aware.
- **Per-mode `OperatorInitRequest`** = `{ s3_config, Option<WithdrawModeConfig> }`. Ceremony enclaves (`setup_new_key`/`rotate_kps`) carry **only** S3 config; withdraw enclaves additionally carry the full `WithdrawModeConfig`. The single `Option` encodes the mode — no correlated fields to keep in sync.
- **`WithdrawModeConfig` / `WithdrawModeState` split.** `WithdrawModeState` (committee, limiter config + state, BTC master pubkey) owns `digest()` — the `state_hash`. `WithdrawModeConfig` wraps it plus `secret_sharing_instance` and `network`, which are delivery-only and **excluded from the digest** (the instance is enforced via `verify_share`; the network is share-irrelevant).
- **`operator_init` is validate-then-commit**: all fallible work (S3 check, rate limiter) runs before an all-or-nothing commit, and a single logging-complete sentinel marks completion. Control-plane RPCs serialize under one `Enclave.control_lock`.
- **`rotate_kps` carries `old_instance` in the request** (operator-relayed, no S3 read), so ceremony enclaves need no instance at boot.
- **`CeremonyLogMessage` is now an enum** — `NewKey { instance }` and `Rotate { old_instance, new_instance }`. Rotations record the consumed `old_instance` so the ceremony chain is auditable from the log alone.

Callers updated to match: `bootstrap_guardian`, the KP provisioner (`hashi-monitor`), and the e2e harness.


crates/hashi-types/proto/sui/hashi/v1alpha/guardian_service.proto
dlukegordon: I think this is supposed to be 33 bytes?

crates/hashi-guardian/src/init.rs
dlukegordon, line 157: This write happens after operator_init has already populated the one-shot enclave state, but before `operator_init_logging_complete` is set. If this or the next S3 write fails/panics, `is_operator_init_complete()` stays false while the OnceLocks are already populated, so a retry passes the completion check and then panics trying to set them again. Can we either reject partial init on retry or move fallible logging before the one-shot state mutation?

crates/hashi-types/proto/sui/hashi/v1alpha/guardian_service.proto
mskd12: umm, because of @0xsiddharthks's PR? maybe this was missed in that PR? in that case, happy to fix it here!

dlukegordon: Relevant code is at `crates/hashi-types/src/guardian/proto_conversions.rs:264`

crates/hashi-guardian/src/init.rs
mskd12: S3 writes in init.rs never throw an Err -- they always panic. My idea was to ensure that either a fully initialized enclave exists or force the operator to provision a new enclave. So the state you are worried about should not occur, right?

dlukegordon: SGTM

crates/hashi-types/proto/sui/hashi/v1alpha/guardian_service.proto
mskd12: Looks like this was just a comment fix - the code paths are all already fixed

dlukegordon: Right

dlukegordon (approved):

0xsiddharthks (approved):

PR #624, opened 2026-05-29
Add --serialize-unsigned-transaction to the hashi CLI

Bridgerz:
Generalizes unsigned-transaction generation (previously only on `register --print-only`) into a uniform mode for every transaction-producing command, matching the Sui CLI so the output plugs directly into `sui keytool sign` + `sui client execute-signed-tx` (e.g. multisig).

- New global flags: `--serialize-unsigned-transaction`, `--sender`, `--gas`, `--gas-price` (and `--gas-budget` is now actually wired into the builder).
- Single shared `finalize()` in `sui_tx_executor`: build per mode (serialize-unsigned / dry-run / sign+submit); sender resolved from `--sender` or the keypair; serialize/dry-run need no keypair. The validator executor's `execute()` delegates to it.
- Governance (proposal vote/create/execute), deposit, withdraw, and register all route through it. Deposit/withdraw executor methods are split into pure `build_*` helpers + `execute_*` wrappers, so the e2e API is unchanged.
- `register --print-only` kept as a deprecated alias of `--serialize-unsigned-transaction`.
- In serialize mode, diagnostics go to stderr so stdout carries only the base64 transaction (safe to pipe).
- Batched `deposit request` rejects serialize/dry-run (it spans multiple txs); use `deposit request-single`.

Interop: `to_bcs_base64()` emits the bare BCS `TransactionData` (no intent prefix, standard base64) — byte-identical to what `sui client --serialize-unsigned-transaction` emits, so the external signer applies the intent.

### DIFF
--- a/crates/hashi/src/cli/commands/deposit.rs
+++ b/crates/hashi/src/cli/commands/deposit.rs
@@ -140,12 +140,14 @@ async fn generate_address(config: &CliConfig, recipient: &str) -> Result<()> {
 
 async fn request(
     config: &CliConfig,
-    _tx_opts: &TxOptions,
+    tx_opts: &TxOptions,
     txid: &str,
     vout: u32,
     amount: u64,
     recipient: Option<&str>,
 ) -> Result<()> {
+    use crate::sui_tx_executor::TxMode;
+
     config.validate()?;
 
     let hashi_ids = crate::config::HashiIds {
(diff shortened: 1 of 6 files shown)


crates/hashi/src/cli/commands/deposit.rs
```diff
@@ -140,62 +140,106 @@ async fn generate_address(config: &CliConfig, recipient: &str) -> Result<()> {
...
-        .load_keypair()?
-        .context("Keypair required for deposit request. Set keypair_path in config.")?;
+    // A keypair is optional: serialize/dry-run only need the sender address.
+    let signer = config.load_keypair()?;
+    if tx_opts.mode() == TxMode::Execute && signer.is_none() {
+        anyhow::bail!(
+            "Keypair required to submit a deposit request (set keypair_path in config), \
+             or use --serialize-unsigned-transaction to emit an unsigned transaction."
+        );
+    }
```
bmwill, line 165: Do we expect folks to use the sui cli to submit the signed txn if `--serialize-unsigned-transaction` is used?

bmwill (approved):

PR #625, opened 2026-05-29
[Guardian] provisioner_init: batch shares like rotate_kps

mskd12:
Stacked on `deepak/iop-406-move-init-state-to-operator`.

## Goal
Toward: *KPs send encrypted shares to a relay which waits until it has enough shares before forwarding them to the enclave.* This PR makes `ProvisionerInit` accept a **batch** of shares in one call (the shape the relay will submit), matching the existing `RotateKps` pattern. **The relay itself is out of scope here.**

## Before / after
`ProvisionerInitRequest` carried a single `GuardianEncryptedShare`; each KP called `provisioner_init` separately and the enclave accumulated shares in `scratchpad.shares` (`decrypted_shares()` mutex) across calls until threshold.

Now it carries `Vec<GuardianEncryptedShare>`: the enclave decrypts/verifies every share under its installed `state_hash` AAD, dedups by share id, threshold-checks, and finalizes in **one** call — mirroring `rotate_kps`. The cross-call accumulation state is gone and the call serializes on `control_lock` (like `operator_init`).

The one deliberate difference from `rotate_kps`: the AAD is the enclave's **installed** `state_hash` (from `operator_init`), not a request-carried `state`. So `ProvisionerInitRequest` stays just the share vec — no extra state field.

## Changes
- **proto / domain:** `encrypted_share` → `repeated/Vec encrypted_shares` (regenerated). `build_from_share` now returns one per-KP `GuardianEncryptedShare`; added `new(Vec)` / `encrypted_shares()` / `into_parts()`.
- **`init.rs`:** rewrote `provisioner_init` to the batch+finalize-once flow; tests rewritten to the batch model.
- **cleanup:** removed dead `scratchpad.shares` + `decrypted_shares()`.
- **callers:**
  - `bootstrap_guardian`: builds all T submissions, submits once.
  - `hashi-monitor` provisioner: builds the per-KP share only; submit/prechecks dropped (forwarding becomes the relay's job). `guardian_endpoint` → `relay_endpoint`, with a `TODO: forward to relay`.


crates/hashi-monitor/src/provisioner/mod.rs
```diff
@@ -130,29 +132,32 @@ pub async fn get_guardian_info_from_s3(
         .map_err(|e| anyhow::anyhow!(e))
 }
 
-async fn submit_provisioner_init_to_guardian(
+/// Run the relay-side pre-checks for this KP's share. The relay fronts the
+/// guardian's `GetGuardianInfo`, so we verify it against the expected session
+/// and config before handing over the share.
+///
+/// TODO: once the relay exposes `single_provisioner_init`, send `encrypted_share`
```
dlukegordon, line 139: Will this break devnet if we don't implement this todo?

mskd12: I dont think devnet uses this path as it mocks the provisioner's (see `bootstrap_guardian.rs`). cc @0xsiddharthks for confirmation

dlukegordon (approved):

crates/hashi-monitor/src/provisioner/mod.rs
0xsiddharthks: +1. yeah we're mocking them. devnet should be good.

PR #631, opened 2026-06-01
Add onchain committee handoff certs for reconfig

dlukegordon:


packages/hashi/sources/core/reconfig.move
```diff
@@ -42,26 +42,126 @@ entry fun start_reconfig(
 entry fun end_reconfig(
     self: &mut Hashi,
     mpc_public_key: vector<u8>,
-    cert: CommitteeSignature,
+    mpc_cert: CommitteeSignature,
+    guardian_handoff_cert: CommitteeSignature,
     ctx: &TxContext,
```
bmwill, line 53: I think we may want to actually break this up into a pre-step that can run in parallel with the mpc key resharing.

We can have the old committee submit their handoff cert (i would probably just drop the "guardian" from the name) and then store that in `pending_epoch_change` (we can expand that from being just an Option<u64>) and then on end_reconfig we require that to be present (when not doing genesis/dkg) at which point we can store the sig

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -33,11 +34,17 @@ public(package) fun create(ctx: &mut TxContext): CommitteeSet {
         members: sui::bag::new(ctx),
         epoch: 0,
         committees: sui::bag::new(ctx),
+        guardian_handoffs: sui::bag::new(ctx),
```
bmwill: idk if we need to have a whole new bag, we could always just store it the `committees` bag using a Handoff { epoch } key.

bmwill:
@Bridgerz can you also take a look at this PR.

zhouwfang:
Is this PR ready for review or still a draft?

dlukegordon:
> Is this PR ready for review or still a draft?

@zhouwfang It is a draft since I don't think it's ready to merge yet, but I'd like people's opinion on the general approach here.

zhouwfang:
> > Is this PR ready for review or still a draft?
> 
> @zhouwfang It is a draft since I don't think it's ready to merge yet, but I'd like people's opinion on the general approach here.

I see. My previous understanding of a draft was "work in progress" and not "ready for review" according to github definition:
Screenshot 2026-06-17 at 12 03 33 PM

Happy to take a look.

dlukegordon:
> > > Is this PR ready for review or still a draft?
> > 
> > 
> > @zhouwfang It is a draft since I don't think it's ready to merge yet, but I'd like people's opinion on the general approach here.
> 
> I see. My previous understanding of a draft was "work in progress" and not "ready for review" according github definition: Screenshot 2026-06-17 at 12 03 33 PM
> 
> Happy to take a look.

Thank you, yes that is the normal definition and I guess I am being confusing.

packages/hashi/sources/core/reconfig.move
```diff
@@ -42,26 +42,126 @@ entry fun start_reconfig(
...
     assert!(self.committee_set().is_reconfiguring(), ENotReconfiguring);
+    let from_epoch = self.committee_set().epoch();
     let next_epoch = self.committee_set().pending_epoch_change().destroy_some();
     let next_committee = self.committee_set().get_committee(next_epoch);
+    let new_committee = *next_committee;
     let message = ReconfigCompletionMessage { epoch: next_epoch, mpc_public_key };
-    self.verify_with_committee(next_committee, message, cert);
+    self.verify_with_committee(next_committee, message, mpc_cert);
+
+    self.verify_with_committee(self.current_committee(), new_committee, guardian_handoff_cert);
```
zhouwfang: Genesis aborts here — `current_committee()` is `committees[epoch()]`, which is empty at epoch 0 during DKG. This verify need to be skipped when there's no prior committee. The parallel pre-step handles this; flagging it as a concrete blocker regardless of the restructure.

crates/hashi/src/onchain/watcher.rs
```diff
@@ -640,6 +640,33 @@ async fn handle_events(
...
+                let pending_committee = state
+                    .hashi
+                    .committees
+                    .committees()
+                    .get(&end_reconfig_event.epoch)
+                    .cloned();
+                if let Some(new_committee) = pending_committee {
+                    let handoff = super::convert_move_guardian_handoff(
+                        hashi_types::move_types::GuardianCommitteeHandoff {
+                            new_committee: (&new_committee).into(),
```
zhouwfang: This now rebuilds `new_committee` from the bag and re-serializes. If the round-trip isn't byte-exact with the signed bytes, the cached cert won't verify. What about adding new_committee to the event, or relying on scrape?

zhouwfang:
Have you investigated the CI test failure?

dlukegordon:
> Have you investigated the CI test failure?

Yes there have been a few issues, some not related to this PR, I'm working on it

The current issue is

```
2026-06-23T18:25:29.4792213Z     test tests::test_new_member_joins_key_rotation_after_rotation ... FAILED
2026-06-23T18:25:29.4792219Z 
2026-06-23T18:25:29.4792302Z     failures:
2026-06-23T18:25:29.4792307Z 
2026-06-23T18:25:29.4792390Z     failures:
2026-06-23T18:25:29.4792553Z         tests::test_new_member_joins_key_rotation_after_rotation
2026-06-23T18:25:29.4792557Z 
2026-06-23T18:25:29.4792838Z     test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 41 filtered out; finished in 635.05s
2026-06-23T18:25:29.4792844Z 
2026-06-23T18:25:29.4793113Z   stderr ───
2026-06-23T18:25:29.4793117Z 
2026-06-23T18:25:29.4793479Z     thread 'tests::test_new_member_joins_key_rotation_after_rotation' (151999) panicked at crates/e2e-tests/src/lib.rs:647:39:
2026-06-23T18:25:29.4793742Z     Node 0 failed to reach epoch 1: Timed out waiting for Hashi epoch 1
2026-06-23T18:25:29.4793969Z     note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2026-06-23T18:25:29.4793973Z 
2026-06-23T18:25:29.4794062Z error: test run failed
```

crates/hashi/src/mpc/service.rs
```diff
@@ -982,6 +982,59 @@ impl MpcService {
...
+        let requires_committee_handoff = !self
+            .inner
+            .onchain_state()
+            .state()
+            .hashi()
+            .committees
+            .mpc_public_key()
+            .is_empty();
+        if requires_committee_handoff {
+            let committee_handoff = loop {
```
zhouwfang: These collect/submit loops retry forever — a MOVE_ABORT isn't retryable, so the bug shows up as a 635s timeout instead of a clear failure. Bound the retries / fail loud on non-transient aborts.

packages/hashi/sources/core/reconfig.move
```diff
@@ -42,26 +43,179 @@ entry fun start_reconfig(
...
-    self.verify_with_committee(next_committee, message, cert);
+    self.verify_with_committee(next_committee, message, mpc_cert);
+
     self.reset_num_consumed_presigs();
-    let epoch = self.committee_set_mut().end_reconfig(mpc_public_key, ctx);
-    sui::event::emit(EndReconfigEvent { epoch, mpc_public_key });
+    let (epoch, committee_handoff_cert) = self
+        .committee_set_mut()
+        .end_reconfig(mpc_public_key, ctx);
+    if (committee_handoff_cert.is_some()) {
```
zhouwfang: end_reconfig stores the handoff only if present, but the off-chain blocks the whole reconfig on the submit succeeding — so the "parallel pre-step" is really a hard prerequisite. Either require it on-chain (skipping genesis) and truly parallelize, or make the off-chain non-blocking.

zhouwfang (commented):
The CI failure is the on-chain form of my earlier round-trip concern: https://github.com/MystenLabs/hashi/pull/631#discussion_r3430900210

The collected cert's bytes don't match what the contract verifies -- almost certainly the re-serialized new_committee. A Committee BCS round-trip test (sign on-chain bytes, verify on-chain) would pin and guard it.

dlukegordon:
> The CI failure is the on-chain form of my earlier round-trip concern: [#631 (comment)](https://github.com/MystenLabs/hashi/pull/631#discussion_r3430900210)
> 
> The collected cert's bytes don't match what the contract verifies -- almost certainly the re-serialized new_committee. A Committee BCS round-trip test (sign on-chain bytes, verify on-chain) would pin and guard it.

Yes, I'm already testing a fix locally

crates/hashi/src/mpc/service.rs
```diff
@@ -982,6 +990,73 @@ impl MpcService {
...
+                            warn!(
+                                "submit_committee_handoff submission for epoch {} failed: {e}, retrying...",
+                                epoch
+                            );
+                            self.sleep_if_still_pending(epoch).await;
+                        }
+                    },
+                }
+            }
+        }
```
zhouwfang: Please pull this handoff block into a function.

crates/hashi/src/mpc/service.rs
```diff
@@ -1090,3 +1185,36 @@ impl MpcService {
...
+    EndReconfigAlreadyCompleted,
+}
+
+impl ReconfigSubmissionErrorKind {
+    fn is_non_retryable(&self) -> bool {
+        !matches!(self, Self::NonMoveAbort)
+    }
+}
+
+fn classify_reconfig_submission_error(err: &anyhow::Error) -> ReconfigSubmissionErrorKind {
```
zhouwfang, line 1214: Classifying by substring-matching the error's Display is fragile by construction, hence the shadowed param, inlined magic fragments, and or-ing across render variants. Match the structured MoveAbort instead, and drop is_non_retryable() in favor of the exhaustive match the other call sites already use.

crates/hashi/src/sui_tx_executor.rs
```diff
@@ -1010,14 +1010,15 @@ impl SuiTxExecutor {
...
             Function::new(
                 self.hashi_ids.package_id,
                 Identifier::from_static("reconfig"),
                 Identifier::from_static("end_reconfig"),
             ),
-            vec![hashi_arg, mpc_public_key_arg, cert_arg],
+            vec![hashi_arg, mpc_public_key_arg, mpc_cert_arg],
         );
         let response = self.execute(builder).await?;
         if !response.transaction().effects().status().success() {
```
zhouwfang: Together with line 1058 below: `execute_end_reconfig/handoff` already hold the typed `effects().status()`, but stringify it into `bail!` forcing `classify_reconfig_submission_error` to recover it by substring-matching Display. Match on the structured MoveAbort instead; drops the SDK-format fragility.

zhouwfang (changes requested):

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -42,6 +41,20 @@ public(package) fun create(ctx: &mut TxContext): CommitteeSet {
...
+}
+
+public struct CommitteeHandoffKey has copy, drop, store {
+    epoch: u64,
+}
+
+public struct CommitteeHandoff has store {
+    new_committee: Committee,
+    cert: committee::CommitteeSignature,
+}
```
bmwill, line 69: Can we have some docs added here to explain what these are and how they play into reconfig?

Eg does "epoch" in the handoff key refer to the current epoch or the next epoch? i think it refers to the current epoch given some code below but best to make that clear

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -42,6 +41,20 @@ public(package) fun create(ctx: &mut TxContext): CommitteeSet {
...
+    epoch: u64,
+    committee_handoff_cert: Option<committee::CommitteeSignature>,
+}
+
+public struct CommitteeHandoffKey has copy, drop, store {
+    epoch: u64,
+}
+
+public struct CommitteeHandoff has store {
+    new_committee: Committee,
```
bmwill: This results in storing the Committee in two different places onchain, once at (epoch -> committee) and another at (CommitteeHandoffKey { prev epoch} -> CommitteeHandoff). This likely isn't a huge issue but we could also just stick a "next_epoch" u64 in here and then require pulling the Committee from (epoch -> committee).

packages/hashi/sources/core/reconfig.move
```diff
@@ -42,26 +47,182 @@ entry fun start_reconfig(
...
+
+entry fun submit_committee_handoff(
+    self: &mut Hashi,
+    committee_handoff_cert: CommitteeSignature,
+    _ctx: &TxContext,
+) {
+    self.config().assert_version_enabled();
+    assert!(self.committee_set().is_reconfiguring(), ENotReconfiguring);
+    let from_epoch = self.committee_set().epoch();
+    assert!(self.committee_set().has_committee(from_epoch), ENoCurrentCommittee);
```
bmwill: Is this semi redundant with line 94?

packages/hashi/sources/core/reconfig.move
```diff
@@ -42,26 +47,182 @@ entry fun start_reconfig(
...
+            .insert_committee_handoff(
+                from_epoch,
+                new_committee,
+                committee_handoff_cert.destroy_some(),
+            );
+    };
+    sui::event::emit(EndReconfigEvent { from_epoch, epoch, mpc_public_key });
+}
+
+entry fun submit_committee_handoff(
```
bmwill, line 82: We should probably add a check that enforces this can only be called during none-genesis reconfigs in order to match the logic in "end_reconfig", otherwise we could get into a deadlock situation

crates/hashi-types/src/move_types/mod.rs
```diff
@@ -1344,6 +1364,24 @@ impl TryFrom<CommitteeMember> for crate::committee::CommitteeMember {
...
+}
+
+fn bls_public_key_from_uncompressed_g1_bytes(
+    public_key: &[u8],
+) -> Result<crate::committee::BLS12381PublicKey, anyhow::Error> {
+    let public_key = blst::min_pk::PublicKey::deserialize(public_key)
+        .map_err(|e| anyhow::anyhow!("invalid public key {e:?}"))?;
+    crate::committee::BLS12381PublicKey::from_bytes(public_key.to_bytes().as_slice())
+        .map_err(|e| anyhow::anyhow!("invalid public key {e}"))
+}
```
bmwill, line 1385: Are these changes required?

bmwill (approved):

crates/hashi-types/src/move_types/mod.rs
dlukegordon: Yes, the move contract calls were failing until I started using this

crates/hashi/src/mpc/service.rs
```diff
@@ -804,13 +806,24 @@ impl MpcService {
             }
             match self.submit_end_reconfig(target_epoch, &output).await {
```
zhouwfang, line 808: The outer fail-fast never fires. `submit_end_reconfig` bails non-retryable aborts, which drops the typed `TransactionExecutionError`, so the downcast-based classifier sees `NonMoveAbort` and retries forever, undoing the "fail fast on non-retryable reconfig aborts" change. Preserve the source instead.

zhouwfang (changes requested):

crates/hashi/src/mpc/service.rs
dlukegordon: Pushed another commit. Does that work?

zhouwfang: Ready for review now? CI test is still running.

dlukegordon: Yes it's ready for review, I had just rebased on main so CI ran again.

zhouwfang: Why did CI test take 26min?

dlukegordon: Idk, it has been taking this long for other PRs as well

dlukegordon: For example: [https://github.com/MystenLabs/hashi/actions/runs/28123021528/job/83279580070?pr=710](<https://github.com/MystenLabs/hashi/actions/runs/28123021528/job/83279580070?pr=710>)

zhouwfang: `test_large_withdrawal_signature_chunking` hit fixed 600s deadline waiting for 500 deposits to confirm. The root cause is upstream: the embedded Sui network was slow to advance checkpoints

zhouwfang (approved):

PR #632, opened 2026-06-01
Refactor key DB to pubkey-keyed storage with committee-walk GC

zhouwfang:
#163

### DIFF
--- a/crates/hashi/src/db.rs
+++ b/crates/hashi/src/db.rs
@@ -26,17 +26,29 @@ pub struct Database {
     db: fjall::Database,
     // keyspaces
 
-    // Column Family used to store encryption keys.
+    // Column Family used to store encryption private keys, keyed by their own public key.
     //
-    // key: big endian u64 for the epoch the key is used for
-    // value: 32-byte RistrettoScalar
+    // key: `EncryptionPublicKey` bytes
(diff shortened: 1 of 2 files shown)


crates/hashi/src/db.rs
```diff
@@ -26,18 +26,30 @@ pub struct Database {
...
+    //
+    // key: big endian u64 epoch
+    // value: `EncryptionPublicKey` bytes (the `encryption_keys` key for that epoch)
+    encryption_epoch_index: Keyspace,
+
+    // Epoch -> signing pubkey side-index.
+    //
+    // key: big endian u64 epoch
+    // value: `BLS12381PublicKey` bytes (the `signing_keys` key for that epoch)
+    signing_epoch_index: Keyspace,
```
bmwill, line 51: Is this index needed when you can always go from Epoch -> Committee -> Pubkey?

zhouwfang: Not for lookup — that path already goes Epoch → Committee → Pubkey. It's for idempotency of next-epoch key generation: a node generates and registers its key for the next epoch before that epoch's committee exists, so on a retry or restart it must return the same pubkey it already registered — without this local epoch→pubkey record it generates a different key and wedges once the committee captures the original.

bmwill (approved):
Thanks for the clarification!

PR #633, opened 2026-06-01
[leader] reload the deposit worklist when the time-delay timer fires

0xsiddharthks:
Follow-up to #630. The delayed pass re-runs `process_deposit_requests` without reloading the worklist, but a successful Approve drops the deposit from it — so a deposit approved between sparse Bitcoin blocks is gone when the timer fires, and its Confirm stalls until the next block. Reload before processing (for all nodes, like the block arm), restoring the reload from the closed #535.


0xsiddharthks:
Closing since this issue has been fixed by #653

PR #634, opened 2026-06-02
[btc-monitor] broadcast withdrawals via sendrawtransaction RPC

0xsiddharthks:
During some stress tests, I noticed that using `sendrawtransaction` directly on the bitcoin full node is much more efficient as compared to submitting transactions via kyoto.

### DIFF
--- a/crates/hashi/src/btc_monitor/monitor.rs
+++ b/crates/hashi/src/btc_monitor/monitor.rs
@@ -693,28 +693,26 @@ impl Monitor {
         tx: bitcoin::Transaction,
         result_tx: oneshot::Sender<Result<()>>,
     ) {
-        // Use the bitcoind RPC, not kyoto's P2P submit_package (unreliable under load).
+        // Broadcast via the bitcoind RPC, not kyoto's P2P submit_package, which
+        // dropped its response or hung under load.
         let txid = tx.compute_txid();
-        let result = btc_rpc_call(&bitcoind_rpc, {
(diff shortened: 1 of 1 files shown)


bmwill (approved):

crates/hashi/src/btc_monitor/monitor.rs
```diff
@@ -691,46 +690,31 @@ impl Monitor {
...
+            Ok(_) => {
+                info!("Transaction {txid} broadcast via Bitcoin Core RPC");
                 let _ = result_tx.send(Ok(()));
             }
             Err(e) => {
-                error!("Failed to broadcast transaction {txid}: {e}");
-                let _ = result_tx.send(Err(anyhow::anyhow!(e)));
+                // sendrawtransaction is idempotent; treat an already-known tx as success.
+                let msg = e.to_string();
+                if msg.to_lowercase().contains("already") {
```
dlukegordon: Can we be more specific?

bmwill: Is there a specific error code that's returned we can match on?

0xsiddharthks: Yeah, I did some investigation. so the actual bitcoind message is this:
`Err -27: tx already confirmed on-chain`

I confirmed that we get the same error code on our hosted btc node as well. so replaced the check with an err code match.

PR #648, opened 2026-06-03
[Guardian] modularize hashi-types/guardian

mskd12:
## What

Reorg pass on `hashi-types/guardian` — break the two grab-bag files into focused modules, plus fold in the `CommitteeTransition` → `CommitteeTransitionRequest` rename.

### Module splits (pure moves)
- **`signing.rs`** — `GuardianSigned` envelope + `IntentType`/`SigningIntent`, with the struct and its `new`/`verify` impl co-located (impl moved out of `crypto.rs`). Keeps the intent registry auditable in one place.
- **`log/{mod,message,envelope}.rs`** — `message.rs` holds the `LogMessage` family + per-message S3 key naming; `envelope.rs` holds the `LogRecord` wrapper + `object_key`/lock-duration; `mod.rs` holds the S3 layout constants.
- **`bitcoin_utils/{mod,utxo,taproot}.rs`** — `utxo.rs` (UTXO/tx types, wire types, signing) vs `taproot.rs` (2-of-2 descriptor/address/derivation + the test suite); `BTC_LIB` + `DerivationPath` stay in `mod.rs`.

`mod.rs` drops from ~1266 → ~640 lines. Request/response domain types intentionally stay in `mod.rs`.

All public paths are preserved via re-exports (`pub use log::*`, `bitcoin_utils` module name kept, etc.), so consumers compile unchanged apart from the rename.

### Minor consolidations (no behavior change)
- Merged the two `impl WithdrawalLogMessage` blocks into one.
- `object_key` tests relocated into `log/envelope.rs`.
- The one `bitcoin_utils` integration test builds outputs via the public `OutputUTXO::new_external`/`new_internal` constructors (private fields aren't reachable from the test's new module).

### Rename
`CommitteeTransition` → `CommitteeTransitionRequest` across the guardian module and its consumers (`hashi-guardian`, `hashi`), plus a `guardian_service.proto` comment tweak. Proto message types are unchanged.

## Verification
- A normalized + sorted fidelity diff against `main` (with `super::`→bare path requalification cancelled) shows **no production-logic changes** — only the consolidations above.
- `hashi-types` lib tests pass (log + bitcoin_utils suites green).

## Not in this PR
Standardizing `proto_conversions` on `From`/`TryFrom` (and splitting it) — deferred to a follow-up since it ripples through call sites in several crates.

### DIFF
--- a/crates/hashi/src/withdrawals.rs
+++ b/crates/hashi/src/withdrawals.rs
@@ -663,7 +663,7 @@ impl Hashi {
             .map(|(_, c)| c)
             .ok_or_else(|| anyhow!("no on-chain committee epoch after {from_epoch}"))?;
 
-        let transition = hashi_types::guardian::CommitteeTransition {
+        let transition = hashi_types::guardian::CommitteeTransitionRequest {
             new_committee: hashi_types::move_types::Committee::from(new_committee),
         };
 
--- a/crates/hashi/src/withdrawals.rs
+++ b/crates/hashi/src/withdrawals.rs
@@ -626,7 +626,7 @@ impl Hashi {
         })
     }
 
-    // --- Guardian: validate and BLS-sign a `CommitteeTransition` ---
+    // --- Guardian: validate and BLS-sign a `CommitteeTransitionRequest` ---
 
     /// Rebuild the transition from on-chain state and sign with the historical key.
     #[tracing::instrument(level = "info", skip_all, fields(from_epoch))]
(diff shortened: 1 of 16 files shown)


crates/hashi/src/withdrawals.rs
dlukegordon: Should be `CommitteeTransitionRequest`

Same for `crates/hashi/src/grpc/bridge_service.rs:140`

dlukegordon (approved):

PR #650, opened 2026-06-03
[Guardian] move auth-relevant fields into signed GuardianInfo; RotateKpsState typed certs

mskd12:
Two related guardian-types changes, kept small.

## 1. `RotateKpsState`: `Vec<String>` → `Vec<PgpPublicCert>`
- Validate KP certs at the wire boundary (proto `TryFrom`) instead of re-parsing inside `finalize_rotation` (re-parse block deleted).
- Add `Eq`/`Ord`/`PartialOrd` to `PgpPublicCert` keyed on `armored` (can't derive — `openpgp::Cert` isn't `Ord`); `PartialEq` keyed on `armored` too for consistency. Canonical sort/dedup now operate on the typed value.

## 2. Move `limiter_state` / `limiter_config` / `current_committee_epoch` into signed `GuardianInfo`
These were unsigned top-level fields on `GetGuardianInfoResponse`; moving them into the guardian-signed `GuardianInfo` lets consumers verify them.

- New `fetch_verified_guardian_info() -> Result<GuardianInfo>`: the single trusted entry point (fetch → pin signing pubkey to on-chain → verify signature → return verified `GuardianInfo`).
- `verify_guardian_signed_info` now returns the verified payload (`Result<GuardianInfo>`) instead of `bool`.
- Bootstrap and `guardian_limiter_and_state` drop the extract-before-verify dance.
- `reconcile_guardian_committee` now verifies the signature before reading `current_committee_epoch` (previously read it unverified), while still propagating RPC/parse failures as `Err`.

Proto field numbers in `GetGuardianInfoResponse` renumbered (no deployed clients, so no `reserved`).


dlukegordon (commented):
crates/internal-tools/src/fetch_guardian_info.rs:55 still reads info.signed_info.data.enclave_btc_pubkey without verifying signed_info first. This tool is used by deploy flow to print keys for on-chain publication, so a tampered /info response can substitute the BTC pubkey unless callers independently verify the signed envelope. With this branch, the BTC key is explicitly inside the signed payload; this tool should call info.signed_info.verify(&info.signing_pub_key) and read the returned GuardianInfo before printing enclave_btc_pubkey.

mskd12:
AFAIK this flow of posting the guardian signing pubkey on-chain will be eventually removed (cc @0xsiddharthks). At the end, we should have some way of posting enclave's PCRs on-chain and all nodes will use that to verify the attestation.

So I think leaving `crates/internal-tools/src/fetch_guardian_info.rs` for now is probably fine.

dlukegordon (approved):

PR #660, opened 2026-06-04
[Guardian] Scrape committee + commitments; simplify provisioner-init verification

mskd12:
Two related changes to `provisioner-init`, building on the merged `GuardianReader` work (#663). No behavior change to the happy path beyond stricter, clearer checks.

### 1. Scrape committee + share commitments from S3
The flow took committee and share commitments as hand-entered YAML; it now scrapes them from the guardian's signed S3 logs:
- **Secret-sharing instance** ← latest `ceremony/` entry, via `GuardianReader::read_latest_ceremony_instance`.
- **Committee** ← latest non-`failure-` `committee-update/` entry, via `GuardianReader::read_latest_committee`; cross-checked against the enclave's reported `current_committee_epoch`. Falls back to `hashi_committee_genesis` (now optional) only before any committee-update exists.
- Both share a pure, unit-tested `pick_latest_key` (lex-last, skip `failure-`).
- Config drops `share_commitments`/`ShareCommitmentInput`; `hashi_committee` → optional `hashi_committee_genesis`.

### 2. Simplify the verification
- Drop the `GuardianConfig` wrapper; verify the enclave's reported `GuardianInfo` **directly**, field by field (clear per-field errors): S3 bucket, limiter config, limiter state, and the full secret-sharing instance (commitments + N/T, stronger than the old commitments-only check).
- Assert the enclave **isn't already provisioner-initialized** (`enclave_btc_pubkey.is_none()`).
- Keep the `state_hash` digest match as the fail-fast backstop for the fields `GuardianInfo` doesn't report individually (committee *membership* + MPC master `G`).
- Add **`GetGuardianInfoResponse::verify` → trusted `GuardianInfo`**: attestation anchors the signing key, then the signed info is checked under it. Self-contained for any caller (KP init now; hashi nodes later). The relay precheck uses it — which also verifies the attestation, a step it previously skipped.
- Add `GuardianInfo::into_parts` (destructure an operator-initialized info, erroring on absent fields; `enclave_btc_pubkey` stays `Option`, set only at `provisioner_init`).
- `verify_enclave_attestation` now also takes the signing pubkey to anchor (no-op until check C).

### Notes
- The `read_latest_*` and `get_info` are `GuardianReader` methods (shared verified-read path + private key cache).
- Full read+verify against a mock bucket is a follow-up (needs a mock serving signed records + attestations); `pick_latest_key` is unit-tested here.


crates/hashi-guardian/src/s3_reader.rs
```diff
@@ -102,6 +109,57 @@ impl GuardianReader {
...
+            CeremonyLogMessage::Rotate { new_instance, .. } => new_instance,
+        }))
+    }
+
+    /// The latest applied committee from `committee-update/` — the lex-last
+    /// non-`failure-` (i.e. highest-epoch Success) entry, attestation- and
+    /// signature-verified. `None` if no committee update has been logged (e.g. a
+    /// fresh deployment whose committee still only exists in the operator's boot
+    /// config).
+    pub async fn read_latest_committee(&mut self) -> anyhow::Result<Option<Committee>> {
```
dlukegordon, line 143: Do you think it would make sense to validate the hashi committee cert here?

mskd12: Hmm not sure what we gain by doing so. First of all, we only log the committee. But more importantly, we only log to S3 after verifying the cert in enclave.

dlukegordon (approved):

PR #666, opened 2026-06-05
[Guardian] Verify Nitro enclave attestation + pin PCR0

mskd12:
## What

Turns `verify_enclave_attestation` from a no-op stub into real AWS Nitro attestation verification, and threads an expected PCR0 measurement through every consumer that reads guardian S3 state.

## Why

A guardian publishes an attestation document alongside its session signing key. Until now we trusted that key blindly. This change makes any reader (KP provisioner-init, hashi monitor/auditor) cryptographically verify, on first use of each session, that:
- the attestation's COSE signature chains to the AWS Nitro root and is fresh;
- the document commits to the same signing key we're about to trust (`doc.public_key == signing_pub_key`);
- the enclave image matches an operator-pinned measurement (`PCR0`).

## How

- **`verify_enclave_attestation(attestation, signing_pubkey, build_pcrs)`** (hashi-types): parses + verifies via `fastcrypto::nitro_attestation`, binds the signing key, and pins PCR0. `parse_nitro_attestation(.., true, true, true)` keeps PCR0 in `pcr_map` even if zero so the pin can't be bypassed by a missing entry.
- **`BuildPcrs`** — known-good measurement an attestation is pinned to. Construction mandates PCR0. Records only PCR0 because in a StageX reproducible single-binary build it's the only measurement carrying signal. Follow-up (noted in `TODO(check C)`): a `commit -> BuildPcrs` allowlist keyed on `untrusted_git_revision`, to span all the valid images during an upgrade.
- **Threading**: `GuardianReader` / `GuardianSessionKeyCache` / `get_verified_enclave_pubkey` / `GetGuardianInfoResponse::verify` all take `build_pcrs`; provisioner and monitor configs gain `expected_pcr0` (hex).
- **`non-enclave-dev` feature** (hashi-types, mirrored from hashi-guardian): stubs verification to a no-op so off-enclave e2e/dev consumers accept the mock document the guardian emits outside an enclave. Never enabled in prod.
- **fastcrypto** pinned to merged main `ecbf72b` (the AWS Nitro module, fastcrypto #971).

## Notes for reviewers
- Sample-config `expected_pcr0` values are all-zero placeholders; real PCR0s are filled in at provisioning from the reproducible build.
- Real verification only runs in enclave (prod) builds; CI/dev exercise the `non-enclave-dev` / `test` stub path.


crates/hashi-guardian-init/src/config.rs
```diff
@@ -36,6 +37,11 @@ pub struct ProvisionerConfig {
     /// MPC committee `G` (on-chain `CommitteeSet.mpc_public_key`) as hex of `bcs(G)`;
     /// the derivation master (NOT the guardian's own key). Must match operator init.
     pub hashi_btc_master_pubkey_hex: String,
+
+    /// Expected enclave-image measurement: PCR0 as hex, pinned against every
+    /// session's attestation. Required (a value is needed even in non-Nitro dev,
+    /// where verification is a no-op).
+    pub expected_pcr0: String,
```
benr-ml: why not BuildPcrs?

benr-ml (approved):

crates/hashi-guardian-init/src/config.rs
mskd12: fixed!

PR #667, opened 2026-06-05
Incremental, durable withdrawal signing

Bridgerz:
## Summary

Replaces the all-or-nothing withdrawal signing path with **incremental, out-of-order, on-chain-durable** signing.

Previously the leader collected every input's MPC signature + the guardian sigs + one cert and wrote them in a single `sign_withdrawal`, all inside a 60s window — any timeout, leader rotation, or restart discarded everything, and large (many-input) withdrawals could fail to converge.

Now per-input MPC signatures accumulate on the withdrawal in cert-gated **chunks** (`commit_input_signatures`), each durable on-chain. **Both the collect and the on-chain write are chunked** (M = `mpc_signing_chunk_size`, off-chain config): each leader pass collects one chunk's worth of signatures, commits it, and a later tick resumes from durable on-chain state — so a slow/failed input or a leader change costs only the in-flight chunk. The guardian signs once at finalize, which opens the broadcast gate. On a committee reconfiguration only the **unsigned tail** is re-presigned (`reallocate_presigs`); completed signatures are epoch-independent and survive.

> **Stacked on #697** (batch-per-peer partial-signature collection) — that should be reviewed/merged first.

## Move
- New `hashi::mpc_signing` module: `SigningBatch` with per-input `MpcSig = Pending(presig_index) | Signed(bytes)` — out-of-order fill, first-writer-wins, nonce-safe unsigned-tail reallocation. `signed_count` / completeness are **derived** (counted, never stored) so they can't drift from the slots.
- `WithdrawalTransaction` now holds `signing` + `guardian_signatures` (drops `signatures` / `presig_start_index` / `epoch` / the stored `fully_signed` flag); "fully signed" is derived from the slots + guardian sigs.
- New entries: `commit_input_signatures`, `finalize_withdrawal`, `reallocate_presigs` (certless — re-points presig indices only; gated `assert_unpaused` + not-reconfiguring). New `MpcInputSignaturesMessage` cert + `WithdrawalInputsSignedEvent`.

## Rust
- BCS mirror types (`MpcSig` / `SigningBatch` / restructured `WithdrawalTransaction`) + event mirror; `signed_count` / `is_fully_signed` derived.
- New `SignMpcInputSignatures` RPC + per-chunk cert validator that **re-verifies every Schnorr sig** before BLS-signing. `input_indices` added to `SignWithdrawalTransactionRequest` so the leader requests a specific chunk (empty = all unsigned).
- Executor builders for the three new entries.
- Leader: per pass, collect one chunk → BLS cert → `commit_input_signatures` → resume from chain; once all inputs are signed, guardian one-shot → `finalize_withdrawal`. Stale epoch with unsigned inputs → `reallocate_presigs`.
- Chunk collection **unions partials across members** (first valid sig per input wins, each verified against its sighash before union), so it makes forward progress under contention instead of requiring one member to return the whole chunk.
- Committee rate-limit check moved off the per-pass signing path to the finalize cert (`validate_and_sign_withdrawal_tx_signing`), so signing is driven unconditionally and the limit is enforced once.
- Watcher refreshes the batch from chain on progress / realloc events and gates broadcast on the derived fully-signed state.
- MPC signer reads `signing.pending_index(i)` and signs only the requested unsigned indices (skips already-signed → resume).

## Nonce safety
Invariants enforced in `mpc_signing`: each `Pending` index is unique within (batch, epoch); the global allocator is monotonic; on reconfig every `Pending` slot is re-pointed before any signing (stale-epoch indices are never reused); `Signed` slots hold no index. `reallocate` asserts the caller allocated exactly `pending_count` — and that count is derived, so an over/under-count can't slip through and re-point past the reserved block.

## Tests
- Move: `mpc_signing` + withdrawal-queue unit tests.
- Rust: unit coverage for chunk selection, the streaming per-member collector, and the cross-member union/verify merge; `cargo fmt` / `cargo clippy --all-targets` clean.
- e2e: `test_large_withdrawal_signature_chunking` drives a >252-input withdrawal through the chunked collect → commit → broadcast path on localnet.

## Follow-ups (TODOs in code)
- 3-state operator visibility (metrics / CLI) before rollout.
- Guardian-seq durability (pre-existing; widened by the longer multi-checkpoint window).
- Confirm the MPC presig pool is addressed by the same global index as `allocate_presigs` (relied on by `pending_index(i)`).


crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -34,6 +36,19 @@ use tracing::info;
...
+/// it bounds the durability granularity (work lost on a crash) against the
+/// number of PTBs per withdrawal (and thus the per-leader-task wall-clock).
+///
+/// 128 keeps a typical/max-size withdrawal to a handful of commit PTBs that
+/// finish well inside `LEADER_TASK_TIMEOUT`, while still chunking genuinely huge
+/// withdrawals for resumable durability. 128 × 64B sigs = 8 KiB, within Sui's
+/// 16 KiB pure-arg limit.
+// TODO(tuning): revisit with PTB gas numbers at scale; can be decoupled from the
+// MPC batch-per-peer collection size.
+const MPC_SIGNING_CHUNK_SIZE: usize = 128;
```
zhouwfang: Please make this configurable.

packages/hashi/sources/btc/withdraw.move
```diff
@@ -237,52 +253,85 @@ entry fun commit_withdrawal_tx(
...
+/// Reassign fresh presignatures to the still-unsigned inputs of a withdrawal
+/// whose signing batch is from a previous epoch. Only the pending tail is
+/// re-presigned; already-collected signatures are final and epoch-independent.
+///
+/// Intentionally NOT gated on `assert_not_reconfiguring`: this is the
+/// post-reconfiguration recovery step. It carries no committee cert because it
+/// only reassigns nonce material and is bounded to once-per-withdrawal-per-epoch
+/// by the `mpc_signing` stale-epoch guard. By design it carries no committee
+/// cert: it authorizes no signatures, only re-points pending presig indices.
+entry fun reallocate_presigs(hashi: &mut Hashi, withdrawal_id: address) {
```
zhouwfang, line 265: nit: unlike commit/finalize, it isn't `assert_unpaused`. Consider adding it for consistency.

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -281,13 +294,13 @@ impl LeaderService {
 
         let signed = aggregator.finish()?;
 
-        // 5. Submit sign_withdrawal to Sui (writes signatures on-chain).
-        // Broadcast + confirm happens via process_signed_withdrawal_txns on the next tick.
-        let included_checkpoint_seq = Self::submit_sign_withdrawal(
+        // 5. Submit finalize_withdrawal to Sui (attaches guardian sigs, flips the
+        // broadcast gate). Broadcast + confirm happen via
+        // process_signed_withdrawal_txns on a later tick.
+        let included_checkpoint_seq = Self::submit_finalize_withdrawal(
```
zhouwfang, line 440: `inner.record_guardian_finalized()` is at line 239. Signing now spans multiple checkpoints, so the consumed-but-not-finalized window is wider. Leader's local limiter mirror can lag the guardian's advanced `next_seq` over the now-multi-checkpoint window. WDYT?

Bridgerz: Threw a TODO for a follow up PR

packages/hashi/sources/core/mpc_signing.move
```diff
@@ -0,0 +1,218 @@
...
+    signed_count: u64,
+    /// Epoch the `Pending` presignature indices belong to.
+    epoch: u64,
+}
+
+// ======== Constructors ========
+
+/// Create a batch for `num_inputs`, contiguously assigning presignature
+/// indices so that input `i` uses `presig_base + i`.
+public(package) fun new(num_inputs: u64, presig_base: u64, epoch: u64): SigningBatch {
```
bmwill, line 75: I think we probably want to include as an argument an &mut ref to the "consumed presigs" right? that way allocation (here) and reassignment is centralized. If we think that is possible

packages/hashi/sources/btc/withdrawal_queue.move
```diff
@@ -88,17 +99,17 @@ public struct WithdrawalTransaction has key, store {
...
+    /// the MPC signatures; together they form the 2-of-2 taproot witness.
+    /// Written once at `finalize_withdrawal` (the guardian signs in one shot).
     guardian_signatures: Option<vector<vector<u8>>>,
-    /// Global presignature start index assigned at construction time.
-    /// Input `i` uses presig at index `presig_start_index + i`.
-    presig_start_index: u64,
-    epoch: u64,
+    /// Set true exactly once at `finalize_withdrawal`, when every input has both
+    /// an MPC and a guardian signature. This is the single broadcast gate.
+    fully_signed: bool,
```
bmwill: do we need the extra flag or can we just derive this from "signing" and "guardian_signatures"?

packages/hashi/sources/core/mpc_signing.move
bmwill: hmm i guess given how things work this may not be easy?

packages/hashi/sources/btc/withdraw.move
```diff
@@ -237,52 +253,87 @@ entry fun commit_withdrawal_tx(
     hashi.bitcoin_mut().withdrawal_queue_mut().insert_withdrawal_txn(withdrawal_txn);
 }
 
-entry fun allocate_presigs_for_withdrawal_txn(
+/// Reassign fresh presignatures to the still-unsigned inputs of a withdrawal
+/// whose signing batch is from a previous epoch. Only the pending tail is
+/// re-presigned; already-collected signatures are final and epoch-independent.
+///
+/// Paused-gated like commit/finalize, but intentionally NOT gated on
+/// `assert_not_reconfiguring`: this is the post-reconfiguration recovery step.
```
bmwill: i think it probably doesn't make sense to do this if we are reconfiguring so gating on not reconfiguring probably makes sense still?

packages/hashi/sources/core/mpc_signing.move
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/// Out-of-order accumulator for one withdrawal's per-input MPC signatures.
+/// Owned by this module (fields are private); embedded in the BTC
+/// `WithdrawalTransaction`.
+public struct SigningBatch has store {
+    /// One slot per input; same length/order as the withdrawal's inputs.
+    signatures: vector<InputSig>,
+    /// Number of `Signed` slots — a cardinality, NOT an in-order prefix.
+    signed_count: u64,
```
bmwill: Do we need this? i think it would be nicer if we just calculated this by iterating over `signatures` (even if it does "cost" a bit more. its one less thing to fall out of sync

packages/hashi/sources/core/mpc_signing.move
bmwill: yeah perhaps we leave as-is for now and file an issue to look into hardening this after

packages/hashi/sources/core/mpc_signing.move
zhouwfang: I think this is a safety point, not just tidiness. `signed_count` sizes reallocate's presig allocation, and because the `allocated_count == pending_count` check is derived from it, a wrong count slips through and can re-point past the reserved block which leads to nonce reuse.

crates/hashi-types/proto/sui/hashi/v1alpha/bridge_service.proto
```diff
@@ -152,6 +156,23 @@ message SignWithdrawalTransactionPartial {
...
+  repeated uint64 indices = 2;
+
+  // One MPC Schnorr signature per index in `indices` (64 bytes each).
+  // Followers verify each against that input's sighash before BLS-signing.
+  repeated bytes signatures = 3;
+}
+
+message SignMpcInputSignaturesResponse {
+  MemberSignature member_signature = 1;
+}
```
bmwill, line 169: i don't think this is used anywhere?

crates/hashi-types/src/move_types/mod.rs
```diff
@@ -306,6 +306,73 @@ pub struct CommittedRequestInfo {
     pub bitcoin_address: Vec<u8>,
 }
 
+/// Rust version of the Move hashi::mpc_signing::InputSig enum. Variant order
+/// MUST match Move (Pending = 0, Signed = 1) for BCS.
+#[derive(Clone, Debug, PartialEq, serde_derive::Deserialize, serde_derive::Serialize)]
+pub enum InputSig {
```
bmwill: i would probably rename this "MpcSignatureRequest" or something slightly more generic.

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -331,15 +341,156 @@ impl LeaderService {
...
+                crate::metrics::GUARDIAN_LIMITER_CALLSITE_LEADER_PRE_MPC,
+            );
+            if let Err(e) = result {
+                warn!(withdrawal_txn_id = %txn.id, "Leader limiter rejected withdrawal; retry next checkpoint: {e}");
+                return Ok(());
+            }
+            Some(next_seq)
+        } else {
+            None
+        };
```
bmwill: iirc the creation of the withdrawal request should take into account the limiter, so once we have a withdrawal txn created we should just drive the signing of it and only recheck the limiter once all signing is done (eg before we send it to the guardian for signing)

bmwill (commented):
I don't see a change to the `GetPartialSignaturesRequest` RPC for requesting less than all of the partial sigs. do we also need to make a change there?

cc @zhouwfang

crates/hashi-types/proto/sui/hashi/v1alpha/bridge_service.proto
Bridgerz: The handler does return it... or am i misunderstanding

bmwill: ok my eyes are not working your right, ignore me

crates/hashi-types/src/move_types/mod.rs
Bridgerz: ok how about MpcSig

crates/hashi/src/leader/withdrawal_transactions.rs
Bridgerz: good catch, lets kill the per signature batch thrashing and just do it at creation and finalize.

Bridgerz:
> I don't see a change to the `GetPartialSignaturesRequest` RPC for requesting less than all of the partial sigs. do we also need to make a change there?
> 
> cc @zhouwfang

Claude - "partials are fetched one input at a time, and the resume issues those calls only for the unsigned tail. The incremental change is already correct here; no GetPartialSignaturesRequest change is warranted."

zhouwfang:
Is this PR ready for review now? When ready, could you "request review":

Screenshot 2026-06-18 at 11 26 13 AM

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -118,30 +120,44 @@ impl LeaderService {
...
     ) -> anyhow::Result<()> {
-        // If the withdrawal transaction is from a previous epoch, reassign its presig
-        // indices from the new epoch's counter before signing.
-        // TODO: Batch multiple stale-epoch withdrawals into a single PTB.
+        // Stale-epoch presigs: reassign only the unsigned tail to the current
+        // epoch, then retry next checkpoint. Completed signatures are
+        // epoch-independent and are left untouched.
         let current_epoch = inner.onchain_state().epoch();
-        if txn.epoch != current_epoch {
+        if txn.signing.epoch != current_epoch {
```
zhouwfang: The stale-epoch reallocate runs before the `unsigned.is_empty()` check below, so a withdrawal stuck at finalize across an epoch boundary takes a pointless 0-pending reallocate_presigs PTB and an extra tick. Gate it on `!unsigned.is_empty()`.

zhouwfang (changes requested):
No `GetPartialSignaturesRequest` change is needed for the incremental signing itself, but there is a related gap. The PR chunks the on-chain write, not the collect. All unsigned inputs are collected in one pass before the first commit, and that pass runs inside the 60s `LEADER_TASK_TIMEOUT`. For the 700-input case a cold collect is minutes, so it times out before any chunk lands. So we should chunk the collect too.

The RPC change @bmwill is gesturing at seems what I have been calling "batch-per-peer" (one `GetPartialSignatures` per peer covering all inputs instead of per-input). It's orthogonal: it removes the HTTP/2 reset storm and adds latency margin, and is not required for liveness once the collect is chunked. I can work on that.

zhouwfang:
> No `GetPartialSignaturesRequest` change is needed for the incremental signing itself, but there is a related gap. The PR chunks the on-chain write, not the collect. All unsigned inputs are collected in one pass before the first commit, and that pass runs inside the 60s `LEADER_TASK_TIMEOUT`. For the 700-input case a cold collect is minutes, so it times out before any chunk lands. So we should chunk the collect too.
> 
> The RPC change @bmwill is gesturing at seems what I have been calling "batch-per-peer" (one `GetPartialSignatures` per peer covering all inputs instead of per-input). It's orthogonal: it removes the HTTP/2 reset storm and adds latency margin, and is not required for liveness once the collect is chunked. I can work on that.

"batch-per-peer" is done in #697.

Bridgerz:
@zhouwfang We now collects + commits one M-sized chunk per leader pass (M = mpc_signing_chunk_size, off-chain config) and resumes from on-chain state next tick, so a big withdrawal doesn't need to finish a full collect inside the 60s task timeout. leader picks the chunk via a new input_indices on SignWithdrawalTransactionRequest.

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -331,58 +445,177 @@ impl LeaderService {
...
+                })?;
+
+        let committee = inner
+            .onchain_state()
+            .current_committee()
+            .expect("No current committee");
+        let required_weight = certificate_threshold(committee.total_weight());
+
+        let mut last_checkpoint = 0u64;
+        for chunk in sigs_by_index.chunks(chunk_size) {
```
zhouwfang: I think this loop only runs once, no?

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -331,58 +445,177 @@ impl LeaderService {
...
+            .is_err()
+        {
+            warn!(withdrawal_txn_id = %txn.id, "Timeout waiting for watcher to reach the committed chunk checkpoint");
+        }
+        Ok(())
+    }
+
+    /// Requests MPC signatures for the given input indices from committee members
+    /// until one member returns the complete set.
     async fn collect_withdrawal_tx_signatures(
```
zhouwfang, line 629: The chunk collect requires the complete chunk from a single member. It is fragile under contention even though the chain accepts any subset out-of-order.
1) Rebase on #697.
2) Commit the union instead so every tick makes forward progress.

crates/hashi/src/leader/withdrawal_transactions.rs
Bridgerz: yep, leftover from when we collected everything in one pass then chunked the writes. dropped it and we commit the single chunk directly now.

crates/hashi/src/leader/withdrawal_transactions.rs
Bridgerz: good call. now we union partials across members. First valid sig per input wins, and each candidate is verified against its sighash before it goes in so one bad/slow member can't poison the chunk cert. Then we commit whatever's ready each tick, resume covers the rest. (and yep, already rebased on #697.)

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -331,58 +484,233 @@ impl LeaderService {
...
         members: &[CommitteeMember],
-    ) -> Option<Vec<SchnorrSignature>> {
-        let futures: Vec<_> = members
+        unsigned: Vec<u64>,
+    ) -> anyhow::Result<()> {
+        // The rate limiter is no longer checked per signing pass — signing is
+        // driven unconditionally and the committee re-validates the limit once at
+        // the finalize cert (see `validate_and_sign_withdrawal_tx_signing`).
+        // `M` (mpc_signing_chunk_size) is both the collection unit and on-chain
+        // write batch here: collect one chunk, commit it immediately, then let a
```
zhouwfang: nit: worth noting that they are actually separable (collect is window-bound, commit is PTB-bound).

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -411,44 +738,30 @@ impl LeaderService {
...
-            if slot.is_some() {
-                anyhow::bail!(
-                    "Withdrawal tx signature stream from {validator_address} returned duplicate input_index {idx}",
-                );
+            let idx = partial.input_index as u64;
+            match collector
+                .record(idx, partial.signature.as_ref())
+                .map_err(|e| {
+                    anyhow::anyhow!("Withdrawal tx signature stream from {validator_address} {e}")
+                })? {
```
zhouwfang: nit: one malformed-length or duplicate sig from a member errors its whole stream here (the ? on record), dropping that member's good sigs too — coarser than merge_into_union's per-candidate drop.

zhouwfang (approved):

bmwill (approved):

zhouwfang:
#613

PR #668, opened 2026-06-07
[guardian] persist the ephemeral keypair across restarts

0xsiddharthks:
Companion infra PR: MystenLabs/sui-operations#8132

## Changes

- In `non-enclave-dev` builds, when `GUARDIAN_EPH_KEY_PATH` is set, the guardian loads its keypair from that file, or generates and persists it
- Adds the `GuardianEncKeyPair` byte (de)serializers (serde with manual byte extraction)


crates/hashi-guardian/src/eph_keys.rs
```diff
@@ -0,0 +1,163 @@
...
+        path: &Path,
+        signing: &GuardianSignKeyPair,
+        encryption: &GuardianEncKeyPair,
+    ) -> Result<()> {
+        let keys = PersistedEphKeys {
+            signing_sk: hex::encode(signing.to_bytes()),
+            enc_sk: hex::encode(encryption.to_secret_bytes()),
+            enc_pk: hex::encode(encryption.public_key_bytes()),
+        };
+        if let Some(parent) = path.parent() {
```
dlukegordon: Fyi I think for a relative path, `path.parent()` would return `Some("")`, so we'd be doing a needless mkdir here.

crates/hashi-guardian/src/main.rs
```diff
@@ -50,8 +48,7 @@ async fn main() -> Result<()> {
     }
 
     let mut rng = rand::thread_rng();
-    let signing_keys = GuardianSignKeyPair::new(&mut rng);
-    let encryption_keys = GuardianEncKeyPair::random(&mut rng);
+    let (signing_keys, encryption_keys) = hashi_guardian::eph_keys::load_or_create(&mut rng)?;
```
dlukegordon, line 51: I think we could now get collisions on S3 object keys that use `session_id`, since `session_id` is derived from the signing key and the signing key is now persisted across restarts.

For init logs, the keys are deterministic per session id (`init/{session_id}-...`), so rerunning `operator_init` after a restart can collide regardless of time. For heartbeat logs, the keys can collide on same-hour restarts because the heartbeat counter starts back at 0 and the path is only hour-granular.

0xsiddharthks:
Just sync'd with @mskd12 and we're changing our logic for this to be more like how we want on mainnet.
and instead we're gonna be storing the PCR0 in the pvc storage and then using that for verification instead of ephemeral key.

0xsiddharthks:
Superseding this. Rather than persisting the ephemeral keypair to a PVC, we're following the mainnet model: the committee node verifies the live ephemeral key via Nitro attestation, so a fresh key per boot is fine and the ephemeral key is never pinned on-chain. The on-chain trust anchor moves to the enclave build (PCR0) in #690, with node-side attestation verification as a stacked follow-up (on #666/#675). Closing in favor of that approach.

PR #670, opened 2026-06-07
[kyoto-hardening][btc-monitor] Resume Kyoto near the tip on restart

0xsiddharthks:
## Summary

Stacked on #669. Even anchored at `start_height`, every supervision restart re-downloads everything since deployment. Once we've synced we know the tip, so on rebuild we re-anchor `KYOTO_RESTART_CHECKPOINT_DEPTH` (2016) blocks below it instead — a restart now resyncs ~2k blocks (seconds) rather than every block since `start_height`.

2016 is far above `bitcoin_confirmation_threshold` (6 on devnet) and any plausible reorg, so deposits still awaiting confirmation are always re-scanned; deeper deposits are already final. Falls back to the start checkpoint before the first sync, when the tip is within one depth of the start, or if the hash can't be fetched.

Second of three btc-monitor efficiency PRs.

## Changes

- `resume_checkpoint`: on rebuild, anchor 2016 blocks below the last known tip (hash from bitcoind), falling back to the start checkpoint.
- `resume_anchor_height`: pure helper for the depth-below-tip policy, unit-tested.

### DIFF
--- a/crates/hashi/src/btc_monitor/monitor.rs
+++ b/crates/hashi/src/btc_monitor/monitor.rs
@@ -36,6 +36,12 @@ const KYOTO_RESTART_DELAY_BASE: Duration = Duration::from_secs(5);
 /// Random additional delay to spread reconnects across pods.
 const KYOTO_MAX_RESTART_DELAY_JITTER: Duration = Duration::from_secs(30);
 
+/// How far below the last known tip to re-anchor Kyoto when rebuilding after a
+/// restart. Must stay well above `bitcoin_confirmation_threshold` and any
+/// plausible reorg so deposits still awaiting confirmation are re-scanned;
+/// 2016 blocks (one difficulty period) clears both comfortably.
+const KYOTO_RESTART_CHECKPOINT_DEPTH: u32 = 2016;
+
 /// How many Bitcoin blocks a deposit observation can go without being
 /// refreshed before it's dropped from the confirmation-metrics cache.
 const STALE_OBSERVATION_BLOCKS: u32 = 10;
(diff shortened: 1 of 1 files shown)


0xsiddharthks:
Holding this off after a closer look at correctness.

Re-anchoring at `tip − KYOTO_RESTART_CHECKPOINT_DEPTH` (2016) drops kyoto's headers below that height. But nothing bounds how far below the tip a deposit's confirming block can sit: a user can fund the BTC tx and submit the Sui `DepositRequest` thousands of blocks later, the Move side only expires the *request* by Sui-time (not Bitcoin age), and the leader retries confirmation indefinitely (`max_retries = u32::MAX`). After a restart, a deposit whose block is more than 2016 below the tip fails kyoto's `height_of_hash` cross-check (`monitor.rs`) permanently — and each later restart pushes the anchor higher, so it never self-heals. Recoverable (no fund risk, the UTXO stays unspent), but a real liveness regression.

The premise this rested on — "deposits awaiting confirmation are within `threshold` of the tip" — only holds for deposits waiting *because they lack confirmations*, not for freshly-submitted requests against old blocks.

The original problem is already solved without this: #669 removes the genesis replay (anchor at `start_height`), and #671 removes the per-header catch-up storm. #669 is also always-correct precisely because no deposit can predate `start_height` — the invariant this PR breaks by anchoring higher.

Leaving the branch up. If we revisit restart-resync cost, the correct shape is a below-anchor fallback: when a pending deposit's block is below the current anchor, re-anchor at `start_checkpoint` on the next rebuild.

PR #675, opened 2026-06-10
[Guardian] Pin attestations with PCR allowlist

mskd12:
## Summary
- Replace the single expected PCR with a PCR allowlist: `current_build` plus configured `prev_builds`, each represented as `BuildPcrs`.
- Add `BuildPolicy::{Current, AnyAllowlisted}` to single-object S3 reader APIs so live-session reads require the current build while historical reads can explicitly accept any allowlisted build.
- Thread resolved `BuildPcrs` into Nitro attestation verification so PCR0 is checked against the allowlist entry for the signed guardian revision.
- Carry verified PCR pins as `build_pcrs` on verified session/log records; batch log reads still expose per-record `build_pcrs` so callers can apply upgrade-boundary policy with local context.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3007,7 +3007,6 @@ dependencies = [
  "e2e-tests",
  "hashi-guardian",
  "hashi-types",
- "hex",
  "jsonrpc",
  "serde",
  "serde_yaml",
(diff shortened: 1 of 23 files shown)


dlukegordon (approved):
In `crates/internal-tools/src/fetch_guardian_info.rs`, we don't verify `signing_pub_key`, but a comment says:

```
//! Calls `GuardianService/GetGuardianInfo` and prints a hex-encoded key
//! to stdout. Used by the deploy workflow to feed `hashi publish
//! --guardian-public-key` / `--guardian-btc-public-key` so the keys get
//! recorded on chain.
```

I think we should either
- Add verification to `fetch_guardian_info.rs`
- Remove that command
- At least update the comment with a TODO that this needs to be removed, like when we stop using the bootstrap process

0xsiddharthks (approved):

PR #683, opened 2026-06-16
Authorize cert submission and governance by the operator key

Bridgerz:
## Summary

Authorizes MPC certificate submission and governance by committee **member** rather than by the raw transaction sender, so a validator can run its node with a delegated **operator key** while keeping the validator (account) key in cold storage.

## Background

On-chain, cert submission and governance required the transaction *sender* to be the member's **validator address** — but the node signs with its **operator key**. The intended model (validator key registers, then cold-stored as backup; operator key runs the node) was unusable; it only worked because devnet/e2e reuse one key for both.

## Changes

- `committee_set::member_authorized(validator, sender)` — true when `sender` is the validator's own key **or** its registered `operator_address`. The validator key always retains authority (backup).
- **Cert submission** authorizes the `dealer` via `member_authorized` (the cert already carries the dealer/validator address).
- **Governance** (`proposal::create`/`vote`/`remove_vote` + the propose wrappers) take the member's validator address as an explicit argument, authorize via `member_authorized`, and record the vote under the validator so quorum stake-weight is correct. The CLI resolves that address off-chain from the committee, so operators don't pass it manually.
- `set_operator_address` is **validator-key-only** — a compromised operator key can't re-delegate; the validator key is the rotation/recovery path.

## Not included

An earlier `HashiOperationCap` approach was dropped: as an owned object the node version-bumps it on every cert/metadata tx, so a concurrent governance tx signed by the same key fails with an object-version conflict (reproduced in e2e). The address-based scheme has no owned object and no contention.

## Testing

- Move unit suite passes, incl. new `operator_delegation_tests`.
- e2e (localnet): DKG + recovery, governance create→vote→execute quorum cycle, key rotation + restart recovery — all pass with a distinct operator key.

### DIFF
--- a/packages/hashi/sources/core/cert_submission.move
+++ b/packages/hashi/sources/core/cert_submission.move
@@ -83,7 +83,7 @@ fun submit_cert_internal(
     // The dealer's own validator key, or the operator key it has delegated to,
     // may submit the dealer's certificate. `member_authorized` also enforces
     // that the dealer is a registered committee member.
-    assert!(hashi.committee_set().member_authorized(dealer, ctx.sender()));
+    assert!(hashi.committee_set().member_authorized(dealer, ctx));
     let pending = hashi.committee_set().pending_epoch_change();
     assert!(epoch == hashi.committee_set().epoch() || pending.contains(&epoch));
     let epoch_certs = hashi.epoch_certs(key, protocol_type, ctx);
--- a/packages/hashi/sources/core/cert_submission.move
+++ b/packages/hashi/sources/core/cert_submission.move
@@ -80,8 +80,10 @@ fun submit_cert_internal(
     ctx: &mut TxContext,
 ) {
     hashi.config().assert_version_enabled();
-    assert!(ctx.sender() == dealer);
-    assert!(hashi.committee_set().has_member(dealer));
+    // The dealer's own validator key, or the operator key it has delegated to,
+    // may submit the dealer's certificate. `member_authorized` also enforces
+    // that the dealer is a registered committee member.
+    assert!(hashi.committee_set().member_authorized(dealer, ctx.sender()));
     let pending = hashi.committee_set().pending_epoch_change();
     assert!(epoch == hashi.committee_set().epoch() || pending.contains(&epoch));
     let epoch_certs = hashi.epoch_certs(key, protocol_type, ctx);
--- a/packages/hashi/sources/core/committee/committee_set.move
+++ b/packages/hashi/sources/core/committee/committee_set.move
@@ -17,10 +17,6 @@ use sui::{
 // Errors
 //
 
-#[error(code = 0)]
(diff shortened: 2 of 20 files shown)


packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -202,15 +229,20 @@ public(package) fun set_next_epoch_encryption_public_key(
...
+/// address; the validator key always remains in control of the delegation.
 public(package) fun set_operator_address(
     self: &mut CommitteeSet,
     validator_address: address,
     operator_address: address,
     ctx: &TxContext,
 ) {
     let member = self.member_mut(validator_address);
-    member.assert_update_permitted(ctx);
+    assert!(ctx.sender() == member.validator_address, ENotValidator);
```
bmwill: Is this actually an issue? if an attacker is able to rotate the operator address, the validator address can still go and rotate the operator address to whatever it wants so its not like the validator is ever locked out of changing this.

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -46,6 +54,25 @@ public(package) fun has_member(self: &CommitteeSet, validator_address: address):
...
+    self: &CommitteeSet,
+    validator_address: address,
+    sender: address,
+): bool {
+    if (!self.has_member(validator_address)) {
+        return false
+    };
+    let member = self.member(validator_address);
+    sender == member.validator_address || sender == member.operator_address
+}
```
bmwill: This looks like a very similar function to `assert_update_permitted`, but without having already loaded member info. Can we actually go and change this to not have `sender` passed in but instead have ctx passed in instead? and can we maybe look at reusing some of the logic in `assert_update_permitted` (maybe need need to reword the name of that function) so that the gating logic ends up checking the same things and its consolidated in one place.

packages/hashi/sources/core/cert_submission.move
```diff
@@ -80,8 +80,10 @@ fun submit_cert_internal(
     ctx: &mut TxContext,
 ) {
     hashi.config().assert_version_enabled();
-    assert!(ctx.sender() == dealer);
-    assert!(hashi.committee_set().has_member(dealer));
+    // The dealer's own validator key, or the operator key it has delegated to,
+    // may submit the dealer's certificate. `member_authorized` also enforces
+    // that the dealer is a registered committee member.
+    assert!(hashi.committee_set().member_authorized(dealer, ctx.sender()));
```
bmwill: ah this is what you meant, nice catch

packages/hashi/sources/core/committee/committee_set.move
Bridgerz: yeah good point. I don't see an issue with leaving it how we had it.

Bridgerz:
This is ready for another round @bmwill

bmwill (approved):

PR #681, opened 2026-06-15
Add guardian ceremony run/verify commands

dlukegordon:

PR #685, opened 2026-06-17
[Guardian] Persist encrypted KP shares to S3 for recovery

mskd12:
Groundwork for the ceremony `run`/`verify` rework in #681: make the **enclave** persist each ceremony's encrypted shares, so a KP who loses their share can recover it — instead of the shares living only in the ephemeral `get_guardian_info` response.

## What
- **New `LogMessage::Shares` / `SharesLogMessage`** written to a `shares/{sharing_seq:020}-{session_id}.json` prefix, keyed by `sharing_seq` so each object pairs with its `ceremony/` instance. Short object lock — a fetch-window guarantee, *not* integrity (that's the enclave signature) — so shares stay readable until purged.
- **Enclave writes shares in both `setup_new_key` and `rotate_kps`**, so recovery is rotation-correct by construction (no genesis-only gap).
- **Roster committed into the attested `ceremony/` log** (`CeremonyLogMessage` now carries the per-share recipient fingerprints), so a KP can check the full recipient set against the immutable record rather than the operator-served shares. See the roster gap discussion on #681.
- **`GuardianReader`**: `read_latest_ceremony` (also returns the writing session id, needed to locate the share object) + `read_shares` point-read.

## Why this shape
Shares are enclave-authored state, so the enclave writes them (signature carries authenticity; the operator can't forge). Reusing the existing locked log path means **no second bucket, no lock-agnostic primitives, no operator-side upload** — which supersedes the artifact-bucket approach in #681. The short lock + separate prefix lets shares be purged without tripping the strict immutable-log reader that guards `ceremony/`.

## Scope / follow-ups
- KP-side recovery **tooling** (a `ceremony` command that reads + decrypts from `shares/`) is deferred; this PR lands the enclave write path + reader primitives.
- Real attestation verification (`verify_enclave_attestation` is a no-op) is tracked separately.

Draft until the above lands / #681 is reworked on top.

### DIFF
--- a/crates/hashi-guardian/src/ceremony_mode/setup.rs
+++ b/crates/hashi-guardian/src/ceremony_mode/setup.rs
@@ -58,6 +58,8 @@ pub async fn setup_new_key(
     let ss_instance = SecretSharingInstance::new(share_commitments.clone(), n, t, 0)
         .expect("(n, t) validated by SetupNewKeyRequest; commitments produced with matching count");
 
+    enclave.log_shares(0, encrypted_shares.clone()).await?;
+
     enclave
         .log_ceremony(CeremonyLogMessage::NewKey {
             instance: ss_instance,
@@ -65,9 +67,9 @@ pub async fn setup_new_key(
         })
         .await?;
 
-    enclave.log_shares(0, encrypted_shares.clone()).await?;
-
-    enclave.set_latest_encrypted_shares(encrypted_shares.clone())?;
+    enclave
+        .set_latest_encrypted_shares(encrypted_shares.clone())
+        .expect("set_latest_encrypted_shares should work if ceremony_complete=false");
 
     let response = enclave.sign(SetupNewKeyResponse {
         encrypted_shares,
--- a/crates/hashi-guardian/src/ceremony_mode/setup.rs
+++ b/crates/hashi-guardian/src/ceremony_mode/setup.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::Enclave;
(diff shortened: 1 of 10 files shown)


crates/hashi-guardian/src/s3_reader.rs
```diff
@@ -126,13 +139,37 @@ impl GuardianReader {
...
+
+    /// Read + verify the encrypted shares at `shares/{seq}-{session}.json`. Point
+    /// read by exact key (recovery anchors on the ceremony instance's seq), so a
+    /// purged older `shares/` object never blocks reading the current one.
+    ///
+    /// Uses the lock-agnostic read: shares carry only a short lock that is
+    /// expected to expire, and their integrity is the enclave signature checked
+    /// below — not S3 immutability — so the immutable-log lock assertion in
+    /// `get_log_record` doesn't apply.
+    pub async fn read_shares(
```
dlukegordon, line 161: Should we sanity check that `record.session_id == session_id` and `msg.sharing_seq == sharing_seq`?

crates/hashi-guardian/src/ceremony_mode/setup.rs
```diff
@@ -60,9 +61,12 @@ pub async fn setup_new_key(
     enclave
         .log_ceremony(CeremonyLogMessage::NewKey {
             instance: ss_instance,
+            roster: recipient_roster(&encrypted_shares),
         })
         .await?;
 
+    enclave.log_shares(0, encrypted_shares.clone()).await?;
```
dlukegordon: Nit: maybe we should call log_shares before log_ceremony, since the ceremony points to shares, but not vice versa?

dlukegordon (approved):

PR #687, opened 2026-06-17
[Guardian] Route ceremony reads through one GuardianReader

mskd12:
Follow-up to #681 (now merged). `ceremony run` and `verify` each built a standalone `GuardianS3Client` alongside a `GuardianReader` and re-implemented the `ceremony/`+`shares/` reads — which re-resolved the session attestation *outside* the key cache (where `check C` will land) and double-read the ceremony record. This routes both commands through a single `GuardianReader`.

## Reader
- `verified_pubkey(session)` — cache-backed accessor, so a cross-check shares the cache instead of a second attestation pass.
- `read_latest_ceremony` now also returns the **roster**; a private `read_ceremony_record(key)` backs the read.
- Both commands drop the standalone client and go through one `GuardianReader`: `run` cross-checks the attested signing key via `verified_pubkey`, then reads the logged state via `read_latest_ceremony`; `verify` reuses the same read. **One reader, one cache: attestation resolved once, no redundant ceremony read.**

## Verified-read surface tightened
- Collapse `get_verified_log_record` — its signed branch had no callers after `read_session_ceremony` was removed; the unsigned attestation read is inlined into `get_verified_enclave_pubkey`.
- `pub(crate)` on `get_log_record` and `get_verified_enclave_pubkey`, so off-crate code can only do verified reads / resolve pubkeys through `GuardianReader`'s cache (the future `check C` chokepoint).

## Consolidated ceremony-state validation
- A single `VerifiedCeremonyState` carries `(session_id, encrypted_shares, secret_sharing_instance)`, built either from the live `SetupNewKeyResponse` or from the `ceremony/`+`shares/` logs, and self-validates (n/t/sharing_seq, share count, recipient roster). `run` compares the live and logged states with one `==`.
- `KPEncryptedShares` newtype flows through `SetupNewKeyResponse`, `RotateKpsResponse`, `latest_encrypted_shares`, `GetGuardianInfoResponse`, and the shares log instead of bare `Vec<KPEncryptedShare>`. It validates share ids are exactly `1..=n` at construction (and `SecretSharingInstance::new` validates commitment ids likewise), replacing the old standalone `verify_response_shape`.
- The proto `SetupNewKeyResponseData` carries the full `SecretSharingInstance` (n/t/seq + commitments) rather than just `share_commitments`.
- `ensure_roster_matches` compares the `ceremony/` roster and `shares/` recipient fingerprints as sorted multisets, so it stays correct regardless of `shares/` storage order.

## Key helpers
- Named `SharesLogMessage::object_key` (in the `InitLogMessage::attestation_object_key` style), backed by one internal layout definition, used at the read site instead of hand-formatting keys.

No behavior change to the ceremony flow itself.


crates/hashi-guardian-init/src/ceremony.rs
```diff
@@ -437,276 +419,198 @@ fn load_kp_certs(paths: &[PathBuf]) -> Result<Vec<PgpPublicCert>> {
...
-    );
-    ensure!(
-        instance.sharing_seq() == sharing_seq,
-        "ceremony/ log sharing_seq ({}) differs from expected ({})",
-        instance.sharing_seq(),
-        sharing_seq
-    );
+    /// Confirm the agreed `ceremony/` roster matches the recipient fingerprints
+    /// on the `shares/` ciphertexts.
+    fn ensure_roster_matches(&self, roster: &[KPFingerprint]) -> Result<()> {
```
dlukegordon, line 506: This now would mean roster [B, A] would pass against [A, B], but the order of the fingerprints matter, an out of order fingerprint list would mean different shares are assigned to different pubkeys.

mskd12: Yeah, fixing it.

mskd12: i introduced a sort() inside KPEncryptedShares's constructor so entries are always stored in the same form. please check!

dlukegordon (approved):

PR #690, opened 2026-06-18
[guardian] remove the onchain ed25519 pin

0xsiddharthks:
## Summary

The node used to pin the guardian's ephemeral ed25519 signing key on-chain (`guardian_public_key`). The guardian regenerates that key on every enclave reboot, so a reboot broke verification and stalled withdrawals.

Per design review, validators don't pin the guardian's signing key or its enclave build (PCR0) — the enclave is a KP/provisioner concern, and the only crypto material a validator needs to trust is the immutable BTC key. So drop `guardian_public_key` from on-chain config entirely: the guardian's on-chain identity is now its URL plus the immutable BTC key. The node reaches it over TLS and verifies withdrawals against the BTC key (node side: #691).

Replaces the earlier "pin the enclave build (PCR0) on-chain" approach.

## Changes

- `config.move`: drop `guardian_public_key`; `set_guardian` → `set_guardian_url`.
- `update_guardian` proposal + `finish_publish`: URL only.
- CLI / publish: drop `--guardian-public-key`.

### DIFF
--- a/crates/e2e-tests/src/lib.rs
+++ b/crates/e2e-tests/src/lib.rs
@@ -253,11 +253,6 @@ impl TestNetworksBuilder {
         let guardian_btc_pubkey = guardian_harness.ensure_btc_pubkey()?;
         let guardian_config = hashi::publish::GuardianConfig {
             url: guardian_harness.endpoint().to_string(),
-            public_key: guardian_harness
-                .enclave()
-                .signing_pubkey()
-                .as_bytes()
-                .to_vec(),
             btc_public_key: guardian_btc_pubkey.serialize().to_vec(),
         };
         tracing::info!(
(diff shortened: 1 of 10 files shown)


dlukegordon (approved):

bmwill:
I would rather treat the ed25519 key as not ephemeral and instead not pin the build hash onchain. Because this would require a gov action to update every single time we do a software release which we don't want to do

bmwill (changes requested):
See comment

mskd12:
> I would rather treat the ed25519 key as not ephemeral and instead not pin the build hash onchain. Because this would require a gov action to update every single time we do a software release which we don't want to do

If the idea is to be able to move fast in devnet, why not simply skip attestation checks? We can toggle this back pretty easily when we want to.

Alternatively, are you suggesting that we should not pin the build hash onchain even in mainnet?

(cc @bmwill)

bmwill:
Yes i'm saying even in mainnet we shouldn't pin this in mainnet, the fact that the guardian is running in an enclave is irrelevant to the validators and really is only relevant for the provisioners.

bmwill (approved):

PR #692, opened 2026-06-18
[guardian] dev tool: recover a restarted guardian's BTC key from S3 shares

0xsiddharthks:
Adds `hashi-guardian-init tools dev-recover`, the recovery counterpart to `dev-bootstrap`. Under the ceremony model the master secret is generated inside the enclave and Shamir-split to the KP PGP keys, with encrypted shares persisted to S3 (#685) and the `SecretSharingInstance` in the ceremony log (#687). A restart loses the in-memory BTC key and nothing currently rebuilds it from S3 — this does.

- **recovery mode**: reads the ceremony's `SecretSharingInstance` + encrypted KP shares from S3, decrypts `t` of them with a local gpg homedir (cross-checking each against the ceremony commitments), then drives `OperatorInit -> GetGuardianInfo -> ProvisionerInit` reusing the ceremony's instance and a *recovered* limiter state (so a restart doesn't reset the rate-limit bucket). Asserts the post-init enclave BTC pubkey matches the on-chain `guardian_btc_public_key`.
- **`--print-master-pubkey`**: decrypts `t` shares and combines them locally to print the x-only master pubkey; no chain/RPC/guardian contact. Used at genesis to capture the pubkey to pin on-chain, since a ceremony-mode guardian never exposes it directly.

Reuses the ceremony gpg-decrypt helper, the S3 `GuardianReader`, and `limiter_recovery`; promotes three `dev_bootstrap` helpers to `pub(crate)`.

Dev-scoped and clearly marked as such: the production recovery path runs each KP's share through the relay (the `provisioner` flow Luke/Deepak own), not one gpg homedir. Not for the prod path.

Decrypted-share plaintext is the 32-byte big-endian secp256k1 scalar (`Share::value.to_bytes()`), share id taken from the `KPEncryptedShare` — same format `ceremony verify` decrypts.

`cargo check`/`clippy -D warnings` clean; `cargo nextest run -p hashi-guardian-init` green (32 tests, incl. 3 new pure share-decrypt/commitment/reconstruct tests that don't need gpg or S3).

### DIFF
--- a/crates/hashi-guardian-init/src/dev_bootstrap.rs
+++ b/crates/hashi-guardian-init/src/dev_bootstrap.rs
@@ -336,13 +336,13 @@ fn parse_master_secret(hex_str: &str) -> Result<k256::SecretKey> {
         .map_err(|e| anyhow!("--master-secret-hex is not a valid secp256k1 scalar: {e}"))
 }
 
-fn required_env(name: &str) -> Result<String> {
+pub(crate) fn required_env(name: &str) -> Result<String> {
     env::var(name).map_err(|_| anyhow!("required env var `{name}` is not set"))
 }
 
 /// Decode the on-chain MPC committee verifying key `G` (`mpc_public_key` = `bcs(G)`),
 /// matching `Hashi::onchain_verifying_key_g` — the point the guardian must derive from.
-fn decode_mpc_master_g(mpc_public_key: &[u8]) -> Result<HashiMasterG> {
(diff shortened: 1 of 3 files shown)


0xsiddharthks:
Closing — superseded. The share-decrypt + commitment checks now live in #702 (kp_roster); recovery is being unified onto the provision path (#700–#704 + `single_provisioner_init`); and the genesis master pubkey will come from the ceremony-mode guardian's signed info rather than reconstructing from shares here. Branch kept in case we revisit.

PR #693, opened 2026-06-18
[guardian] dev-only hashi-guardian-init image for in-cluster self-heal

0xsiddharthks:
Phase-3, dev-only. New `docker/hashi-guardian-init/Containerfile` that builds the `hashi-guardian-init` binary into a debian-slim runtime with `gpg` + `ca-certificates`.

It is the image for an in-cluster guardian self-heal CronJob (sui-operations side) that re-runs `tools dev-recover` after an ad-hoc guardian pod restart *between* deploys — recovery that today only happens on the next deploy. Same musl-static build as the sibling Containerfiles; the runtime needs `gpg` because `dev-recover` shells out to it to decrypt the ceremony KP shares.

Stacked on #692 (which adds `tools dev-recover`); merge after it. The Containerfile compiles (`cargo check` passes); the full multi-stage Docker build is unvalidated here.

### DIFF
--- a/docker/hashi-guardian-init/Containerfile
+++ b/docker/hashi-guardian-init/Containerfile
@@ -1,0 +1,85 @@
+# syntax=docker/dockerfile:1
+
+# Off-enclave guardian init tooling (`hashi-guardian-init`).
+#
+# Phase-3, DEV-ONLY image. Built so an ad-hoc guardian pod restart between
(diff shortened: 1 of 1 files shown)


0xsiddharthks:
Closing with #692 — this image only existed to run `dev-recover` in-cluster, which is superseded by the provision flow.

PR #694, opened 2026-06-18
Update guardian provision command

dlukegordon:


crates/hashi-guardian-init/provision.sample.yaml
```diff
@@ -1,20 +1,43 @@
...
+kp_pgp_cert_paths:
+  - /path/to/kp1.asc
+  - /path/to/kp2.asc
+  - /path/to/kp3.asc
+  - /path/to/kp4.asc
+  - /path/to/kp5.asc
+
+# Expected secret-sharing sequence. Use 0 if provisioning against the genesis
+# key, or N+1 after a rotation to sharing_seq N.
+sharing_seq: 0
```
mskd12: why is the sharing sequence number needed? you are already fetching that from s3, no?

mskd12 (approved):
Just one small comment. LGTM otherwise!

crates/hashi-guardian-init/provision.sample.yaml
dlukegordon: Good point, I removed it.

PR #696, opened 2026-06-18
Add `proposal create emergency-pause` CLI command

Bridgerz:
## What

Adds `hashi proposal create emergency-pause [--unpause]`, wired to the on-chain `emergency_pause::propose(pause)`. Default proposes a **pause**; `--unpause` proposes an **unpause**.

```
hashi proposal create emergency-pause              # propose pause   (5% quorum)
hashi proposal create emergency-pause --unpause    # propose unpause (2/3 quorum)
```

## Why

`emergency_pause` was the **only** on-chain governance proposal type with no CLI `create` path. The CLI already handled EmergencyPause for `vote` / `remove-vote` / `view` / `list` / `execute` (the type is wired into every dispatch in `client.rs`), but a committee member could not *initiate* one. That left the deliberately-low **5% emergency-pause threshold unreachable from the CLI** — the only way to set `paused` was `update-config paused bool:true` at the normal **2/3** threshold, which defeats the purpose of a fast halt.

This was surfaced during a CLI coverage audit of the validator/protocol-management surface; emergency-pause-create was the single gap.

## Changes

- `cli/mod.rs` — `CreateProposalCommands::EmergencyPause { unpause, metadata }` + dispatch (`pause = !unpause`).
- `cli/client.rs` — `CreateProposalParams::EmergencyPause` + `build_create_proposal_transaction` arm calling `emergency_pause::propose`.
- `cli/commands/proposal.rs` — `create_emergency_pause_proposal` handler (mirrors the other `create_*` handlers).

No Move changes — the on-chain function already existed.

## Testing

Verified end-to-end on a 4-validator localnet:

- **Pause:** proposal shows `Threshold: 500 bps (5.00%)`, reaches quorum on the proposer's single vote; after `execute`, a `withdraw request` aborts with `ESystemPaused = System is currently paused`.
- **Unpause:** proposal shows `Threshold: 6667 bps (66.67%)`, requires 3 of 4 validators; after `vote --execute`, withdrawals succeed again.
- Regression pass over the existing `proposal` create types (`--dry-run`), `vote`/`remove-vote`/`execute`, committee/config queries, and a full deposit->balance->withdraw flow — all green.

### DIFF
--- a/crates/hashi/src/cli/commands/proposal.rs
+++ b/crates/hashi/src/cli/commands/proposal.rs
@@ -670,6 +670,41 @@ pub async fn create_update_guardian_proposal(
     Ok(())
 }
 
+/// Create an emergency pause (or, with `pause == false`, unpause) proposal.
+///
+/// Pausing uses a deliberately low quorum (default 5%) so a small fraction of
+/// committee weight can halt the protocol quickly; unpausing requires the
+/// normal 2/3 supermajority. Both paths target `emergency_pause::propose`.
+pub async fn create_emergency_pause_proposal(
+    config: &CliConfig,
+    pause: bool,
+    metadata: Vec<(String, String)>,
+    tx_opts: &TxOptions,
+) -> Result<()> {
+    let action = if pause { "Pause" } else { "Unpause" };
+    let title = format!("Creating Emergency {action} Proposal:");
+    print_detail(&format!("\n{}", title.as_str().bold()));
+    print_detail(&format!("  Action: {action}"));
+    print_metadata(&metadata);
+
+    prompt_continue(
+        &format!("create this emergency {} proposal", action.to_lowercase()),
+        tx_opts,
+    )
+    .await?;
+
+    let mut client = HashiClient::new(config).await?;
+    let tx = client.build_create_proposal_transaction(CreateProposalParams::EmergencyPause {
+        pause,
+        metadata,
+    })?;
+
+    print_info("Transaction: emergency_pause::propose");
+    let response = execute_or_simulate(&mut client, tx, tx_opts).await?;
+    print_created_proposal_id(response.as_ref());
+    Ok(())
+}
+
 // ============ Helper Functions ============
 
 fn print_proposal_detailed(
--- a/crates/hashi/src/cli/commands/proposal.rs
+++ b/crates/hashi/src/cli/commands/proposal.rs
@@ -670,20 +670,20 @@ pub async fn create_update_guardian_proposal(
     Ok(())
 }
 
-/// Create an emergency pause (or, with `pause == false`, unpause) proposal.
+/// Create an emergency pause (or, with `unpause == true`, unpause) proposal.
 ///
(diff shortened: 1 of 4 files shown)


crates/hashi/src/cli/commands/proposal.rs
```diff
@@ -670,6 +670,41 @@ pub async fn create_update_guardian_proposal(
...
+/// normal 2/3 supermajority. Both paths target `emergency_pause::propose`.
+pub async fn create_emergency_pause_proposal(
+    config: &CliConfig,
+    pause: bool,
+    metadata: Vec<(String, String)>,
+    tx_opts: &TxOptions,
+) -> Result<()> {
+    let action = if pause { "Pause" } else { "Unpause" };
+    let title = format!("Creating Emergency {action} Proposal:");
+    print_detail(&format!("\n{}", title.as_str().bold()));
```
dlukegordon: I think `.as_str()` isn't necessary here

crates/hashi/src/cli/mod.rs
```diff
@@ -896,6 +910,15 @@ pub async fn run(opts: CliGlobalOpts, command: CliCommand) -> anyhow::Result<()>
                     )
                     .await?;
                 }
+                CreateProposalCommands::EmergencyPause { unpause, metadata } => {
+                    commands::proposal::create_emergency_pause_proposal(
+                        &config,
+                        !unpause,
```
dlukegordon: Should we just make the function take unpause instead of pause?

crates/hashi/src/cli/commands/proposal.rs
```diff
@@ -670,6 +670,41 @@ pub async fn create_update_guardian_proposal(
     Ok(())
 }
 
+/// Create an emergency pause (or, with `pause == false`, unpause) proposal.
```
dlukegordon: I think `design/docs/node-operator-runbook.mdx` needs to be updated as well

crates/hashi/src/cli/commands/proposal.rs
Bridgerz: Good catch — dropped it. `colored` implements `Colorize` for `String`, so `title.bold()` works directly. Fixed in 722782411.

crates/hashi/src/cli/mod.rs
Bridgerz: Done — `create_emergency_pause_proposal` now takes `unpause`, so the dispatch no longer negates. The `pause: !unpause` conversion lives at the single point where the proposal params are built. 722782411.

crates/hashi/src/cli/commands/proposal.rs
Bridgerz: Added to the runbook in 722782411: §6.6 now lists `proposal create emergency-pause [--unpause]`, and the §7.2 note that called it "not part of the standard `proposal create` CLI" is updated.

dlukegordon (approved):

PR #697, opened 2026-06-19
Batch partial-signature collection per peer

zhouwfang:
#179 #515 #613


crates/hashi/src/mpc/signing.rs
```diff
@@ -371,117 +472,164 @@ impl SigningManager {
...
-            )
-            .await;
+        Ok((public_nonce, partial_sigs))
+    }
+}
+
+pub struct SignInput {
+    pub signing_id: Address,
+    pub message: Vec<u8>,
+    pub global_presig_index: u64,
```
bmwill, line 482: do we want to include an epoch here so its a tuple (epoch, presig idx)?

bmwill (approved):
lgtm with a small question

crates/hashi-types/proto/sui/hashi/v1alpha/mpc_service.proto
```diff
@@ -153,11 +153,13 @@ message GetPartialSignaturesRequest {
...
 }
 
 message GetPartialSignaturesResponse {
-  // Partial signatures.
-  optional sui.rpc.v2.Bcs partial_sigs = 1;
+  // Partial signatures keyed by signing_id.
+  // Only ids for which the peer has already computed its local partial are present;
+  // an absent id means "not ready yet", not an error.
+  map<string, sui.rpc.v2.Bcs> partial_sigs = 1;
 }
```
bmwill, line 165: If we wanted to improve things even further we could make this RPC return a stream of partial sigs so that as they become available we stream them back to the requester. Its a further optimization we can do but not necessary if we time bound the response (which we do with the timeout in `sign`)

crates/hashi/src/mpc/signing.rs
zhouwfang: Hmm not needed — the exchange is already epoch-gated at the RPC layer.

bmwill: missed that, thanks!

crates/hashi-types/proto/sui/hashi/v1alpha/mpc_service.proto
zhouwfang: Good future optimization. Created an issue about it.

PR #698, opened 2026-06-19
Add network rx/tx bandwidth metrics

dlukegordon:


bmwill (changes requested):
I don't think this is necessary, all of our instances should already have cpu and network bandwidth metrics available. Cc @tharbert for this.

dlukegordon:
Closing since we already push bandwidth metrics from cadvisor

PR #705, opened 2026-06-22
[guardian] verify the mock attestation in dev builds

0xsiddharthks:
Off-enclave (`non-enclave-dev`) the guardian's attestation was a placeholder string and #675's `verify` a no-op, so the PCR0 / signing-key binding was never exercised in dev.

- `NitroAttestation::mock(signing_key, pcr0)` encodes the fields `verify` checks; the off-enclave guardian emits it instead of the placeholder.
- The `non-enclave-dev` `verify` arm checks that binding (`verify_mock_attestation`) rather than returning `Ok`; `cfg(test)` stays a no-op so existing tests are undisturbed.

Stacked on #691; touches #675's attestation module, so flagging for Deepak.

Now stacked on #675, a dev deploy pins an empty git revision, so the node verifies signature-only and only reaches this when a real build is pinned. It's a thin completeness check on top of #675's git-revision match — fine to fold into #675 or drop if you'd prefer.

### DIFF
--- a/crates/e2e-tests/src/lib.rs
+++ b/crates/e2e-tests/src/lib.rs
@@ -253,7 +253,8 @@ impl TestNetworksBuilder {
         let guardian_btc_pubkey = guardian_harness.ensure_btc_pubkey()?;
         let guardian_config = hashi::publish::GuardianConfig {
             url: guardian_harness.endpoint().to_string(),
-            // Non-Nitro dev: attestation is a no-op, so an all-zero build is fine.
+            // Non-Nitro dev: no real enclave to attest, so leave the build
+            // unpinned (empty git revision) — the node verifies signature-only.
             git_revision: String::new(),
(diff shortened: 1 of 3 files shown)


0xsiddharthks:
Closing — subsumed by #675 once the stack rebased on top of it.

The meaningful dev-side check (guardian build identity) is now #675's `untrusted_git_revision == expected build` check in `GetGuardianInfoResponse::verify`, which runs in all builds. And #691 routes devnet to signature-only (empty on-chain git revision → no build pinned), so the strict `non-enclave-dev` attestation arm this PR adds isn't reached on the node path anyway.

The mock-attestation machinery here (`NitroAttestation::mock` + `verify_mock_attestation`) would only matter if dev/staging pins a real git revision to exercise the full attestation path — out of scope for now, and its cleanest home would be #675's attestation module. Branch kept (reopenable) for reference.

PR #706, opened 2026-06-22
Add `nonce_generation_protocol` MPC config selector

zhouwfang:
#531


crates/hashi/src/mpc/types.rs
```diff
@@ -40,6 +40,31 @@ pub struct NonceMessage {
...
+            1 => Self::Avid,
+            other => {
+                let default = Self::default();
+                tracing::warn!(
+                    value = other,
+                    ?default,
+                    "unknown mpc_nonce_generation_protocol; using default"
+                );
+                default
+            }
```
zhouwfang, line 65: This becomes necessary because on-chain value validation for the MPC config params was removed in #506. Created an issue to add the validation back.

zhouwfang: Restored in #710.

packages/hashi/sources/core/committee/committee.move
```diff
@@ -42,6 +42,8 @@ public struct Committee has copy, drop, store {
     mpc_weight_reduction_allowed_delta: u64,
     /// MPC max faulty parties in basis points
     mpc_max_faulty_in_basis_points: u64,
+    /// Nonce-generation protocol: 0 = vanilla, 1 = AVID
+    mpc_nonce_generation_protocol: u64,
```
bmwill, line 47: So this is a semi-issue since we can't ever add need fields to this struct once we go to mainnet, so we probably should embed a "Config" into this struct and copy the values we need to pin for the epoch so that we can trivially add new configs later.

bmwill: cc @Bridgerz i think we should maybe look at making `Config` a bit more general purpose (maybe pulling out the allowed versions and upgrade cap) so we can trivially reuse the infra for `Config`s in other places in our package. Thoughts?

zhouwfang: Good call-out. This field itself is transitional and will be removed before mainnet. The concern does hold for the other params that persist and any future epoch-pinned config. I think embedding a general `Config` into `Committee` is the right fix, but as a dedicated change, not this PR.

zhouwfang: Bonus for that refactor: an embedded `Config` (with default-on-missing accessors) lets you add or drop pinned params post-mainnet via KV, no struct edit. So even this selector wouldn't need deleting before the freeze, as we'd just drop the key once vanilla is gone.

zhouwfang: I created an issue: IOP-455.

bmwill: Yep agreed, we can address this in a follow up

bmwill (approved):

PR #711, opened 2026-06-24
[onchain] Reconnect checkpoint watcher when the Sui stream stalls

0xsiddharthks:
This fixes an issue where the on-chain watcher's checkpoint subscription can hang forever on a half-open gRPC stream, freezing the node's view of on-chain state.

## Changes

- Wrap each checkpoint read in a 120s `timeout`; a stalled stream now exits the loop instead of hanging (the SDK client's h2 keepalive only catches fully-dead connections, not a live connection whose stream went silent).
- Re-subscribe and rescrape after the loop exits (stall, stream error, or close) to recover any gap, and rebuild the Sui client on each reconnect so a stalled h2 connection isn't reused.

### DIFF
--- a/crates/hashi/src/onchain/watcher.rs
+++ b/crates/hashi/src/onchain/watcher.rs
@@ -31,8 +31,14 @@ use crate::onchain::types::ProposalType;
 use crate::onchain::types::WithdrawalRequest;
 use crate::withdrawals::withdrawal_limiter_consumption_amount;
 
+/// Reconnect if the checkpoint stream goes silent this long. A half-open h2
+/// stream yields neither an item nor an error, so an unbounded read hangs — the
+/// SDK's keepalive only trips on a fully-dead connection, not a live one whose
+/// server silently stopped sending checkpoints.
+const CHECKPOINT_STREAM_STALL_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(120);
+
 #[tracing::instrument(name = "watcher", skip_all)]
-pub async fn watcher(mut client: Client, state: OnchainState, metrics: Option<Arc<Metrics>>) {
+pub async fn watcher(sui_rpc_url: String, state: OnchainState, metrics: Option<Arc<Metrics>>) {
     let subscription_read_mask = FieldMask::from_paths([
         Checkpoint::path_builder().sequence_number(),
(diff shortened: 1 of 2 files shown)


crates/hashi/src/onchain/watcher.rs
```diff
@@ -100,7 +115,22 @@ pub async fn watcher(mut client: Client, state: OnchainState, metrics: Option<Ar
...
+                    Ok(None) => break,
+                    Err(_elapsed) => {
+                        tracing::warn!(
+                            "checkpoint stream stalled for {}s; reconnecting",
+                            CHECKPOINT_STREAM_STALL_TIMEOUT.as_secs()
+                        );
+                        rescrape_state = true;
+                        break;
+                    }
+                };
```
bmwill: I would just have this be 
```suggestion
while let Ok(Some(item)) = timeout(CHECKPOINT_STREAM_STALL_TIMEOUT, subscription.next()).await {
```
Theres no real strong reason to have all this other logic. (we may need to say to rescrape if the while loop exits though)

crates/hashi/src/onchain/watcher.rs
```diff
@@ -54,6 +58,17 @@ pub async fn watcher(mut client: Client, state: OnchainState, metrics: Option<Ar
     let mut rescrape_state = false;
 
     loop {
+        // Fresh client each reconnect so a half-open h2 channel isn't reused.
+        let mut client = match Client::new(&sui_rpc_url) {
```
bmwill: I feel like there should be some timeout paramaters that we should be able to set instead of needing to build a new client every time. Internally the Client will build a new h2 connection as needed

dlukegordon (approved):

crates/hashi/src/onchain/watcher.rs
```diff
@@ -160,6 +164,9 @@ pub async fn watcher(mut client: Client, state: OnchainState, metrics: Option<Ar
                 metrics.update_onchain_state(&state);
             }
         }
+
+        // Our subscription broke (stall, error, or close); rescrape on reconnect.
```
bmwill: Filtered list apis are coming and soon we wont need to rescrape!

bmwill (approved):

PR #713, opened 2026-06-25
[onchain] Bound scrape page size + raise gRPC decode limit to 32 MiB

0xsiddharthks:
Defensive hardening for scraping a large on-chain state. `scrape_hashi` reads the full deposit/withdrawal queues and committee data; at sufficient scale a single response can overflow the scrape client's gRPC decode limit (`unexpected end of input`) or just be needlessly large.

- Bound the `list_*` page size to 1000 — the SDK still pages through every entry.
- Apply the decode limit to the scrape clients (they read the largest responses but used tonic's 4 MiB default) and raise that default to 32 MiB.

### DIFF
--- a/crates/hashi/src/onchain/mod.rs
+++ b/crates/hashi/src/onchain/mod.rs
@@ -42,6 +42,10 @@ use hashi_types::move_types;
 
 const BROADCAST_CHANNEL_CAPACITY: usize = 100;
 
+/// Bounded so a huge queue isn't returned as one oversized page that overflows
+/// the gRPC decode limit; the SDK still pages through every entry.
+const SCRAPE_PAGE_SIZE: u32 = 1000;
+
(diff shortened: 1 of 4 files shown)


crates/hashi/src/onchain/mod.rs
```diff
@@ -705,6 +705,10 @@ async fn scrape_package_versions(
     Ok(package_versions)
 }
 
+/// Bounded so a huge queue isn't returned as one oversized page that overflows
+/// the gRPC decode limit; the SDK still pages through every entry.
+const SCRAPE_PAGE_SIZE: u32 = 1000;
```
dlukegordon: Nit: maybe move this to the top of the file since it's used in more places than just `scrape_hashi`?

dlukegordon (approved):

PR #715, opened 2026-06-25
[guardian] add hashi-guardian-proxy. cache + restricted forwarding + provisioning relay

0xsiddharthks:
We are adding an explicit `hashi-guardian-proxy` , which is the out-of-enclave front for the guardian, and the only internet-facing piece allowing the guardian to stay VPC internal.

This will be doing the following:
- cache withdrawals
- forward just the RPCs a node actually calls
- run the relay that provisions the guardian from KP shares.

## Changes

- `cache.rs`: wid-keyed `StandardWithdrawal` cache. This replaces the existing WID cache in the guardian. (which is removed in #740)
- `forward.rs`: forwards the four node RPCs and denies the operator/ceremony ones + raw `ProvisionerInit` with `PERMISSION_DENIED`.
- `relay.rs` + `guardian_relay_service.proto`: the `SingleProvisionerInit` relay. 
  - This does the following:
    - it pins the backend session
    - dedupes by id
    - accumulates a threshold
    - then submits one batch `ProvisionerInit`
- `kp_provision.rs`: the old `todo!()` is now a real `SingleProvisionerInit` call


crates/hashi-guardian-proxy/src/relay.rs
```diff
@@ -0,0 +1,289 @@
...
+impl GuardianRelayService for Relay {
+    async fn single_provisioner_init(
+        &self,
+        request: Request<proto::SingleProvisionerInitRequest>,
+    ) -> Result<Response<proto::SingleProvisionerInitResponse>, Status> {
+        let req = request.into_inner();
+        let share = req
+            .encrypted_share
+            .ok_or_else(|| Status::invalid_argument("missing encrypted_share"))?;
+        let id = share_id(&share)
```
dlukegordon, line 160: Should we do a quick sanity check here that the passed share id is valid, ie > 0 and <= num_shares?

crates/hashi-guardian-proxy/src/relay.rs
```diff
@@ -0,0 +1,289 @@
...
+        }
+        let threshold = status.threshold.ok_or_else(|| {
+            Status::failed_precondition(
+                "guardian has no secret_sharing_instance yet; run `operator provision` first",
+            )
+        })?;
+
+        // Hold the accumulator across the possible batch submission so at most
+        // one ProvisionerInit is ever in flight.
+        let mut acc = self.accumulator.lock().await;
```
dlukegordon: Should we move this up, before `let status = self.backend_status().await?;`

dlukegordon:
In `crates/hashi-guardian/src/main.rs:61`, we are still wrapping the enclave guardian in the old `CachingGuardianGrpc`, so the cache now exists both in the new proxy and inside the guardian. Should we remove this?

0xsiddharthks:
@dlukegordon , Yes, we are already removing the inprocess cache of the guardian in one of the following PRs in the stack ( #740 )

crates/hashi-guardian-proxy/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+[dependencies]
+tokio.workspace = true
+anyhow.workspace = true
+tracing.workspace = true
+tonic.workspace = true
+tonic-health.workspace = true
+hashi-types = { path = "../hashi-types" }
+
+[dev-dependencies]
+tokio-stream = "0.1"
```
dlukegordon: I think we can use the workspace dep for this

crates/hashi-guardian-proxy/src/cache.rs
```diff
@@ -0,0 +1,397 @@
...
+struct CacheEntry {
+    /// The seq the guardian consumed this wid at. Kept for observability only —
+    /// the cache is keyed by `wid`, so lookups ignore the requester's seq.
+    consumed_seq: u64,
+    response: proto::SignedStandardWithdrawalResponse,
+}
+
+pub struct CachingGuardianGrpc<S> {
+    inner: S,
+    cache: Mutex<HashMap<WithdrawalID, CacheEntry>>,
```
dlukegordon: This is unbounded, should we add a TTL or max number of entries or something?

0xsiddharthks: Thanks! Yeah, previously since devnet was small I had an unbounded cache. just fixed it. I added a 10K cache capacity for now.

dlukegordon (approved):

PR #718, opened 2026-06-25
make config general purpose

bmwill:
#368 #531


zhouwfang (changes requested):
`test_upgrade_v1_to_v2` fails because the refactor moved const `PACKAGE_VERSION` to `versioning.move` but two Rust consumers still hardcode `config.move`, which has no test and will silently break upgrade builds. Repointing both to `sources/core/versioning.move` should fix it.

packages/hashi/sources/core/config/config.move
```diff
@@ -76,12 +82,6 @@ public(package) fun is_valid_config_update(self: &Config, key: &String, value: &
 
 // ======== Core Accessors ========
 
-/// Assert that the package version is the current version.
-#[allow(implicit_const_copy)]
-public(package) fun assert_version_enabled(self: &Config) {
-    assert!(self.enabled_versions.contains(&PACKAGE_VERSION), EVersionDisabled);
-}
-
 public(package) fun paused(self: &Config): bool {
```
zhouwfang, line 85: The other accessors use `try_get` and default, but this uses `get()`. Worth making uniform, since the same `Config` type also backs the `empty()` per-epoch snapshot.

packages/hashi/sources/core/mpc_config.move
```diff
@@ -82,3 +82,51 @@ public(package) fun init_defaults(config: &mut Config) {
...
     );
 }
+
+/// Snapshot the MPC parameters from `config` into a fresh store to pin onto a
+/// `Committee` for the epoch. The full key set is always materialized (using
+/// defaults for absent keys) and inserted in a fixed canonical order, so the
+/// snapshot's BCS encoding is deterministic. The committee is signed, so the
+/// Rust mirror (`move_types::Committee`'s `From`) must build this map in the
+/// same order with the same keys.
+public(package) fun pin(config: &Config): Config {
```
zhouwfang, line 92: `pin` puts a `VecMap<String,Value>` into the signed committee, mirrored by hand in Rust. The BCS bytes must match exactly on both sides, so a future param or key-order change on one side would silently break handoff-cert verification. A unit test pinning the bytes would guard against that.

bmwill: ah this is a gap, will fix (we should probably not construct the order by hand) and instead just have the order be what is in the Move object)

Bridgerz (approved):
From a move perspective this feels solid. Versioning module is much more ergonomic and generalised config is 🤌

zhouwfang:
Is this ready for review now? I asked because there is no re-request to review:
Screenshot 2026-06-29 at 12 05 41 PM

bmwill:
@zhouwfang yes it is

crates/hashi-types/src/guardian/proto_conversions.rs
zhouwfang, line 1305: nit: set it to be 1 so the test actually exercises the parameter the old wire format dropped.

zhouwfang (approved):
nit: https://github.com/MystenLabs/hashi/pull/718#discussion_r3477689723 is not resolved, but it is minor.

PR #719, opened 2026-06-26
Adopt `fastcrypto` AVSS audit changes

zhouwfang:
#14 #95 #179


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -3913,16 +4009,39 @@ impl MpcManager {
...
+        } else {
+            self.previous_committee.as_ref().ok_or_else(|| {
+                MpcError::InvalidConfig("No previous committee for cross-epoch complaint".into())
+            })
         }
     }
 
+    fn accuser_party_id(&self, epoch: u64, caller: &Address) -> MpcResult<PartyId> {
+        self.committee_for_epoch(epoch)?
+            .index_of(caller)
```
jonas-lj, line 4031: I think there are quite a few calls to `index_of` here in different contexts and with different errors. Perhaps create a single getter with a consistent error?

jonas-lj (approved):

PR #730, opened 2026-06-29
Enforce committee stake floor only at genesis bootstrap

Bridgerz:
## What

`start_reconfig` asserted that the forming committee represented **≥95% of total Sui stake on every epoch change**. This change enforces that floor **only at genesis bootstrap** and skips it on normal reconfigs.

## Why

The 95% floor only makes sense as a launch gate. Once the bridge has bootstrapped, Hashi must follow Sui's validator set on every epoch change *regardless* of how much stake re-registered. Enforcing the floor on a normal reconfig would let validators **brick reconfiguration** — and with it all deposits/withdrawals — simply by withholding registration. A committee that ends up too small to complete key resharing is already handled off-chain and via `abort_reconfig`.

## How

- Gate the check on `self.mpc_public_key.is_empty()` — the same initial-reconfig signal `end_reconfig` already uses (the MPC key is minted once, at the genesis DKG).
- Move the decision into a pure `threshold::genesis_stake_satisfied(is_bootstrap, committee_weight, total_sui_weight)` and replace the magic `9500` with a named `GENESIS_STAKE_THRESHOLD_BPS`.
- Behaviour vs. `main`: identical at genesis (enforce 95%); no enforcement afterward. Gas is unchanged from baseline (the stake sum was already computed unconditionally).

## Tests

- `packages/hashi/tests/threshold_tests.move` — decision-table unit tests, including the rows that pin this change: a post-bootstrap reconfig at 0 / 1 / 500 / 949-of-1000 committee weight all pass; genesis still aborts below 95%.
- An end-to-end test of the skip (a normal reconfig actually proceeding under 95%) is **deferred**: it requires growing the validator set mid-run (persistent `next_epoch` keys mean a static set never drops below its genesis fraction), which the e2e harness has no seam for today.

### DIFF
--- a/packages/hashi/sources/core/committee/committee_set.move
+++ b/packages/hashi/sources/core/committee/committee_set.move
@@ -465,15 +465,25 @@ public(package) fun start_reconfig(
         ctx,
     );
 
-    // Ensure 95% of stake has registered.
-    let mut sui_system_weight = 0;
-    let (_, weights) = sui_system.active_validator_voting_powers().into_keys_values();
-    weights.do!(|weight| {
-        sui_system_weight = sui_system_weight + weight;
-    });
-    assert!(
-        committee.total_weight() >= hashi::threshold::weight_threshold(sui_system_weight, 9500),
-    );
+    // At genesis bootstrap the forming committee must represent >=95% of total
+    // Sui stake, so the bridge only launches once a supermajority of validators
+    // has registered. After bootstrap, Hashi follows Sui's validator set
+    // unconditionally -- enforcing a floor on a normal reconfig would let
+    // validators brick reconfiguration (and with it all deposits/withdrawals)
+    // by withholding registration.
+    if (self.mpc_public_key.is_empty()) {
+        let mut sui_system_weight = 0;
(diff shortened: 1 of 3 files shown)


packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -477,14 +477,24 @@ public(package) fun start_reconfig(
...
         sui_system_weight = sui_system_weight + weight;
     });
     assert!(
-        committee.total_weight() >= hashi::threshold::weight_threshold(sui_system_weight, 9500),
+        hashi::threshold::genesis_stake_satisfied(
+            self.mpc_public_key.is_empty(),
+            committee.total_weight(),
+            sui_system_weight,
+        ),
     );
```
bmwill: can we have this be an `if genesis { assert }` and then the "is_bootsrap" logic can be removed from the fn in threshold?

Bridgerz: good with me 👍

bmwill (approved):
thanks! Before testnet we may need to lower this more, or maybe even come up with a different trigger, we'll see

PR #733, opened 2026-06-30
Add the AVID certificate layer

zhouwfang:
#531

### DIFF
--- a/crates/hashi/src/mpc/types.rs
+++ b/crates/hashi/src/mpc/types.rs
@@ -4,11 +4,15 @@
 //! Core types for the DKG protocol
 
 use fastcrypto::error::FastCryptoError;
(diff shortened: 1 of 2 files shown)


crates/hashi/src/mpc/types.rs
```diff
@@ -462,6 +468,96 @@ impl CertificateV1 {
...
+    fn signers(&self) -> &BTreeSet<PartyId> {
+        &self.signers
+    }
+
+    fn payload(&self) -> &P {
+        &self.payload
+    }
+
+    fn verify(&self) -> FastCryptoResult<()> {
+        self.committee
```
jonas-lj, line 498: Perhaps add a note that it's checked on construction that the payload for this and dealer_cert are the same.

jonas-lj (approved):

PR #734, opened 2026-06-30
Add AVID round-state storage

zhouwfang:
#531

### DIFF
--- a/crates/hashi/src/mpc/types.rs
+++ b/crates/hashi/src/mpc/types.rs
@@ -7,6 +7,7 @@ use fastcrypto::error::FastCryptoError;
 use fastcrypto::error::FastCryptoResult;
 use fastcrypto::hash::Blake2b256;
 use fastcrypto::hash::HashFunction;
+use fastcrypto_tbls::ecies_v1::Ciphertext;
 use fastcrypto_tbls::ecies_v1::PrivateKey;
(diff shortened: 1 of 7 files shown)


crates/hashi/src/mpc/types.rs
```diff
@@ -59,6 +60,13 @@ pub enum AvidNonceMessageKind {
     },
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct AvidRoundState {
+    pub common: batch_avss_avid::AvssCommonMessage,
```
jonas-lj, line 71: Shouldn't the content of the AvidVote be here too - the top_root and the recipients? Or is that supposed to be stored somewhere else? It's needed to verify echoes, so we need it somewhere.

zhouwfang: The AvidVote is the Vote cert's payload — it lives on the TOB (H(AvidVote) on-chain and full AvidVote P2P, hash-pinned) and is re-fetched on reload, not duplicated here. AvidRoundState holds only the node's locally-gathered, unrecoverable state (v, own ciphertext, echoes). Echo re-verification needs the verified cert anyway, so raw top_root/recipients alone wouldn't suffice. Reload/resume is a follow-up PR; this one is just persist/load/prune.

jonas-lj (approved):

PR #736, opened 2026-07-01
Process stale deposit approvals on epoch changes to avoid waiting for the next Bitcoin block

dlukegordon:


crates/hashi/src/leader/deposits.rs
```diff
@@ -96,6 +97,17 @@ impl LeaderService {
         }
     }
 
+    /// Clear all ConfirmDepositFailed entries. Called when an `EndReconfig`
```
zhouwfang: Redundant with prune() just below. It already drops all non-current-epoch entries each checkpoint. "Recover at once" also overstates it: a stranded deposit needs re-approval before it can confirm.

crates/hashi/src/leader/retry.rs
```diff
@@ -76,6 +76,13 @@ where
         self.state.lock().unwrap().remove(request_id);
     }
 
+    pub(super) fn clear_by_kind(&self, kind: K) -> usize {
```
zhouwfang: Fine as a helper, but its value hinges on the redundancy question.

crates/hashi/src/leader/mod.rs
```diff
@@ -311,6 +313,11 @@ impl LeaderService {
                     }
                     Self::log_task_result("guardian_committee_reconcile", result);
                 }
+                Ok(notification) = notification_rx.recv() => {
```
zhouwfang: broadcast::recv() can return Err(Lagged/Closed), which the Ok(...) = pattern silently drops. Match Err explicitly, warn! on lag.

zhouwfang (commented):
I think clear_by_kind is redundant with the per-checkpoint prune() — at EndReconfig(N), every ConfirmDepositFailed entry is a deposit approved in an epoch < N, which prune() drops next checkpoint anyway. Can you add a test where clear_by_kind frees a deposit prune() wouldn't?

crates/hashi/src/leader/deposits.rs
```diff
@@ -26,14 +26,30 @@ use tracing::info;
...
-        self.reload_pending_unapproved_deposit_requests();
+    pub(super) fn process_deposits_on_bitcoin_block(&mut self) {
+        self.reload_pending_unapproved_deposit_requests(UnapprovedDepositReloadMode::All);
         self.process_unapproved_deposit_requests();
     }
 
-    fn reload_pending_unapproved_deposit_requests(&mut self) {
+    pub(super) fn process_stale_unapproved_deposits_if_new_epoch(&mut self) {
+        let current_epoch = self.inner.onchain_state().epoch();
+        if !self.is_reconfiguring() && self.last_unapproved_deposit_epoch != Some(current_epoch) {
```
zhouwfang, line 46: The guard is a heuristic ("I'm not mid-reconfig"), not a guarantee ("my view is current"). Agree?

zhouwfang (commented):
Could you add a test that epoch change re-processes a stale-cert deposit with no new BTC block? It should fail on main.

crates/hashi/src/leader/deposits.rs
dlukegordon: Yes. We use this same check elsewhere to prevent useless operations where the Move calls would be rejected. If the view is not current, and we proceed despite actually being in reconfig, the Move calls would just be rejected, so not a big issue.

zhouwfang: Could you add a comment about the assumption?

zhouwfang:
#613

dlukegordon:
FYI I already have this linked<br>[IOP-472](https://linear.app/mysten-labs/issue/IOP-472/improve-deposit-processing-during-epoch-changes)

zhouwfang:
> FYI I already have this linked[IOP-472](https://linear.app/mysten-labs/issue/IOP-472/improve-deposit-processing-during-epoch-changes)

It is just easier for me to find the PR's I closely worked with this way.

crates/hashi/src/leader/deposits.rs
dlukegordon: Added

dlukegordon:
I pushed a test, and confirms it fails on main

zhouwfang:
> I pushed a test, and confirms it fails on main

I don't see the test. Forgot to push the commit?

dlukegordon:
My bad I forgot to push, it's up now

crates/e2e-tests/src/e2e_flow.rs
```diff
@@ -350,6 +380,90 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_stale_cert_deposit_reprocessed_after_epoch_change_without_btc_block() -> Result<()>
```
zhouwfang, line 397: It would be great to have the negative case too (a brand-new/uncertified deposit is not picked up), so it doesn't explicitly pin StaleEpochApprovalOnly vs All.

zhouwfang (approved):

crates/e2e-tests/src/e2e_flow.rs
dlukegordon: Pushed

PR #738, opened 2026-07-01
Add AVID optimistic-phase crypto wrappers

zhouwfang:
#531

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -11,6 +11,8 @@ use crate::metrics::MPC_LABEL_DKG;
 use crate::metrics::MPC_LABEL_KEY_ROTATION;
 use crate::metrics::MPC_LABEL_NONCE_GENERATION;
 use crate::metrics::Metrics;
+use crate::mpc::types::AvidNonceMessage;
+use crate::mpc::types::AvidNonceMessageKind;
 use crate::mpc::types::AvidRoundState;
 use crate::mpc::types::CertificateV1;
 pub use crate::mpc::types::ComplainRequest;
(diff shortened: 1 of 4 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1983,6 +1988,125 @@ impl MpcManager {
...
+        let builder = dealer
+            .create_avss_messages(rng)
+            .map_err(|e| MpcError::CryptoError(e.to_string()))?;
+        let messages = self
+            .committee
+            .members()
+            .iter()
+            .enumerate()
+            .map(|(j, member)| {
+                let message = builder.message_for(j as u16).ok_or_else(|| {
```
jonas-lj: nit: you can just zip with the builder.messages iterator here.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1983,6 +1988,125 @@ impl MpcManager {
...
+                    MpcError::CryptoError(format!("no optimistic message for recipient {j}"))
+                })?;
+                Ok((
+                    member.validator_address(),
+                    Messages::NonceGenerationAvid(AvidNonceMessage {
```
jonas-lj, line 2063: nit: I think you can produce all of these lazily as an iterator from the AvssMessageBuilder instead of creating all of them here.

jonas-lj (approved):

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Agreed that's the right shape once we persist the builder — I'd just defer it. In the optimistic phase the dealer sends every recipient a message anyway, so lazy saves nothing here, and doing it now means returning only the builder and pushing the message derivation into the send path, which is a later PR. It earns its keep once the dealer persists the builder (via your PR) and re-derives on restart. I'll switch to "hold the builder, derive on demand" then.

PR #739, opened 2026-07-02
[guardian] Surface the BTC master pubkey at the ceremony

0xsiddharthks:
## Summary

The guardian's BTC master pubkey has to be recorded on-chain (`guardian_btc_public_key`) before the guardian can be provisioned — but it was only observable via `GetGuardianInfo` *after* `provisioner_init`, which needs an on-chain committee, which needs the pubkey. This breaks that cycle by surfacing the x-only pubkey straight from the ceremony, where the enclave already holds the freshly-generated secret.

## Changes

- Add `btc_master_pubkey` to `SetupNewKeyResponse` (+ proto) and the signed `ceremony/` audit log (`CeremonyLogMessage::{NewKey,Rotate}`) — so it's covered by the response signature and cross-checkable against the on-chain key from the immutable log.
- `operator ceremony` prints `GUARDIAN_BTC_PUBKEY=<x-only hex>` once every ceremony check passes, for the deploy to capture and publish.
- Validate the pubkey against the S3 ceremony log before publishing: the existing `ceremony/`-vs-response cross-check (`logged == live`) now covers `btc_master_pubkey`, so the on-chain key can't diverge from the immutable audit log (Luke's `469008980`, plus a regression test that locks the pubkey into that equality).
- New `k256_sk_to_btc_xonly_pubkey` helper; updated tests + regenerated proto.

Bottom of the guardian-enclave-migration stack; unblocks dropping `dev-bootstrap`.


dlukegordon (commented):
As discussed in our DMs, please consider the changes I proposed [here](<https://github.com/MystenLabs/hashi/commit/469008980054ef2ed4c08768e5e12875eed6ac21>).

0xsiddharthks:
@dlukegordon just cherry-picked your commit! Thanks!

dlukegordon (approved):

PR #742, opened 2026-07-02
[guardian] Authenticate relay share submissions with KP signatures

0xsiddharthks:
## Summary

The `SingleProvisionerInit` call is currently open to the internet and unauthenticated.
 
So we are requiring submissions to be signed by a KP on a configured roster.

## Changes

- KPs detached-sign the exact (session, share) bytes with their offline key
  - the relay checks the signer against its roster and verifies the signature before buffering (`signer_cert` + `kp_signature` on the request).
- The roster is proxy config (`AUTHORIZED_KP_FINGERPRINTS`), sourced from the ceremony; empty fail-closes the relay. 
  - The enclave still verifies every share.
- The local replica's `make ceremony` hands the roster to its proxy.


crates/hashi-guardian-proxy/src/relay.rs
```diff
@@ -187,6 +240,17 @@ impl GuardianRelayService for Relay {
                 )),
             };
         check_share_id(id, num_shares)?;
+
+        // Authenticate before buffering: only a rostered KP may enqueue a
+        // share, so a stranger can't slip in a bad one and fail the batch.
+        verify_kp_submission(
```
dlukegordon: I think we can move this to before we acquire the lock.

docker/hashi-guardian-local/Makefile
```diff
@@ -27,6 +27,10 @@ ceremony: ## Generate KP keys + run the genesis ceremony (captures the BTC pubke
 	$(COMPOSE) --profile ceremony up -d --build ceremony
 	$(COMPOSE) --profile init run --rm --build init /scripts/ceremony.sh
 	$(COMPOSE) --profile ceremony stop ceremony
+	@fps=$$($(COMPOSE) --profile init run --rm -T --entrypoint cat init /work/kp-fingerprints.csv); \
+	echo "AUTHORIZED_KP_FINGERPRINTS=$$fps" > kp-roster.env; \
+	echo "Handing the KP roster to the proxy relay: $$fps"; \
+	$(COMPOSE) up -d proxy
```
dlukegordon, line 34: If the `compose run ... cat /work/kp-fingerprints.csv` fails (stale image, missing file), fps is empty, the recipe still writes `AUTHORIZED_KP_FINGERPRINTS=` and restarts the proxy, and we get a fail-closed relay with no obvious cause until provisioning mysteriously `412`s. The `localnet-cmd` target already guards its equivalent (`if [ -z "$$pubkey" ] ...`, Makefile:40-41); let's do the same here.

crates/hashi-guardian-proxy/src/relay.rs
```diff
@@ -116,6 +126,41 @@ impl Relay {
...
+        return Err(Status::unauthenticated(
+            "submission is missing signer_cert or kp_signature",
+        ));
+    }
+    let cert = PgpPublicCert::new(signer_cert.to_string())
+        .map_err(|e| Status::unauthenticated(format!("invalid signer_cert: {e}")))?;
+    let fingerprint = normalize_fingerprint(&cert.fingerprint());
+    if !authorized_kp_fingerprints
+        .iter()
+        .any(|f| normalize_fingerprint(f) == fingerprint)
```
dlukegordon: `parse_kp_roster` already normalizes every entry at startup ([config.rs:62](<http://config.rs:62>)), then the relay re-normalizes each roster entry on every submission. We could normalize once in `Relay::new` and compare directly.<br><br>Another alternative is to create a `Fingerprint` type (or use an existing one) that normalizes on construction.

crates/hashi-types/src/pgp.rs
```diff
@@ -339,6 +350,87 @@ pub fn decrypt_armored_via_gpg(armored: &str, homedir: Option<&Path>) -> Result<
...
+    command.stdout(Stdio::piped());
+    command.stderr(Stdio::inherit());
+    let mut child = command
+        .spawn()
+        .context("Failed to run `gpg --detach-sign`; is gpg installed and on PATH?")?;
+    child
+        .stdin
+        .take()
+        .ok_or_else(|| anyhow::anyhow!("Failed to open gpg stdin"))?
+        .write_all(payload)
```
dlukegordon: If this `write_all fails` (e.g. `EPIPE` because gpg died early), we return before `wait_with_output()` and the child is never reaped, so we get a zombie until the CLI exits. We could capture the write result instead of ?-ing it, call `wait_with_output()`, and check the exit status before surfacing the write error. Minor since this is a one-shot CLI, but the decrypt paths already reap carefully.

```
let write_result = match child.stdin.take() {
    Some(mut stdin) => stdin.write_all(payload),
    None => Err(io::Error::other("failed to open gpg stdin")),
};
// Dropping the `ChildStdin` above signals EOF, making gpg emit the signature.
// Reap unconditionally, even if the write failed (EPIPE = gpg died early).
let output = child
    .wait_with_output()
    .context("wait for `gpg --detach-sign`")?;
if !output.status.success() {
    anyhow::bail!("`gpg --detach-sign` exited with status {}", output.status);
}
write_result.context("write payload to gpg stdin")?;
String::from_utf8(output.stdout).context("gpg produced a non-UTF8 armored signature")
```

dlukegordon (approved):

PR #743, opened 2026-07-02
Add AVID pessimistic-phase crypto wrappers

zhouwfang:
#531

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -11,6 +11,7 @@ use crate::metrics::MPC_LABEL_DKG;
 use crate::metrics::MPC_LABEL_KEY_ROTATION;
 use crate::metrics::MPC_LABEL_NONCE_GENERATION;
 use crate::metrics::Metrics;
+use crate::mpc::types::AvidCertificate;
 use crate::mpc::types::AvidNonceMessage;
 use crate::mpc::types::AvidNonceMessageKind;
 use crate::mpc::types::AvidRoundState;
@@ -51,6 +52,8 @@ use crate::mpc::types::RotationReconstructionContext;
 pub use crate::mpc::types::SendMessagesRequest;
 pub use crate::mpc::types::SendMessagesResponse;
 pub use crate::mpc::types::SessionId;
+use crate::mpc::types::VerifiedAvidVoteCert;
+use crate::mpc::types::hash_avid_vote;
 use crate::onchain::types::CommitteeSet;
 use crate::storage::PublicMessagesStore;
 use fastcrypto::bls12381::min_pk::BLS12381Signature;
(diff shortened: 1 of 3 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2104,6 +2114,163 @@ impl MpcManager {
...
+        confirm_cert: DealerCertificate,
+        batch_index: u32,
+    ) -> MpcResult<Vec<(Address, Messages)>> {
+        let dealer_sid = SessionId::nonce_dealer_session_id(
+            &self.chain_id,
+            self.mpc_config.epoch,
+            batch_index,
+            &self.address,
+        );
+        let nodes = self.maybe_corrupt_nodes_for_testing(&self.mpc_config.nodes);
```
benr-ml, line 2130: for_testing - do we need that?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2104,6 +2114,163 @@ impl MpcManager {
...
+        dispersal: batch_avss_avid::Dispersal,
+        confirm_cert: DealerCertificate,
+    ) -> MpcResult<AvidEchoAndVote> {
+        let receiver = self.create_avid_nonce_receiver(dealer, batch_index)?;
+        let verified_common = receiver
+            .verify_common_message(common)
+            .map_err(|e| MpcError::CryptoError(e.to_string()))?;
+        let avid_confirm =
+            AvidCertificate::confirm(confirm_cert, Arc::new(self.committee.clone()))?;
+        let is_confirmer = avid_confirm.signers().contains(&self.party_id);
```
benr-ml, line 2195: i cannot follow from this what avid_confirm means - are the signers are the same one from confirm_cert?
what exactly is checked below for voting (L2192)?

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: This re-creates the dealer that minted the builder, so its inputs must match the optimistic-phase call exactly — including this hook. The hook is the existing vanilla nonce generation fault-injection (no-op in production), which the AVID complaint tests will reuse.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: avid_confirm is the same confirm_cert, adapted to fastcrypto's Certificate trait — same signers, payload = the certified H(v); process_avid_message verifies it in-band. The check below gates the vote on actually holding this round's decrypted output: since the cert's signatures are verified, a listed confirmer really did confirm — the case this catches is a confirmer that lost its output (e.g. restart) and must not add weight toward consuming a round it can't produce its share for. Non-confirmers vote on the dispersal itself (the vote attests the dispersal layer only) and decode later.

benr-ml (approved):

PR #748, opened 2026-07-06
btc: Standardize lifecycle timestamps on created_timestamp_ms

Bridgerz:
## Motivation

Per the review thread: `created` matches the other lifecycle fields (`signed`, `confirmed`), so this PR now standardizes **both** request structs on `created_timestamp_ms` — including flipping the deposit side that #747 merged as `creation_timestamp_ms`.

## Changes

- `DepositRequest.creation_timestamp_ms` → `created_timestamp_ms` (flips #747's naming)
- `WithdrawalRequest.timestamp_ms` → `created_timestamp_ms`
- `WithdrawalTransaction.timestamp_ms` → `created_timestamp_ms`
- Accessors follow (`request_created_timestamp_ms`)
- Rust mirrors and all consumers renamed; **no BCS layout change** (renames only), event and proto field names unchanged
- Also fixes the `#[cfg(test)]` `make_txn` helper that broke CI (clippy builds test targets; local check missed `--all-targets`)

## Test plan

- [x] 141/141 Move tests pass

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -438,7 +438,7 @@ pub struct WithdrawalRequest {
     pub sender: Address,
     pub btc_amount: u64,
     pub bitcoin_address: Vec<u8>,
-    pub timestamp_ms: u64,
+    pub created_timestamp_ms: u64,
     pub status: WithdrawalStatus,
     pub withdrawal_txn_id: Option<Address>,
     pub sui_tx_digest: Digest,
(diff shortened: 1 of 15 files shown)


crates/hashi-types/src/move_types/mod.rs
```diff
@@ -438,7 +438,7 @@ pub struct WithdrawalRequest {
     pub sender: Address,
     pub btc_amount: u64,
     pub bitcoin_address: Vec<u8>,
-    pub timestamp_ms: u64,
+    pub creation_timestamp_ms: u64,
```
bmwill: creation vs created, which do we want to align on?

Bridgerz: good catch ty

bmwill: i think `created` seems to match better with the other fields `signed` `confirmed` etc

Bridgerz:
Rebased onto main and standardized on `created_timestamp_ms` per the thread — including flipping `DepositRequest` (merged in #747 as `creation_`). Also fixed the test-target compile error that failed CI (`make_txn` in `withdrawals.rs` tests; local check had missed `--all-targets`).

bmwill (approved):

PR #749, opened 2026-07-06
withdraw: Record signed and confirmed timestamps on WithdrawalTransaction

Bridgerz:
## Motivation

Stacked on #748. Completes the `WithdrawalTransaction` lifecycle timestamps (`created` / `signed` / `confirmed`). The added fields change the struct's BCS layout, so they must land before mainnet freezes it.

## Changes

**Move:**
- New `WithdrawalTransaction.signed_timestamp_ms: Option<u64>` — set in `finalize_withdrawal_txn` alongside the guardian signatures
- New `WithdrawalTransaction.confirmed_timestamp_ms: Option<u64>` — set by the new `mark_confirmed` in `confirm_withdrawal` before the txn moves to the confirmed bag
- `finalize_withdrawal` and `confirm_withdrawal` entry functions gain a `&Clock` parameter (entry-only, signature change is free)

**Rust:**
- `hashi-types` `WithdrawalTransaction` mirror gains the two `Option<u64>` fields (matching Move BCS layout)
- `execute_finalize_withdrawal` / `execute_confirm_withdrawal` PTB builders pass the Clock object (0x6, immutable)

## Test plan

- [x] 141/141 Move tests pass
- [x] `cargo check --workspace` clean
- [x] prettier-move + cargo fmt clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #755, opened 2026-07-06
withdraw: Store the approval certificate and timestamp on WithdrawalRequest

Bridgerz:
## Motivation

Stacked on #749 (last PR of the B stack). `approve_request` verifies the committee cert over `RequestApprovalMessage` and then drops it — no on-chain record of which committee approved a withdrawal or when. This persists both, mirroring the deposit flow. Field additions change the BCS layout, so pre-mainnet only.

## Changes

**Move:**
- `WithdrawalRequest.approval_cert: Option<CommitteeSignature>` + `approved_timestamp_ms: Option<u64>`, set by `approve_withdrawal`; re-approval overwrites
- `approve_request` gains a `&Clock` parameter (entry-only signature)
- `DepositRequest.approval_timestamp_ms` → `approved_timestamp_ms` so all lifecycle timestamps follow one convention (`created`/`approved`/`signed`/`confirmed`)
- Accessors follow; event field names unchanged (`DepositApprovedEvent.approval_timestamp_ms` stays — happy to align event fields separately if wanted)

**Rust:**
- Mirrors + consumers updated; `execute_approve_withdrawal_requests` PTB passes the Clock object

## Test plan

- [x] 141/141 Move tests pass

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -545,6 +545,12 @@ pub struct WithdrawalTransaction {
     /// `withdrawal_outputs.len() + j`. Empty when there is no change.
     pub change_outputs: Vec<OutputUtxo>,
     pub created_timestamp_ms: u64,
+    /// Clock timestamp at which the transaction became fully signed
+    /// (guardian signatures attached). `None` until `finalize_withdrawal`.
+    pub signed_timestamp_ms: Option<u64>,
+    /// Clock timestamp at which the Bitcoin transaction was confirmed.
+    /// `None` until `confirm_withdrawal`.
+    pub confirmed_timestamp_ms: Option<u64>,
     pub randomness: Vec<u8>,
     /// Per-input MPC signatures, accumulated incrementally and out-of-order.
     pub signing: SigningBatch,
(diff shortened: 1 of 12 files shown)


crates/hashi-types/src/move_types/mod.rs
```diff
@@ -440,6 +440,9 @@ pub struct WithdrawalRequest {
     pub bitcoin_address: Vec<u8>,
     pub created_timestamp_ms: u64,
     pub status: WithdrawalStatus,
+    /// Committee certificate recorded at approval time. `None` until
+    /// `approve_request` has been called.
+    pub approval_cert: Option<CommitteeSignature>,
```
bmwill, line 446: do we want an `approved_timestamp_ms` here as well?

Bridgerz:
Added `approved_timestamp_ms` alongside the cert (set in `approve_withdrawal`; `approve_request` now takes `&Clock`). Also renamed the deposit side's `approval_timestamp_ms` → `approved_timestamp_ms` in the same pass so lifecycle timestamps follow one convention: `created` / `approved` / `signed` / `confirmed`. Event field names left untouched as before.

bmwill (approved):

PR #756, opened 2026-07-06
hashi: Domain-separate committee-signed messages with intent tags

Bridgerz:
## Motivation

Every committee certificate is a BLS signature over `bcs(epoch) || bcs(message)` — the signed bytes record *no* message type, so if two message types share a BCS layout, a cert for one verifies as the other. This is already latent on `main`: `RequestApprovalMessage { request_id: address }` and `WithdrawalConfirmationMessage { withdrawal_id: address }` are both a single 32-byte address and are byte-identical. Only Sui object-id global uniqueness prevents cross-flow cert replay today — an accident of content, not a designed guarantee.

Adding chains multiplies the signed-message set and introduces non-unique 32-byte values (BTC txids already ride in `address` fields), so this must be fixed before the wire format freezes at mainnet. Post-launch it becomes a flag-day migration of the on-chain verifier and every validator's signer in lockstep.

## How it works

The preimage becomes `bcs(epoch) || intent (1 byte) || bcs(message)`. Each signed message type gets a unique intent byte from a mirrored registry (`hashi::intent` / `hashi-types::intent`; ranges: `0..=63` core, `64..=127` bitcoin, `128..` future chains). Two types with identical payloads now produce different signed bytes, so cross-type replay is structurally impossible — the same pattern as Sui's own `IntentScope`.

- **On-chain:** `verify_certificate` takes an `intent: u8`, threaded through `verify` / `verify_with_committee`; each of the 8 verify sites names its constant. The proof-of-possession preimage gets its own tag too.
- **Off-chain:** an `IntentMessage { const INTENT: u8 }` trait is required by the sign/verify helpers, so the compiler forces every signed type to declare a domain and the two sides cannot disagree.

## Test plan

- [x] 141/141 Move tests pass

### DIFF
--- a/crates/hashi-types/src/intent.rs
+++ b/crates/hashi-types/src/intent.rs
@@ -3,51 +3,101 @@
 
 //! Domain-separation intents for everything signed by Hashi member BLS keys.
 //!
-//! The signing preimage is `bcs(epoch) || intent || bcs(message)`. Every
-//! message type signed under the committee's keys carries a unique intent
-//! byte, so a certificate produced for one message type can never verify as
-//! another — regardless of whether two types happen to share a BCS layout.
+//! The signing preimage is `intent (u16 LE) || bcs(epoch) || bcs(message)`.
+//! Every message type signed under the committee's keys carries a unique
+//! intent value, so a certificate produced for one message type can never
+//! verify as another — regardless of whether two types happen to share a BCS
+//! layout.
+//!
+//! [`Intent`] is a `#[repr(u16)]` enum with explicit discriminants: the
+//! compiler rejects two variants sharing a value, and each message type names
+//! a variant (not a raw number), so a typo'd or reused intent fails to compile.
+//! The wire value is the discriminant, taken via `as u16` — never serde's enum
+//! encoding (which is the variant index, not the discriminant).
 //!
 //! This registry mirrors `packages/hashi/sources/core/intent.move`; the two
-//! MUST stay in sync. Allocation ranges:
-//! - `0..=63`: core protocol (committee lifecycle, MPC ceremonies)
-//! - `64..=127`: Bitcoin
-//! - `128..`: reserved for future chains
+//! MUST stay in sync — `intent_values_are_stable` below pins the on-wire
+//! values so the Rust side cannot drift silently, and the e2e suite is the
+//! true cross-language check. Allocation blocks, one per domain, with room to
+//! grow within each:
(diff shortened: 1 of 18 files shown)


crates/hashi-types/src/intent.rs
```diff
@@ -0,0 +1,53 @@
...
+/// Proof of possession of a member BLS key (epoch, address, public key).
+pub const PROOF_OF_POSSESSION: u8 = 0;
+/// Committee handoff: the outgoing committee attests the next committee.
+pub const COMMITTEE_TRANSITION: u8 = 1;
+/// Reconfiguration completion: the next committee holds the MPC key(s).
+pub const RECONFIG_COMPLETION: u8 = 2;
+/// AVID/TOB dealer-messages hash certificates (MPC ceremonies).
+pub const DEALER_MESSAGES_HASH: u8 = 3;
+
+// ==== Bitcoin (64..=127) ====
```
bmwill: should we make these a `u16` to give us more room to grow?

bmwill: I think in Rust we can make these an enum and `repr(u16)` it and assign the exact numbers we want to each variant, allowing the compiler to help catch overlap.

packages/hashi/sources/core/committee/committee.move
```diff
@@ -212,8 +213,11 @@ public(package) fun verify_certificate<T>(
     // Verify the signature
     let pub_key_bytes = group_ops::bytes(&aggregate_key);
 
-    // Signing message is always prefixed with the epoch
+    // Signing preimage: bcs(epoch) || intent || bcs(message). The intent
+    // byte domain-separates message types signed under the same keys (see
+    // hashi::intent).
```
bmwill: Can we switch this slightly so that the intent comes first?

Bridgerz:
Addressed all three: intent is now a `#[repr(u16)]` enum with explicit discriminants (compiler catches value collisions, mirrors the guardian `IntentType` pattern), widened to u16 with per-domain blocks (core `0x00xx`, bitcoin `0x01xx`, future chains `0x02xx+`), and the preimage now leads with the intent — `intent (u16 LE) || bcs(epoch) || bcs(message)` — across the verify path, PoP preimage, off-chain signer, and test helpers. One subtlety worth noting: the wire value is the discriminant via `as u16`, not serde's enum encoding (which would serialize the variant *index*, not the value). Added `intent_values_are_stable` + a preimage-layout test to lock the Rust side; the values still have to match `intent.move` by hand (e2e is the true cross-language check).

bmwill (approved):

PR #761, opened 2026-07-06
deposit: Never garbage-collect an approved deposit request

Bridgerz:
## Motivation

`delete_expired` (`btc/deposit_queue.move`) removed any active request older than `MAX_DEPOSIT_REQUEST_AGE_MS` based purely on the request timestamp, with no check of approval state — and `delete_expired_deposit` exposes that path **permissionlessly**.

Once `approve_deposit` records a committee certificate, the request is awaiting `confirm_deposit`, which can be transiently blocked (pause / reconfiguration) or gated behind the configured time delay. So anyone could delete a valid, mint-pending approved deposit at the expiry boundary — destroying the stored certificate and forcing the user to resubmit and get re-approved. Low severity (recoverable, and the timing window is narrow) but a real griefing vector.

## Fix

Reject deletion of approved requests. An approval is a committee commitment, so a request carrying an `approval_cert` is never expirable — only never-approved (genuinely abandoned) requests are collectable.

- **On-chain:** `delete_expired` asserts `approval_cert.is_none()` (`ECannotDeleteApprovedDeposit`).
- **Off-chain:** the leader's GC scan excludes approved requests, so it doesn't propose deletions that would now abort.

**Why not "expire relative to approval time" instead:** that still leaves a hole when the confirmation delay exceeds the request age (the request would become deletable before it's confirmable). Rejecting approved requests outright closes it completely.

**Why this can't leak storage:** approval requires a committee-verified real Bitcoin deposit, so the set of approved-but-unconfirmed requests is bounded by genuine deposit activity — an attacker cannot fabricate approved entries to pin storage. Abandoned approved requests (e.g. cert went stale across a committee rotation and was never re-approved) are rare and small; if that ever needs cleanup, it can be an authorized action rather than the permissionless GC.

## Test plan

- [x] New Move test `test_delete_approved_deposit_request_fails`: an approved, expired request aborts on delete with `ECannotDeleteApprovedDeposit`
- [x] 142/142 Move tests pass
- [x] `cargo clippy -p hashi --all-features` clean; prettier + fmt clean

### DIFF
--- a/crates/hashi/src/leader/garbage_collection.rs
+++ b/crates/hashi/src/leader/garbage_collection.rs
@@ -45,6 +45,10 @@ impl LeaderService {
         }
 
         let mut deposit_requests = self.inner.onchain_state().deposit_requests();
+        // Approved requests carry a committee certificate and are awaiting
+        // confirmation; the on-chain delete path rejects them, so exclude them
+        // from GC entirely rather than proposing deletions that would abort.
+        deposit_requests.retain(|r| r.approval_cert.is_none());
         deposit_requests.sort_by_key(|r| r.created_timestamp_ms);
 
         let Some(oldest_request) = deposit_requests.first() else {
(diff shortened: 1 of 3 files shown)


bmwill:
The chances of this happening are really small and a user can always resubmit so i think we should actually drop this as it just adds additional complexity to the code that exists. Thoughts?

PR #762, opened 2026-07-06
utxo_pool: Rename UtxoRecord.locked_by to spent_by

Bridgerz:
## Motivation

From the audit-notes cleanup list. Renames `UtxoRecord.locked_by` → `spent_by` so it pairs with the sibling `produced_by` for a clean producer/consumer reading: `produced_by` is the withdrawal that created a (change) UTXO, `spent_by` is the withdrawal that consumes it. Since a locked UTXO is never unlocked, the reserving withdrawal is the one that ends up spending it.

Pure field rename — BCS-compatible, and the field is internal live bookkeeping that never appears in a retained/historical object (a spent `UtxoRecord` is destroyed at `cleanup_spent`, leaving only a compact `spent_utxos` marker).

## Scope decisions (called out for review)

- **`spent_epoch` intentionally left unchanged.** The original note suggested `spent_epoch → confirmed_epoch`, but `spent_epoch` is consistent across the record field, the `spent_utxos` bag, and `UtxoSpentEvent`, and `confirmed_epoch` reads ambiguously (it sounds like the epoch the UTXO was *confirmed as a deposit/change*, the other end of its life). Renaming just the record field would desync the vocabulary. Happy to do the full cascade if preferred.
- **`lock()` / `EUtxoAlreadyLocked` kept.** The verb still describes the reservation action, but note the minor asymmetry it introduces: `lock()` now assigns `record.spent_by`. The struct-level `produced_by`/`spent_by` symmetry is worth more than that mutation-site oddity IMO, but flagging it — easy to revert to `locked_by` if you'd rather keep `lock` and the field name aligned.

## Changes

- Move: `UtxoRecord.locked_by` → `spent_by` + doc comments.
- Rust: `hashi-types` mirror and all consumers (leader, watcher, metrics, coin selection, e2e).

## Test plan

- [x] 141/141 Move tests pass
- [x] `cargo check --workspace --all-targets` clean; `clippy --all-features` clean; prettier + fmt clean

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -634,7 +634,7 @@ pub struct Utxo {
 pub struct UtxoRecord {
     pub utxo: Utxo,
     pub produced_by: Option<Address>,
-    pub locked_by: Option<Address>,
+    pub spent_by: Option<Address>,
     pub spent_epoch: Option<u64>,
 }
 
--- a/crates/e2e-tests/src/e2e_flow.rs
+++ b/crates/e2e-tests/src/e2e_flow.rs
@@ -954,7 +954,7 @@ mod tests {
 
         // Wait for the committee to commit withdrawal 1. At this point the
         // deposit UTXO is locked and the change UTXO is inserted as pending
-        // (produced_by = Some, locked_by = None). No Bitcoin blocks have been
+        // (produced_by = Some, spent_by = None). No Bitcoin blocks have been
         // mined, so neither the deposit spend nor the change output is
         // confirmed on-chain.
         let picked1 =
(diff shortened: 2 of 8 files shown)


crates/hashi-types/src/move_types/mod.rs
```diff
@@ -634,7 +634,7 @@ pub struct Utxo {
 pub struct UtxoRecord {
     pub utxo: Utxo,
     pub produced_by: Option<Address>,
-    pub locked_by: Option<Address>,
+    pub spent_by: Option<Address>,
     pub spent_epoch: Option<u64>,
```
bmwill, line 638: did we decide to make this one `confirmed_epoch` or something like that?

bmwill (approved):

crates/hashi-types/src/move_types/mod.rs
Bridgerz: I think spent_epoch makes the most sense because it records when a UTXO was spent.

bmwill: Does it record the epoch in which the txn was constructed or confirmed?

PR #765, opened 2026-07-06
[1/4] guardian: add standby init config lifecycle core

mskd12:
## Summary
- add InitConfig and ActivationState typing/hashing for standby activation
- add guardian OperatorActivate RPC plumbing through guardian/proxy/init paths
- keep reader-backed activation safety checks as explicit TODO placeholders for later stack PRs

## Notes
- this PR intentionally does not add genesis bootstrap APIs/logs
- activation still uses the existing committee-update reader path and genesis limiter state placeholder; later PRs replace those with genesis fallback plus live S3 activation checks

## Checks
- cargo fmt -- --config imports_granularity=Item --config format_code_in_doc_comments=true --check
- cargo run -p proto-build
- git diff --check


crates/hashi-guardian/src/operator_activate.rs
```diff
@@ -0,0 +1,122 @@
...
+    let activation_state = ActivationState::new(
+        config_hash,
+        armed_instance,
+        committee.clone(),
+        limiter_state,
+    )?;
+    let state_hash = activation_state.digest();
+    if &state_hash != request.expected_state_hash() {
+        return Err(InvalidInputs(format!(
+            "ActivationState hash mismatch: expected {:?}, got {:?}",
```
dlukegordon: Can we use `hex::encode` here, to match error messages elsewhere.

crates/hashi-guardian/src/operator_activate.rs
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) Mysten Labs, Inc.
```
dlukegordon, line 1: Can we get some test coverage for this file?

crates/hashi-types/src/guardian/mod.rs
```diff
@@ -410,114 +414,137 @@ impl OperatorInitRequest {
...
-impl WithdrawModeState {
+impl ActivationState {
     pub fn new(
+        config_hash: [u8; 32],
+        secret_sharing_instance: SecretSharingInstance,
         committee: HashiCommittee,
-        limiter_config: LimiterConfig,
         limiter_state: LimiterState,
-        hashi_btc_master_pubkey: HashiMasterG,
     ) -> GuardianResult<Self> {
```
dlukegordon: I think we can drop `GuardianResult` now, and just return `Self`

crates/hashi-types/src/guardian/mod.rs
```diff
@@ -410,114 +414,137 @@ impl OperatorInitRequest {
...
+        Self {
+            expected_state_hash,
+        }
+    }
+
+    pub fn expected_state_hash(&self) -> &[u8; 32] {
+        &self.expected_state_hash
+    }
+
+    pub fn into_parts(self) -> [u8; 32] {
```
dlukegordon: I think we can drop this method, `into_parts` seems like a strange name for this, since it just returns a single array field, and we already have `expected_state_hash` above.

crates/hashi-types/src/guardian/test_utils.rs
```diff
@@ -216,23 +215,21 @@ fn mock_committee_with_one_member(epoch: u64) -> HashiCommittee {
...
 }
 
-impl WithdrawModeConfig {
+impl InitConfig {
     pub fn from_parts_for_testing(
         limiter_config: LimiterConfig,
-        limiter_state: LimiterState,
-        committee: HashiCommittee,
+        _limiter_state: LimiterState,
+        _committee: HashiCommittee,
```
dlukegordon: Should we remove these two fields?

crates/hashi-guardian/src/operator_activate.rs
mskd12: Hmm, I am not sure how to write meaningful tests for this file as it calls complex s3 functions underneath.

dlukegordon: Ok all good then

dlukegordon (approved):

PR #769, opened 2026-07-06
[3/4] guardian: activate standby from live S3 state

mskd12:
## Summary
- move heartbeat activation checks and limiter recovery into `GuardianReader`
- have `operator_activate` require the pinned standby session to have a recent heartbeat and reject activation when any other session has a recent heartbeat
- recover activation limiter state from the latest successful withdrawal logs, falling back to genesis only when no success exists
- cap recovered limiter tokens to the new limiter config capacity before constructing `RateLimiter`
- remove the now-unused `hashi-guardian-init` heartbeat/limiter helper modules

## Notes
- This PR does not add the `hashi-guardian-init operator activate` CLI; that stays in the next PR.
- `operator_activate` still verifies latest ceremony instance, latest committee/genesis state, and the operator-pinned `ActivationState` hash.

## Verification
- `cargo fmt -- --config imports_granularity=Item --config format_code_in_doc_comments=true --check`
- `git diff --check`
- `cargo test -p hashi-guardian s3_reader::heartbeat_checks`

Full compile/test coverage is left to CI for this stack.


crates/hashi-guardian/src/s3_reader/heartbeat_checks.rs
```diff
@@ -0,0 +1,266 @@
...
+            session_id = %live_session,
+            first_heartbeat = live_session_info.first_heartbeat,
+            last_heartbeat = live_session_info.last_heartbeat,
+            age_secs = now.saturating_sub(live_session_info.last_heartbeat),
+            "activation heartbeat check passed"
+        );
+        Ok(())
+    }
+
+    async fn read_recent_heartbeats(&mut self) -> anyhow::Result<Vec<VerifiedLogRecord>> {
```
dlukegordon, line 58: To account for possible clock skew, we could do this

```
let one_hour_ago = now_timestamp_secs().saturating_sub(60 * 60);
let mut cursor = heartbeat_cursor(one_hour_ago);
let mut logs = Vec::new();
for _ in 0..3 {
    logs.extend(self.read_dir(&cursor).await?);
    cursor = cursor.next_dir();
}
```

crates/hashi-guardian/src/s3_reader/limiter_recovery.rs
```diff
@@ -1,69 +1,76 @@
...
-    // a higher-seq log in the prior hour bucket.
-    let hit = bucket_max_post_state(reader.read_dir(&cursor).await?);
-    cursor = cursor.prev_dir();
-    let peek = bucket_max_post_state(reader.read_dir(&cursor).await?);
-    let best = [hit, peek].into_iter().flatten().max_by_key(|s| s.next_seq);
+impl GuardianReader {
+    /// Derive the activation limiter state from withdrawal logs. Uses the
+    /// global max-seq Success post-state when present, otherwise genesis, and
+    /// caps tokens to the supplied config in case capacity was lowered.
+    pub async fn recover_limiter_state(
```
dlukegordon, line 47: Maybe add

```
        /// Precondition: the caller must have already verified that every
        /// non-standby session is quiet (`ensure_session_live_and_others_quiet`).
        /// This read deliberately skips the `write_completion_time` gate, so it
        /// is only sound once the prior session's final writes are guaranteed
        /// visible.
```

dlukegordon (approved):

crates/hashi-guardian/src/s3_reader/heartbeat_checks.rs
mskd12: thanks! nice catch :-)

PR #771, opened 2026-07-07
[guardian-proxy] Back the wid cache with the guardian's S3 withdrawal log

0xsiddharthks:
## Changes

- For the in-process LRU cache on the proxy:
  - on a miss, scan `withdraw/` for the wid's success record (newest-first hour buckets, bounded by the request's seq, capped LIST count).
  - Replay a record only if the enclave actually committed it (its seq is below `next_seq` from GetGuardianInfo) and its signatures verify against the guardian BTC key over sighashes recomputed from the request.
  - Fail closed with a retriable `UNAVAILABLE` whenever the log can't answer (S3 error, over-cap scan, guardian info unreachable, failed verification) — never forward blind.
- Replayed responses carry a zeroed envelope signature: the record predates the envelope, and nodes don't verify it.
- Per-outcome cache metrics on a new `/metrics` endpoint.
- New proxy env: `GUARDIAN_LOG_BUCKET` / `GUARDIAN_LOG_REGION` / `BTC_NETWORK`, creds via the AWS default chain. 
- Local replica: the proxy reads the compose MinIO directly; the redis sidecar is gone.

### DIFF
--- a/crates/hashi-guardian-proxy/src/cache.rs
+++ b/crates/hashi-guardian-proxy/src/cache.rs
@@ -14,11 +14,15 @@
 //! immutable (only signatures change), so the response is stable.
 //!
 //! The enclave persists every signed withdrawal before releasing it, so the
-//! proxy has nothing durable of its own to lose. A log replay passes two gates
-//! (see `replay_from_log`); when the log can't answer, the proxy fails closed
-//! with `UNAVAILABLE` rather than forwarding blind — it cannot distinguish
-//! "never signed" from "signed but unreadable", and forwarding the latter
-//! re-signs and double-debits the limiter.
+//! proxy has nothing durable of its own to lose. On a log hit it verifies the
+//! record's signatures spend the incoming request, then replays them (see
+//! `replay_from_log`); it does not reason about the limiter `seq`, which is the
+//! guardian's to own — the node's mirror self-heals any divergence
+//! (`hashi/src/guardian_limiter.rs`). When the log can't answer, the proxy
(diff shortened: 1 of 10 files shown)


dlukegordon (approved):
Generally LGTM, can we improve metrics though, so we can track cache hits/misses?

crates/hashi-guardian-proxy/src/cache.rs
```diff
@@ -89,6 +138,162 @@ impl<S> CachingGuardianGrpc<S> {
...
+            Err(WidLogError::Store(e)) => {
+                self.metrics.outcome(metrics::OUTCOME_UNAVAILABLE_LOG_STORE);
+                error!(%wid, requested_seq, error = %e, "Wid log unavailable; refusing to forward blind.");
+                return Err(unavailable());
+            }
+        };
+
+        // Gate 1 (committed): a record whose S3 ack never reached the enclave
+        // was reverted, not debited — `seq >= next_seq` — and must re-sign.
+        let info = match self.guardian_info().await {
```
mskd12, line 183: this check would fail if the request is for an old/stale record, no? maybe just do gate 2 and leave gate 1 out?

mskd12 (approved):
LGTM: one small change suggested.

crates/hashi-guardian-proxy/src/cache.rs
0xsiddharthks: Makes sense. the proxy shouldn't be reasoning about seq
should just do idempotency by wid, and the node's mirror self-heals any drift via the stall reconcile.

PR #772, opened 2026-07-07
reconfig: Gate genesis on upgrade-cap registration instead of 95% stake

Bridgerz:
## What

Replaces the automatic genesis trigger (any validator's `start_reconfig` succeeding once **≥95% of total Sui stake** has registered) with an explicit **launch switch** controlled by the publisher — and per review, the launch switch is `finish_publish` itself:

1. `hashi publish` now runs the publish transaction only. The `UpgradeCap` stays in the publisher's wallet and the deploy is unconfigured (init defaults only).
2. Validators register and boot as usual; nodes register their next-epoch keys on startup and idle-retry `start_reconfig` (fails at simulation, gas-free).
3. When the roster is ready, the publisher sends `hashi::finish_publish` (unchanged from main: consumes and custodies the cap, sets chain id / guardian / overrides, creates the BTC currency). The genesis branch of `start_reconfig` requires the custodied cap (`EGenesisNotAuthorized`), so this one cap-gated call configures the deploy **and** starts genesis. `GENESIS_STAKE_THRESHOLD_BPS` / `genesis_stake_satisfied` are deleted; post-genesis reconfiguration is unchanged.

Because configure and launch are one call, there is no ordering between them to get wrong, and the guardian parameters are only needed at launch time (the ceremony can run during the registration window).

## Why

Requiring 95% of all Sui validator stake to opt in before the bridge can launch is operationally unrealistic and gives no control over the launch moment. The cap holder already has strictly more power (upgrade authority), so gating launch on the same key adds no new trust assumption.

## Node software: two contained pre-launch tolerances

Nodes must boot before the launch (the launch waits for keys that only booted nodes register), which puts them in a window where the on-chain config has only init defaults:

- `verify_bitcoin_chain_id` skips with a warning while the on-chain value is absent; any restart after launch re-verifies.
- The guardian client resolves lazily from on-chain config on first use when no endpoint was available at startup (all consumers already handle an absent guardian).

Every e2e run exercises this pre-launch boot window. The genesis retry loop itself is untouched (failing attempts die at simulation — no gas).

## Harness changes

`publish_and_init` splits into `publish_package` + `finish_publish` (with a `build_finish_publish_tx` variant for offline/multisig signing). The harness launches via `HashiNetwork::launch_genesis`: it waits until every initially-active validator is **fully key-registered**, asserts genesis has not started (the gate held), then sends `finish_publish`. Manual localnet mode uses the same path after the operator registers and starts the validators.

## Tests

- Genesis-gate Move tests in `tests/reconfig_tests.move` (gate requires cap / passes with cap / skipped post-bootstrap); 143 Move tests pass.
- The gate-held assertion runs inside every e2e network build, so every e2e test verifies genesis cannot start before the launch.
- `tests::test_dkg`, `upgrade_tests::test_upgrade_v1_to_v2` (cap custody via the launch still authorizes governance upgrades), and `e2e_flow::tests::test_bitcoin_deposit_e2e_flow` (launch-time guardian/chain-id config works end-to-end) pass locally, plus 416/417 hashi unit tests (the one failure needs a local `gpg` binary and is unrelated).

Stacked PR #773 adds the operator-facing `hashi launch` CLI command and docs.


bmwill:
Let's get @zhouwfang to also take a look

packages/hashi/sources/hashi.move
```diff
@@ -102,15 +102,17 @@ public(package) fun assert_not_reconfiguring(self: &Hashi) {
...
+// some input parameters and register BTC.
 //
 // The guardian URL and BTC key are both required: the guardian is a
 // load-bearing component of every deposit address (2-of-2 taproot leaf)
 // and every withdrawal signature, so a deploy without them would produce
 // a non-functional bridge.
+//
+// The package's `UpgradeCap` stays with the publisher; handing it in later
+// via `register_upgrade_cap` is the launch switch that unlocks genesis.
 entry fun finish_publish(
```
zhouwfang, line 117: Dropping the UpgradeCap param also dropped finish_publish's only access control: anyone can now call it first and set the immutable guardian_btc_public_key / guardian_url / chain id, and it can't be atomic with publish (the Hashi object is shared), so the window is real. Suggest cap: `&sui::package::UpgradeCap` with the same `package() == original_id` assert.

crates/hashi/src/cli/mod.rs
```diff
@@ -1226,6 +1229,11 @@ pub async fn run_publish(opts: PublishOpts) -> anyhow::Result<()> {
     .await?;
     print_success(&format!("package_id:      {}", ids.package_id));
     print_success(&format!("hashi_object_id: {}", ids.hashi_object_id));
+    print_success(&format!("upgrade_cap_id:  {upgrade_cap_id}"));
+    print_info(
+        "The UpgradeCap stays in the publisher's wallet. Once all expected validators \
+         have registered, send `hashi::register_upgrade_cap` to unlock genesis.",
```
zhouwfang: CLI says to send `hashi::register_upgrade_cap` but offers no subcommand; a wrapper avoids a hand-crafted mainnet-launch tx. Intentional?

zhouwfang (changes requested):

packages/hashi/sources/hashi.move
Bridgerz: great catch

crates/hashi/src/cli/mod.rs
Bridgerz: yes, reference #773

packages/hashi/sources/hashi.move
bmwill: Can we not leave this as-is and just call "finish_publish" to trigger allowing genesis?

packages/hashi/sources/hashi.move
```diff
@@ -154,6 +159,24 @@ entry fun finish_publish(
     self.treasury.register_metadata_cap(metadata_cap);
 }
 
+/// The launch switch. Hands the package's `UpgradeCap` into on-chain custody
+/// (`Versioning`). Validators can register beforehand, but
+/// `reconfig::start_reconfig` refuses to form the initial committee until the
+/// cap is registered. `Option::fill` inside `set_upgrade_cap` makes this
+/// call-once.
+entry fun register_upgrade_cap(self: &mut Hashi, upgrade_cap: sui::package::UpgradeCap) {
```
zhouwfang: This has no on-chain guard that finish_publish ran. The CLI bundles finish_publish into publish so a normal launch can't precede it, but nothing enforces it on-chain, and #773's manual/multisig launch path makes a hand-ordered mistake reachable. A cheap assert here (e.g. guardian key set) would enforce it.

packages/hashi/sources/hashi.move
Bridgerz: wait ya this is better. on it

packages/hashi/sources/hashi.move
Bridgerz: Going with Brandon's suggestion above (finish_publish is the launch), which removes the ordering entirely.

packages/hashi/sources/hashi.move
zhouwfang: One consideration: servers self-register on boot behind verify_bitcoin_chain_id, which fails until finish_publish sets the on-chain bitcoin_chain_id. So with the split, operators can boot anytime and launch late; one-step would mean booting only after launch (or loosening that check). May be a reason to lean toward the split.

Bridgerz: I think we go with loosening. Boot-after-launch is circular since launch waits for keys that only booted nodes register. verify_bitcoin_chain_id now skips while the value is absent (restarts re-verify) and the guardian client resolves lazily from on-chain config; also every e2e run boots nodes pre-launch, so the window is covered.

bmwill: agreed, we can also do an explicit verify of the chain id as a part of reconfig if we want to as well (because it would be set at that point)

bmwill (approved):
lgtm, do note that once we land this we'll also need to fix our deployment workflows for devnet (and maybe private testnet) to take into account this change.

Then we'll likely want to do a fresh deploy of devnet with this (and any other move changes we need to make) to verify things work as expected then we should be good to cut a release.

packages/hashi/sources/hashi.move
Bridgerz: done

zhouwfang (approved):
nit: finish_publish now carries register_upgrade_cap's guards — wrong-cap (EWrongUpgradeCap) and call-once, but their tests were dropped with it. Worth small test.

0xsiddharthks (approved):

PR #775, opened 2026-07-08
config: Add U128 and U256 variants to the config Value enum

Bridgerz:
## What

Appends `U128(u128)` and `U256(u256)` variants to `config_value::Value`, with the matching constructor/predicate/extractor functions and `same_variant` arms. The Rust mirror (`hashi_types::move_types::ConfigValue`) gains the same variants — u256 represented as its 32 little-endian BCS bytes, which a fixed-size `[u8; 32]` encodes byte-identically — and the proposal transaction builder maps the new variants to their Move constructors. A pinned BCS round-trip test covers the new tags (U128 = 5, U256 = 6).

No config keys use the new variants yet; this is purely reserving representational space.

## Why now

Pre-mainnet hardening from the native-bridge→Hashi asset-migration readiness analysis: Sui's compatible-upgrade policy permits adding modules and functions but **never adding enum variants** (`move-binary-format` rejects any upgrade where an enum has more variants than before). Once we publish to mainnet, `Value` is frozen forever.

The concrete future need is wide-integer config values: ETH-era parameters are wei-denominated, and u64 overflows at ~18.4 ETH — too small for values like daily limits. Without these variants, post-mainnet ETH config would be stuck encoding amounts as `Bytes` or off-unit denominations permanently. Adding them now costs nothing.

## Notes

- Variants are appended at the end — BCS decodes by variant index, so existing on-chain values and the Rust mirror are unaffected.
- CLI `update-config` string parsing (`u64:`/`bool:` prefixes) deliberately does not gain `u128:`/`u256:` yet — that's Rust-side and can be added whenever the first wide-integer key exists.

## Testing

- Move: `config_tests::test_config_value_wide_integers_round_trip` + `test_config_value_same_variant_distinguishes_integer_widths`; full suite 144/144.
- Rust: `move_types::tests::config_value_wide_integer_bcs_round_trip` pins the exact BCS bytes; hashi-types and hashi suites green.

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -216,6 +216,10 @@ pub enum ConfigValue {
     String(String),
     Bool(bool),
     Bytes(Vec<u8>),
+    U128(u128),
+    /// Move `u256`, as its 32 little-endian BCS bytes (no Rust u256 primitive;
+    /// a fixed-size array BCS-encodes to exactly the same 32 bytes).
+    U256([u8; 32]),
 }
 
 /// MPC parameter keys, in the canonical order Move's `mpc_config::pin` writes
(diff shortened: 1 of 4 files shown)


crates/hashi-types/src/move_types/mod.rs
```diff
@@ -216,6 +216,10 @@ pub enum ConfigValue {
     String(String),
     Bool(bool),
     Bytes(Vec<u8>),
+    U128(u128),
+    /// Move `u256`, as its 32 little-endian BCS bytes (no Rust u256 primitive;
+    /// a fixed-size array BCS-encodes to exactly the same 32 bytes).
+    U256([u8; 32]),
```
bmwill, line 218: Do you want to reorder these so they're next to the U64?

Bridgerz: yes lol getting the other PRs sorted first

bmwill (approved):

PR #778, opened 2026-07-08
[guardian-proxy] Serve /info + /health on the gRPC server's port

0xsiddharthks:
## Changes

- gRPC + `/info` + `/health` now share one `axum::serve` on `:3000` (each tonic service mounted as an axum route-service). Dropped `HTTP_LISTEN_ADDR` / `:3001`.
- Added a `single_port` test proving native gRPC and plain HTTP coexist on the one socket.
- `/info` JSON contract unchanged; Prometheus `/metrics` stays on its own internal port.

## Companion PRs

- MystenLabs/sui-operations#8350
- MystenLabs/hashi-ts-sdk#38

### DIFF
--- a/crates/hashi-guardian-proxy/src/info.rs
+++ b/crates/hashi-guardian-proxy/src/info.rs
@@ -1,0 +1,283 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Read-only HTTP surface for the proxy: `GET /info` (a curated JSON projection
+//! of the enclave's signed `GetGuardianInfo`) and `GET /health` (proxy
+//! liveness), with permissive CORS. It lets browser/`fetch` clients like the
+//! hashi-ts-sdk read the withdrawal rate-limiter state that the gRPC surface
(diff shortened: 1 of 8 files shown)


bmwill:
@0xsiddharthks I'm not sure i follow here. why do we need to add a new interface vs adding these routes to the existing one?

In particular we will only expose a single URL for the guardian and so everything we want to be publicly accessible from that needs to be on a single interface

crates/hashi-guardian-proxy/src/info.rs
```diff
@@ -0,0 +1,533 @@
...
+use tonic::transport::Channel;
+use tower_http::cors::Any;
+use tower_http::cors::CorsLayer;
+use tracing::error;
+
+/// Curated, read-only projection of `GuardianInfo` served at `GET /info`.
+/// Pubkeys are hex; `signed_at_ms` is a freshness signal.
+#[derive(Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct GuardianInfoView {
```
bmwill, line 48: This is ok for now, but proto does have a canonical json encoding that we should use given this data does come from a proto

bmwill (approved):

PR #781, opened 2026-07-09
[guardian-proxy] Source the relay KP roster from the S3 share log

0xsiddharthks:
## Summary

The relay's KP roster came from deploy config (`AUTHORIZED_KP_FINGERPRINTS`), so a ceremony or re-deal that changed who holds shares needed a matching proxy redeploy to stay in sync. The ceremony already commits that set: every encrypted share in the S3 log is labelled with its recipient's fingerprint, so the committed share log *is* the roster. Read it from there instead.

The read is a DoS-tier gate only — the enclave still verifies every share cryptographically — so it is deliberately unverified, and it fails closed when no share log exists.

## Changes

- New `roster` module: resolves the roster the way the enclave does, anchoring on the latest `ceremony/` record and reading the share state for its `sharing_seq`. Shares are published before the ceremony record, so taking the newest `kp-shares/` dir would pick up an orphan from a ceremony that never finished.
- Build both prefixes with hashi-types' key constructors, the same ones the enclave's reader uses, so the layout has a single definition. The tests still spell the keys out by hand and check that agreement.
- Accept both share-recipient shapes: testnet's deployed guardian writes `schema_version: 1` (`recipient_fingerprint`), current builds write `2` (`ciphertexts_by_fingerprint`). A rotation reads the old before writing the new, so both are live.
- `Relay` is generic over the log store and holds a TTL-cached roster; `S3LogStore` is now `Clone`, shared with the wid cache.
- The relay and the forwarder share one cache, so a successful `ProvisionerRotateCert` drops it — otherwise the replacement cert is rejected until the TTL lapses. "No ceremony yet" gets a much shorter TTL of its own.
- Verify a submission's signature before the roster read, so nothing that isn't internally consistent costs an S3 read.
- Drop `AUTHORIZED_KP_FINGERPRINTS` and its parsing.

### DIFF
--- a/crates/hashi-guardian-proxy/src/relay.rs
+++ b/crates/hashi-guardian-proxy/src/relay.rs
@@ -15,10 +15,8 @@
 
 use std::collections::BTreeMap;
 use std::sync::Arc;
-use std::time::Duration;
-use std::time::Instant;
 
-use crate::roster::latest_kp_roster;
+use crate::roster::RosterCache;
 use crate::widlog::LogStore;
 use hashi_types::guardian::GetGuardianInfoResponse;
 use hashi_types::guardian::KpSigned;
@@ -86,51 +84,6 @@ struct BackendStatus {
     provisioned: bool,
 }
 
-/// The committed roster changes only at ceremonies/rotations; a short TTL
-/// bounds S3 reads under submission spam without staleness that matters.
-const ROSTER_TTL: Duration = Duration::from_secs(60);
-
-/// TTL-cached view of the S3-committed KP roster. The mutex is held across the
-/// fetch, so concurrent misses collapse into one S3 read.
-struct RosterCache<L> {
-    store: L,
-    cached: Mutex<Option<(Instant, Arc<Vec<Fingerprint>>)>>,
-}
-
-impl<L: LogStore> RosterCache<L> {
-    fn new(store: L) -> Self {
-        Self {
-            store,
-            cached: Mutex::new(None),
-        }
-    }
-
-    async fn get(&self) -> Result<Arc<Vec<Fingerprint>>, Status> {
-        let mut cached = self.cached.lock().await;
-        if let Some((at, roster)) = cached.as_ref() {
-            if at.elapsed() < ROSTER_TTL {
-                return Ok(roster.clone());
-            }
-        }
-        match latest_kp_roster(&self.store).await {
-            Ok(Some(roster)) => {
-                let roster = Arc::new(roster);
-                *cached = Some((Instant::now(), roster.clone()));
-                Ok(roster)
-            }
-            // No ceremony has committed a share set yet: fail closed, uncached
-            // (so the first ceremony is authorized the moment its log lands).
-            Ok(None) => Err(Status::failed_precondition(
-                "no KP share log in the guardian bucket; run the key ceremony first",
-            )),
-            Err(e) => {
-                warn!(error = %format!("{e:#}"), "KP roster read failed");
-                Err(Status::unavailable("KP roster unavailable; retry"))
-            }
-        }
-    }
-}
-
 #[derive(Clone)]
 pub struct Relay<L> {
     client: GuardianServiceClient<Channel>,
@@ -139,11 +92,26 @@ pub struct Relay<L> {
 }
 
 impl<L: LogStore> Relay<L> {
-    pub fn new(channel: Channel, roster_store: L) -> Self {
+    pub fn new(channel: Channel, roster: Arc<RosterCache<L>>) -> Self {
         Self {
             client: GuardianServiceClient::new(channel),
             accumulator: Arc::new(Mutex::new(Accumulator::default())),
-            roster: Arc::new(RosterCache::new(roster_store)),
+            roster,
+        }
+    }
+
+    /// The ceremony's committed roster, mapped onto the relay's failure modes:
+    /// no share log yet is a definitive "not ready", a read error is transient.
+    async fn authorized_kps(&self) -> Result<Arc<Vec<Fingerprint>>, Status> {
+        match self.roster.get().await {
+            Ok(Some(roster)) => Ok(roster),
+            Ok(None) => Err(Status::failed_precondition(
+                "no KP share log in the guardian bucket; run the key ceremony first",
+            )),
+            Err(e) => {
+                warn!(error = %format!("{e:#}"), "KP roster read failed");
+                Err(Status::unavailable("KP roster unavailable; retry"))
+            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


crates/hashi-guardian-proxy/src/relay.rs
```diff
@@ -79,21 +88,64 @@ struct BackendStatus {
     provisioned: bool,
 }
 
+/// The committed roster changes only at ceremonies/rotations; a short TTL
+/// bounds S3 reads under submission spam without staleness that matters.
+const ROSTER_TTL: Duration = Duration::from_secs(60);
+
+/// TTL-cached view of the S3-committed KP roster. The mutex is held across the
+/// fetch, so concurrent misses collapse into one S3 read.
+struct RosterCache<L> {
```
mskd12: I think a cache is probably an overkill as these certs wont change much. I propose we do the following:
- Read it once when the service begins -> store it in a config
- Trigger a refresh when rotate_kp_certs endpoint is triggered

Note that rotate_kp_certs is introduced in #785 - so perhaps you can land this PR after that..

0xsiddharthks: On reading it once at startup though, I think that breaks the deploy ordering: the proxy is deployed *before* the ceremony runs, so on a fresh env it would come up with no share log, cache that, and then reject every submission until someone restarted it. The current code deliberately leaves the "no roster yet" case uncached so the first ceremony is authorized the moment its log lands.

The other reason for the TTL is cost rather than freshness. `single_provisioner_init` is a pre-auth DoS gate, so without a cache every spammed submission turns into S3 LISTs.

That said, your freshness point stands for cert rotation. How about keeping the cache but invalidating it in `provisioner_rotate_cert`? That gives the immediate refresh you're after without losing either property. Happy to do it here, or as a follow-up if you'd rather land this as-is — with the 60s TTL the staleness window is bounded either way.

crates/hashi-guardian-proxy/src/relay.rs
```diff
@@ -82,21 +86,64 @@ struct BackendStatus {
...
+
+impl<L: LogStore> RosterCache<L> {
+    fn new(store: L) -> Self {
+        Self {
+            store,
+            cached: Mutex::new(None),
+        }
+    }
+
+    async fn get(&self) -> Result<Arc<Vec<Fingerprint>>, Status> {
```
dlukegordon: This cache stays valid after `ProvisionerRotateCert`, so the new certificate can be rejected for up to 60 seconds. Please invalidate the roster cache after a successful certificate rotation.

crates/hashi-guardian-proxy/src/relay.rs
```diff
@@ -177,27 +224,23 @@ fn check_share_id(id: u32, num_shares: usize) -> Result<(), Status> {
...
-        if self.authorized_kp_fingerprints.is_empty() {
-            return Err(Status::failed_precondition(
-                "relay has no authorized KP roster configured; \
-                 set AUTHORIZED_KP_FINGERPRINTS on the proxy",
-            ));
-        }
+        // The ceremony's committed roster, not deploy config: a rotation
+        // re-deals shares without a proxy redeploy. Fails closed when no share
+        // log exists yet.
+        let roster = self.roster.get().await?;
```
dlukegordon: This reads S3 before checking whether the request is valid. Please decode and verify the signature first, then load the roster and check membership.

dlukegordon: Decoding first handles malformed requests, but signature verification still happens after the S3 read. Please verify the signature before calling `authorized_kps()` so well-formed requests with bad signatures do not hit S3.

0xsiddharthks: @dlukegordon Yeah I think that ordering was deliberate.
the signature is made by the cert the request itself carries, so it excludes nobody , and membership is the only check that actually rejects a caller, which is why I had it first.

But yeah, since the cert gets parsed and policy-validated during decode anyway, so most of the crypto is already paid by that point and verifying the signature first costs little.

I flipped the ordering, with a test pinning the order: the store has no share log, so anything reaching the roster answers FailedPrecondition (an Unauthenticated verdict proves the signature ran first).

crates/hashi-guardian-proxy/src/roster.rs
```diff
@@ -0,0 +1,483 @@
...
+    let bytes = log.get(&key).await?;
+    let roster = parse_roster(&bytes).with_context(|| format!("parse share log {key}"))?;
+    Ok(Some(roster))
+}
+
+/// Key of the latest share-state record: the lex-greatest object under the
+/// lex-greatest `kp-shares/` sharing-seq dir, else the lex-greatest flat
+/// `shares/` key (zero-padded seqs make lex order the seq order).
+async fn latest_share_log_key<L: LogStore>(log: &L) -> anyhow::Result<Option<String>> {
+    if let Some(dir) = log.list_dirs(KP_SHARES_PREFIX).await?.into_iter().max() {
```
dlukegordon: This picks the newest `kp-shares/` directory even if its ceremony never completed. Since the guardian writes `kp-shares/` before `ceremony/`, a failed ceremony can leave an orphan roster that blocks the current KPs. Please read the latest completed ceremony first, then load the latest certificate state for that ceremony’s `sharing_seq`.

crates/hashi-guardian-proxy/src/relay.rs
mskd12: Fine with leaving a TODO for `provisioner_rotate_cert` invalidation or doing it in this PR. Upto you!

mskd12 (approved):

crates/hashi-guardian-proxy/src/relay.rs
0xsiddharthks: Circled back on this — kept the TTL but added the invalidation you were after, in 97716b5. The relay and the forwarder share one `RosterCache` now and `provisioner_rotate_cert` drops it after a successful rotation, so a replacement cert is live immediately instead of waiting out the minute.

crates/hashi-guardian-proxy/src/relay.rs
0xsiddharthks: Already in — 97716b5, which landed just before this review. The relay and the forwarder share one `RosterCache` and `provisioner_rotate_cert` invalidates it after a successful rotation, so the new cert isn't rejected for the TTL.

crates/hashi-guardian-proxy/src/roster.rs
0xsiddharthks: Good catch, this one was real. Fixed in 6bfbb52.

The reader now takes the lex-greatest `ceremony/` key, pulls `sharing_seq` out of it and reads the share state under that seq — same resolution `read_latest_ceremony_state` does, which is the point: the relay has to agree with the enclave or it rejects the exact KPs the enclave would accept. A ceremony with no `kp-shares/` for its seq is a hard error now rather than a fallback to a roster that ceremony never authorised.

Two tests: an orphan `kp-shares/1/` with no `ceremony/1` doesn't move the roster, and it does move once that ceremony commits. Both fail if I put the old newest-dir-wins logic back, so they're actually pinning it.

Also swapped the hardcoded prefixes for hashi-types' `object_key_dir()` constructors while I was in there, so the layout has one definition — `widlog.rs` was already importing its prefix that way.

PR #782, opened 2026-07-09
[guardian-proxy] Route provisioning to a standby guardian backend

0xsiddharthks:
## Summary

During a rotation, KPs need to provision the *standby* guardian while nodes keep being served by the active one. The proxy only had one backend, so arming a standby meant pointing the whole proxy at it.

Give the relay its own backend: when `STANDBY_GUARDIAN_URL` is set it targets the standby, while forwarding, the wid cache and `/info` stay on the active guardian. Unset, it provisions the active backend as before.

## Changes

- New optional `STANDBY_GUARDIAN_URL` config, with its own lazy channel for the relay.
- New `GetStandbyInfo` RPC on the relay, exposing the relay backend's `GetGuardianInfo` so KP tooling pins the session it is actually submitting to — the node-facing `GetGuardianInfo` always answers for the active guardian.
- Confirm a submission's pinned session and hashes before treating a backend as already provisioned, on both paths that can answer `done()`. Two backends means a KP's `GetStandbyInfo` and its submission can reach differently-routed proxies mid-rollout, and the active guardian is always provisioned, so the old check would have reported a standby submission complete.
- Cache `GetStandbyInfo` for a few seconds: it is unauthenticated and every call makes the enclave mint a fresh Nitro attestation.
- `kp_provision` reads the target session through `GetStandbyInfo`, and the provisioning flow doc follows.


crates/hashi-guardian-proxy/src/main.rs
```diff
@@ -57,18 +58,21 @@ async fn main() -> Result<()> {
...
+    let relay_channel = match &config.standby_backend_url {
+        Some(url) => lazy_channel(url, &config)?,
+        None => channel.clone(),
+    };
 
     // One roster cache, shared: the relay authorizes submissions against it and
     // a cert rotation through the forwarder invalidates it.
     let roster = Arc::new(RosterCache::new(log_store.clone()));
-    let relay_svc = Relay::new(channel.clone(), roster.clone());
+    let relay_svc = Relay::new(relay_channel, roster.clone());
```
dlukegordon, line 75: With separate active and relay backends, the info request and submission can hit proxy instances with different routing during a rollout. `relay.rs:243` returns success for an already-provisioned backend before checking the session the KP pinned, so the active guardian can falsely complete a standby submission. Please check the session, config hash, and genesis hash before returning `done()`.

crates/hashi-guardian-proxy/src/relay.rs
```diff
@@ -197,6 +203,16 @@ fn check_share_id(id: u32, num_shares: usize) -> Result<(), Status> {
 
 #[tonic::async_trait]
 impl<L: LogStore> GuardianRelayService for Relay<L> {
+    async fn get_standby_info(
```
dlukegordon: This public RPC generates a fresh Nitro attestation on the standby for every request. An unauthenticated caller could flood it and block provisioning. Please add rate limiting or a short-lived cached response.

dlukegordon (commented):
`design/docs/guardian.mdx:100` still tells KPs to call `GetGuardianInfo`, which now returns the active guardian. Please update the flow to call the relay’s `GetStandbyInfo` RPC.

crates/hashi-guardian-proxy/src/main.rs
0xsiddharthks: Fixed. You're right and it's worse than just the rollout case — `provisioned` is only `enclave_btc_pubkey.is_some()`, which is permanently true for the active guardian, so it would also have accepted a backend provisioned under a different `config_hash`, which is exactly what blows up at activate after the stop.

Pulled it into a `match_backend`: session first (always readable, it's derived from the signing pubkey), then the config/genesis hashes while the arming is still there. Activation clears the arming, so a late retry on the pinned session is the one case that goes on session alone.

Also routed the post-batch re-check through the same function — it had the same gap in a narrower window, and leaving one `provisioned`-only path would have reopened it.

crates/hashi-guardian-proxy/src/relay.rs
0xsiddharthks: Yes, just confirmed it.

Added a short single-slot cache with the lock held across the fetch, so a burst collapses to one backend call. Safe to cache because the request carries no nonce, and the only effect of staleness is a KP pinning a session that has since moved — which `single_provisioner_init` reads fresh and rejects loudly.

Worth noting the node-facing `GetGuardianInfo` passthrough in forward.rs has the same exposure and predates this stack. 

I'll add a todo on linear to fix this.

0xsiddharthks:
On the `guardian.mdx:100` note — updated, KPs now go through the proxy's `GetStandbyInfo` there. Left line 171 alone since that's the MPC → active guardian flow, where `GetGuardianInfo` is still the right call.

dlukegordon (approved):

crates/hashi-guardian-proxy/src/relay.rs
mskd12: In theory, we don't need to expose Nitro attestation's via public RPC, no? (since hashi nodes don't check them) So maybe the right long-term fix is to introduce another Guardian /info endpoint without the attestation?

crates/hashi-guardian-proxy/src/relay.rs
```diff
@@ -204,6 +300,31 @@ fn check_share_id(id: u32, num_shares: usize) -> Result<(), Status> {
 
 #[tonic::async_trait]
 impl<L: LogStore> GuardianRelayService for Relay<L> {
+    /// The relay backend's `GetGuardianInfo`, verbatim. The lock is held across
+    /// the fetch, so a burst collapses into one backend call.
+    async fn get_standby_info(
```
mskd12: `GetStandbyInfo` is slightly misleading because, when `STANDBY_GUARDIAN_URL` is unset, it returns the active backend’s info. Could we name this `GetProvisioningTargetInfo` to reflect both single-guardian and standby flows?

mskd12 (approved):

crates/hashi-guardian-proxy/src/relay.rs
0xsiddharthks: Agreed on the direction, though I think the split falls the other way round.

The KP path is the one caller that genuinely needs the attestation — this endpoint exists so a KP can verify PCRs before it encrypts a share to the session key. It's the node-facing `GetGuardianInfo` that mints an attestation nobody reads, and that's the one on the hot path.

So the long-term shape is probably: keep the attestation on the provisioning surface, and give nodes an attestation-free `GuardianInfo`. That's a guardian API change rather than a proxy one, so I'll leave it out of here and put it on the Linear todo along with the `forward.rs` passthrough.

crates/hashi-guardian-proxy/src/relay.rs
0xsiddharthks: Fair, that name only describes the rotation case. Renamed to `GetProvisioningTargetInfo` in 1a05cd9 — proto, the init client, and the guardian.mdx diagram. It's unreleased, so nothing to keep compatible.

Took `verified_standby_guardian_info` → `verified_provisioning_target_info` on the init side too, since it read the same way. Left `verify_provisioned_standby_info` in `operator_activate` alone — that one really is about the standby.

dlukegordon (approved):

PR #783, opened 2026-07-09
[guardian-local] Rehearse guardian rotation in the local replica

0xsiddharthks:
## Summary

The local replica could bring up a single guardian but not rehearse a rotation, so the switchover sequence — arm a standby, flip, stop the old guardian, wait out its heartbeat fence, activate — could only be exercised against real infrastructure. This adds a second guardian to the compose stack and drives that sequence end to end.

## Changes

- Second guardian service in `compose.yaml`, plus `make` targets for the arm / flip / activate phases.
- `provision.sh` handles a standby target and the activation wait.
- Allow `GUARDIAN_OTHER_SESSION_QUIET_SECS` to shorten the quiet-period fence, gated on `non-enclave-dev` so a real build still compiles the ten-minute constant in.


0xsiddharthks:
Parking this — it's the local-replica rehearsal harness, not something the testnet rotation needs. Its only non-`docker/` change is the quiet-fence env override, and that's `#[cfg(feature = "non-enclave-dev")]`-gated, so a real-attestation build compiles the constant in regardless.

Reopening after the rotation.

PR #785, opened 2026-07-09
[Guardian] Support individual KP cert rotation

mskd12:
Stacked on #791.

## Summary

This PR lets a key provisioner replace one OpenPGP certificate in its roster entry without changing the share or Bitcoin key.

The command uses `kp_pgp_cert_path` to decrypt and sign. `--target-kp-pgp-fingerprint` selects the certificate to replace. The signing certificate may be the target or another certificate in the same roster entry. This allows a lost YubiKey to be replaced when another certificate for that KP is still available.

The proxy verifies the KP signature before forwarding the request. This is admission control that discards bulk unsigned or invalid traffic before it reaches the enclave. The guardian verifies the signature again, checks that the signer and target belong to the same share, verifies the submitted share against its commitment, replaces only the target ciphertext, and writes the next `cert_seq`.

The proxy code has a TODO for [#781](https://github.com/MystenLabs/hashi/pull/781). Once that PR provides an S3-backed authorization roster, the proxy can also check signer membership and invalidate its cached roster after a successful rotation.

## Validation

- `make fmt`
- CI


crates/hashi-guardian/src/withdraw_mode/provisioner_rotate_cert.rs
```diff
@@ -0,0 +1,174 @@
...
+        .map_err(|_| InvalidInputs("S3 logger not set".into()))?
+        .clone();
+    let mut reader = GuardianReader::from_s3_client(s3, init_config.pcr_allowlist().clone());
+
+    let latest_state = reader
+        .read_latest_ceremony_state(BuildPolicy::AnyAllowlisted)
+        .await
+        .map_err(|e| InternalError(format!("read latest ceremony state: {e}")))?
+        .ok_or_else(|| InvalidInputs("no ceremony log found during cert rotation".into()))?;
+    let active_instance = enclave.secret_sharing_instance()?;
```
dlukegordon: Activation clears ceremony_state, which is the only source used by secret_sharing_instance(). This therefore always fails on the activated guardian required by this RPC. Please retain the active instance in serving state or derive it from the verified S3 state.

mskd12: Good catch. I plan to do a deeper cleaning in a future PR to make sure this API is not allowed when the enclave is activated. For now, I just removed the call..

jessiemongeon1:
## Style Guide Audit

**7** violation(s) found. All must be fixed before merge.

### `design/docs/ts-sdk.mdx` (7 violation(s))

_7 violation(s) (4 regex, 3 claude)_

- **Line 13** — H1 only in frontmatter
  - Current: `# ⚠️ This repository is no longer active`
  - Fix: `Use ## or lower for section headings`
- **Line 15** — No em dashes in prose
  - Current: `—`
  - Fix: `Rewrite with comma, parentheses, or split sentence`
- **Line 15** — Use present tense, not future
  - Current: `will`
  - Fix: `Use present tense verb`
- **Line 19** — No em dashes in prose
  - Current: `—`
  - Fix: `Rewrite with comma, parentheses, or split sentence`
- **Line 13** — No H1 in body text
  - Current: `# ⚠️ This repository is no longer active`
  - Fix: `## This repository is no longer active`
- **Line 15** — Passive voice
  - Current: `the code has been transferred there`
  - Fix: `Mysten Labs transferred the code there`
- **Line 15** — Passive voice
  - Current: `this repository will not receive further updates and is kept for historical reference only`
  - Fix: `this repository does not receive further updates and exists for historical reference only`

---

crates/hashi-guardian/src/withdraw_mode/provisioner_rotate_cert.rs
```diff
@@ -0,0 +1,167 @@
...
+        &target_fingerprint,
+        new_recipient_fingerprint.clone(),
+        replacement_ciphertext,
+    )?;
+    let next_cert_seq = latest_state
+        .cert_seq
+        .checked_add(1)
+        .ok_or_else(|| InvalidInputs("cert_seq overflow".into()))?;
+
+    enclave
```
dlukegordon, line 149: write_log_record_or_abort detaches the S3 writer. If this RPC is cancelled while awaiting it, control_lock is released while the write continues. Another rotation can then read the old cert_seq and attempt different randomized ciphertext at the same S3 key, causing verify_existing_write to panic and abort the guardian. Can we make the complete read/modify/write transaction cancellation-safe and idempotent?

crates/hashi-guardian/src/withdraw_mode/provisioner_rotate_cert.rs
```diff
@@ -0,0 +1,167 @@
...
+            "signer fingerprint {signer_fingerprint} is assigned share id {}, not submitted share \
+             id {}",
+            signer_share_id.get(),
+            share_id.get()
+        )));
+    }
+
+    // The KP signature binds the ciphertext and the rest of this request to the
+    // current session and cert sequence, so no additional HPKE AAD is needed.
+    let share = decrypt_share(&encrypted_share, enclave.encryption_secret_key(), None)?;
```
dlukegordon, line 128: decrypt_share passes the decrypted plaintext to FieldBytes::from_slice without checking that it is exactly 32 bytes. An authorized KP can submit a validly encrypted and signed payload of another length, causing a panic and process abort. Please validate the plaintext length before scalar decoding.

crates/hashi-guardian-init/README.md
```diff
@@ -164,6 +165,41 @@ config. This command uses `kp_pgp_cert_path`, `relay_endpoint`, `hashi`,
...
+the KP's other certificates.
+
+It:
+
+1. Loads the signing cert from `kp_pgp_cert_path` and the current roster from
+   `kp_roster`. The signing cert and target fingerprint must belong to the same
+   KP/share entry. They may identify the same cert for a planned rotation or
+   different certs when replacing a lost YubiKey. The replacement cert must not
+   collide with another roster fingerprint.
+2. Fetches and verifies the active guardian's `GuardianInfo` through
```
dlukegordon, line 181: This says the active guardian’s sharing instance must match the latest ceremony log, but the implementation only checks that the BTC public key matches. If using the latest sharing instance regardless of which instance activated the guardian is intentional, please update this description accordingly.

crates/hashi-guardian-proxy/src/forward.rs
```diff
@@ -37,6 +39,17 @@ fn denied(rpc: &str) -> Status {
     ))
 }
 
+fn verify_provisioner_rotate_cert_signature(
+    request: &proto::SignedProvisionerRotateCertRequest,
+) -> Result<(), Status> {
+    let signed_request = KpSigned::<ProvisionerRotateCertRequest>::try_from(request.clone())
+        .map_err(|e| Status::invalid_argument(format!("malformed request: {e}")))?;
+    signed_request
```
dlukegordon, line 47: This verifies only that the supplied certificate signed the request; it does not establish roster membership. Anyone can generate a certificate and valid signature, so arbitrary requests still reach the enclave, acquire control_lock, and perform S3 reads. Please add roster-backed authorization or equivalent admission limiting before exposing this RPC.

crates/hashi-guardian-init/src/kp_rotate_cert.rs
```diff
@@ -0,0 +1,225 @@
...
+    let mut reader = GuardianReader::new(&guardian_s3, allowlist.clone())
+        .await
+        .context("connect to guardian log bucket")?;
+    let mut client =
+        pb::guardian_service_client::GuardianServiceClient::connect(cfg.relay_endpoint.clone())
+            .await
+            .with_context(|| {
+                format!("failed to connect to relay endpoint {}", cfg.relay_endpoint)
+            })?;
+    let endpoint_verified = verified_live_guardian_info(&mut client, allowlist.current_build())
```
dlukegordon, line 96: The attested endpoint's bucket_info is not compared with the configured guardian_s3 bucket. A wrong relay with the same BTC key and matching sequence could mutate another bucket before the local post-check fails. Please cross-check the signed bucket and S3 session information as kp_provision does.

crates/hashi-guardian-init/src/kp_rotate_cert.rs
```diff
@@ -0,0 +1,225 @@
...
+            &target_fingerprint_hex,
+            new_fingerprint.clone(),
+            replacement_ciphertext,
+        )?;
+    anyhow::ensure!(
+        response.encrypted_shares == expected_changed_entry,
+        "ProvisionerRotateCert response changed ciphertexts other than the requested certificate"
+    );
+
+    let updated_state = reader
```
dlukegordon, line 216: Requiring this response's snapshot to remain globally latest creates false failures when another rotation completes before this read. A lost response is also not safely retryable because the old configured roster no longer matches S3. Please verify the exact (sharing_seq, response.cert_seq) snapshot and provide an idempotent recovery path.

dlukegordon (changes requested):

crates/hashi-guardian/src/withdraw_mode/provisioner_rotate_cert.rs
mskd12: This seems like a broader pattern that needs to be fixed across multiple API surfaces. I will add a TODO to make sure control_lock is not released even if RPC is cancelled.

mskd12: Re idempotency: sure, we could add it. not sure how strongly you feel abt it. personally, it feels like a nice to have that can be deferred to future pr's to me.

crates/hashi-guardian/src/withdraw_mode/provisioner_rotate_cert.rs
mskd12: Good catch! Again a generic thing, but easy to fix, so I pushed the fix in this PR itself..

crates/hashi-guardian-proxy/src/forward.rs
mskd12: Yes, this was intentionally deferred to #781, as noted in the existing TODO and PR description.

mskd12: We could keep the proxy RPC disabled until that authorization is available if you prefer..

crates/hashi-guardian-init/src/kp_rotate_cert.rs
mskd12: Hmm, a concurrent KP cert rotation feels like an advanced concern for this PR.. I left a TODO for now so that we can address it later..

dlukegordon (approved):

PR #789, opened 2026-07-09
[guardian][enclave] Fix issue with eif_builds booting up in nitro enclave

0xsiddharthks:
The `eif_build` enclave never actually booted in a real Nitro enclave — it had only ever been run as the docker replica (TCP, not vsock), so a stack of bugs stayed hidden until I brought up the `testing` env. Four of them, all fixed here:

- **The stagex kernel is clang/LLD-linked and just doesn't execute under Nitro** — blank console, then `VsockTimeout`. A gcc build of the same config boots fine, so we build the kernel ourselves from checksummed upstream source in a hermetic stagex stage (with one host-tool-only objtool patch for musl's small `SIGSTKSZ`). Reproducible via `SOURCE_DATE_EPOCH`; the config is `COPY`d from the pinned `user-linux-nitro` image, so nothing is duplicated in-tree.
- **The cmdline asked for an old-style ramdisk root** (`initrd=…,SIZE root=/dev/ram0`), which panics the kernel (`Cannot open root device`). Nitro hands us an initramfs and execs `/init`, so we drop both.
- **nit was broken two ways**: the `user-nit` image moved its binary, so `COPY … /bin/init` landed it at `/nit` not `/init`; and its own `mount(proc, hidepid=2)` fails `EINVAL` on this kernel regardless. Dropped it — `run.sh` is the init now, doing the mounts and the Nitro heartbeat itself.
- Plus `CONFIG_NSM=y` and the `/proc /sys /dev /tmp` mountpoints.

Validated end-to-end on m5 and m7i: the make-built EIF boots and the guardian serves gRPC in a real enclave, with a reproducible PCR0.

Supersedes #784 (its initrd-size tweak was chasing a symptom of the cmdline this removes).

### DIFF
--- a/docker/hashi-guardian/Containerfile
+++ b/docker/hashi-guardian/Containerfile
@@ -20,11 +20,91 @@ FROM stagex/core-busybox@sha256:e4a30addc8939c8e232472de904d1d9e97fc2e735fca9a97
 FROM stagex/core-libzstd@sha256:bfe96f975a2a2a7a48aeb418b44accc79dd26d601a94d9f02b0562eaa0f89fdd AS core-libzstd
 FROM stagex/user-eif_build@sha256:8bbd8fb500322e719063645460f3942d04ca99048089adea237026443c9517b7 AS user-eif_build
 FROM stagex/user-gen_initramfs@sha256:1c957893705011044a1b11aefbcffd80dc0797f439b997bfbfe87df0a4a3756e AS user-gen_initramfs
-FROM stagex/user-linux-nitro@sha256:edc0a74392d7b3ca904fa0e999307aabbe4a84ad780328439a2def6bdee55189 AS user-linux-nitro
 FROM stagex/user-cpio@sha256:46ca4a247908c502e31a26842925b02cc7133fec619b19e0d5c3833f6d073d8e AS user-cpio
 FROM stagex/user-socat@sha256:70b21dec5de83d366babfa90215ce510118df88b798d4106f429bde3bf55973a AS user-socat
 FROM stagex/user-jq@sha256:1b2dddde7a2a35395430ebd99e07117bbc48fa29a0dba8a6b18d6d74bf762368 AS user-jq
-FROM stagex/user-nit@sha256:6a59292f829293117994c63fae27326e4dd95dea591441bf148571dfec5fa8d3 AS user-nit
+# Kept only for its kernel config; its prebuilt bzImage is clang-linked and does
+# not boot on Nitro, so we build the kernel ourselves (see the kernel stage).
+FROM stagex/user-linux-nitro@sha256:edc0a74392d7b3ca904fa0e999307aabbe4a84ad780328439a2def6bdee55189 AS user-linux-nitro
+
+# Toolchain for building the enclave kernel from source (see the kernel stage).
+FROM stagex/core-gmp@sha256:f90653ed07062074363542a77e26a2d16a142f9be5ebf6900b53f4ab8a82d39d AS core-gmp
+FROM stagex/core-mpfr@sha256:894c11d709646f4589541a6e49181aabaac843bbf429578acf767fabf5a40838 AS core-mpfr
+FROM stagex/core-mpc@sha256:6151912722b77722863de0e2784897f864bce2cc362cc8e4d777872536d060e4 AS core-mpc
+FROM stagex/core-isl@sha256:2f1b39d0339a5c8ba26b6d83d7dd1fa67047c385a521f14b27aee0fe79c994f5 AS core-isl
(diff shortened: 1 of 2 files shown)


docker/hashi-guardian/Containerfile
```diff
@@ -20,11 +20,91 @@ FROM stagex/core-busybox@sha256:e4a30addc8939c8e232472de904d1d9e97fc2e735fca9a97
...
-FROM stagex/user-linux-nitro@sha256:edc0a74392d7b3ca904fa0e999307aabbe4a84ad780328439a2def6bdee55189 AS user-linux-nitro
 FROM stagex/user-cpio@sha256:46ca4a247908c502e31a26842925b02cc7133fec619b19e0d5c3833f6d073d8e AS user-cpio
 FROM stagex/user-socat@sha256:70b21dec5de83d366babfa90215ce510118df88b798d4106f429bde3bf55973a AS user-socat
 FROM stagex/user-jq@sha256:1b2dddde7a2a35395430ebd99e07117bbc48fa29a0dba8a6b18d6d74bf762368 AS user-jq
-FROM stagex/user-nit@sha256:6a59292f829293117994c63fae27326e4dd95dea591441bf148571dfec5fa8d3 AS user-nit
+# Kept only for its kernel config; its prebuilt bzImage is clang-linked and does
+# not boot on Nitro, so we build the kernel ourselves (see the kernel stage).
+FROM stagex/user-linux-nitro@sha256:edc0a74392d7b3ca904fa0e999307aabbe4a84ad780328439a2def6bdee55189 AS user-linux-nitro
+
+# Toolchain for building the enclave kernel from source (see the kernel stage).
```
dlukegordon, line 30: Are you sure we need all these new deps? There isn't some umbrella package?

docker/hashi-guardian/Containerfile
```diff
@@ -20,11 +20,91 @@ FROM stagex/core-busybox@sha256:e4a30addc8939c8e232472de904d1d9e97fc2e735fca9a97
...
+	https://cdn.kernel.org/pub/linux/kernel/v7.x/linux-7.0.14.tar.xz /linux.tar.xz
+COPY --from=user-linux-nitro /linux.config /config
+ENV SOURCE_DATE_EPOCH=1 KBUILD_BUILD_TIMESTAMP=1 KBUILD_BUILD_USER=stagex KBUILD_BUILD_HOST=stagex
+RUN <<-EOF
+	set -eux
+	tar -xf /linux.tar.xz
+	cd linux-7.0.14
+	cp /config .config
+	sed -i "/CONFIG_CC_IS_CLANG/d; /CONFIG_LD_IS_LLD/d; /CONFIG_AS_IS_LLVM/d; /CONFIG_CC_VERSION_TEXT/d; /CONFIG_GCC_/d; /CONFIG_CLANG_/d; /CONFIG_LD_VERSION/d; /CONFIG_LLD_VERSION/d" .config
+	sed -i "s/\bSIGSTKSZ\b/65536/g" tools/objtool/signal.c
```
dlukegordon, line 102: These seds looks pretty hacky. I think fine if it unblocks us for now, but we should investigate further later. Can you add a TODO?

docker/hashi-guardian/run.sh
```diff
@@ -2,16 +2,35 @@
...
+export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/
 export LD_LIBRARY_PATH=/lib:$LD_LIBRARY_PATH
+echo "run.sh script is running"
+
+# The Nitro loader hands us a bare initramfs root; mount the pseudo-filesystems.
+# Tolerate an already-mounted fs (the kernel auto-mounts devtmpfs).
+busybox mount -t proc proc /proc 2>/dev/null || :
+busybox mount -t sysfs sysfs /sys 2>/dev/null || :
+busybox mount -t devtmpfs devtmpfs /dev 2>/dev/null || :
+busybox mount -t tmpfs tmpfs /tmp 2>/dev/null || :
```
dlukegordon, line 23: For these lines, can we do `||` instead of `2>/dev/null ||`  so we don't silence errors?

docker/hashi-guardian/run.sh
```diff
@@ -2,16 +2,35 @@
 # Copyright (c), Mysten Labs, Inc.
```
dlukegordon, line 2: docker/hashi-guardian-local/run.local.sh is documented (README:38) as a hand-maintained mirror of run.sh, and it wasn't touched. The new mount/heartbeat blocks are legitimately enclave-only, but exec /guardian and the PATH export aren't mirrored either. Fine to skip, but consider a comment in run.sh marking which sections are enclave-only so the next person diffing the two files knows what's intentional.

dlukegordon (approved):

PR #791, opened 2026-07-09
[Guardian] Support multiple KP certs per share

mskd12:
## Summary

This PR lets each KP roster entry contain more than one PGP certificate while still assigning one share per KP. The guardian encrypts that share once per accepted certificate, stores the ciphertexts by fingerprint, rejects duplicate fingerprints, and verifies each ciphertext against its matching certificate. KP ceremony verification decrypts and commitment-checks every copy; provisioning uses the selected certificate.

New guardian logs use schema V2 for the multi-certificate KP-share shape. Readers retain the deployed V1 one-certificate shape for signature verification, then normalize verified records to the current `LogMessage`. Coverage includes a real signed V1 testnet KP-share record.

## Validation

- `make fmt`
- `cargo test -p hashi-types deployed_v1_kp_share_state_verifies_and_normalizes`
- CI


jessiemongeon1:
## Style Guide Audit

crates/hashi-types/src/guardian/log/message.rs
```diff
@@ -148,11 +148,11 @@ impl GenesisLogMessage {
 pub struct KpShareStateLogMessage {
     pub sharing_seq: u64,
     pub cert_seq: u64,
-    pub encrypted_shares: KPEncryptedShares,
+    pub encrypted_shares: KPEncryptedSharesRoster,
```
dlukegordon, line 198: This changes the persisted kp-shares v1 format, so existing signed recovery records will no longer deserialize. Can we introduce a v2 log schema and retain v1 decoding? Otherwise upgraded guardians cannot recover the existing BTC key.

crates/hashi-guardian-init/README.md
```diff
@@ -134,11 +140,11 @@ is held in this process' memory long enough to verify and re-encrypt it. It:
...
    enclave.
-5. Reads this KP's encrypted share from the latest `kp-shares/{seq}/` state,
-   verifies every share's recipients against the
-   roster, finds the one labeled for this KP's cert fingerprint, decrypts it via
-   the yubikey (`gpg --decrypt` over a pipe; the plaintext stays in memory and
-   never touches disk), and verifies the decrypted share against its commitment.
+5. Reads this KP's PGP-encrypted share from the latest `kp-shares/{seq}/`
+   state, verifies each encrypted copy's recipient against the roster, then
+   decrypts and commitment-checks only the copy selected by
+   `kp_pgp_cert_path` (`gpg --decrypt` over a pipe; the plaintext stays in
```
dlukegordon, line 149: Selecting an alternate cert here will still fail at the relay unless its fingerprint is included in `AUTHORIZED_KP_FINGERPRINTS`. Can we document that every cert from every roster entry must be added to the relay allowlist, and ideally derive it automatically?

mskd12: @siddharthsharma has a draft PR (#781) to derive it automatically. I was assuming this will be taken care of.

crates/hashi-guardian-init/src/kp_roster.rs
```diff
@@ -44,149 +44,223 @@ use zeroize::Zeroizing;
...
+            .kp_pgp_cert_paths
+            .iter()
+            .enumerate()
+            .map(|(idx, cert_paths)| {
+                let certs = cert_paths
+                    .paths()
+                    .into_iter()
+                    .map(load_cert)
+                    .collect::<Result<Vec<_>>>()?;
+                KpCerts::new(certs)
```
dlukegordon, line 119: PgpPublicCert only requires an encryption key, but every accepted certificate must later sign provisioning requests. Please reject certificates without a usable signing key here so an invalid backup is not discovered during recovery.

mskd12: added it into `validate_pgp_cert` - pls take a look

dlukegordon (approved):

PR #792, opened 2026-07-09
[Guardian] Improve S3 log write robustness

mskd12:
## Problem

The enclave's committed limiter/seq state can silently lag S3 (the source of truth): an S3 log write can return an error while the record actually persists (a lost ACK), so the enclave reverts its limiter while a durable record exists. Since the seq gate checks only local next_seq, the enclave can then reuse a seq S3 already committed (a collision that corrupts recovery) or under-count the limiter.

## Fix

- **Shared fail-stop writer:** every non-init log creates one LogRecord and retries that exact key and body every 10 seconds. If no attempt is confirmed within five minutes, the enclave aborts. The writer runs as a detached Tokio task so caller cancellation cannot abandon an in-flight write.
- **Init exception:** only log_init skips grace-period retries, providing quick fail-stop for basic S3 write/access issues. A failed init leaves an incomplete enclave that cannot serve and is expected to restart, so S3 being ahead after a lost acknowledgement is acceptable.
- **Verified conditional creates:** writes use If-None-Match: *. A 412 triggers a locked point read; matching bytes confirm the intended record is durable, while different bytes indicate an unrecoverable write-once key collision and abort immediately.
- **Withdrawal commit ownership:** a successful withdrawal task owns the limiter mutex guard until its success log is confirmed. There is no rollback path after consumption; any unexpected post-consume error aborts, while the guard keeps later withdrawals serialized behind the durable write.
- **Simplified heartbeat:** heartbeat now only schedules records and advances its sequence after confirmed durability. Retry and fail-stop behavior comes from the shared writer.
- **Restart recovery:** after fail-stop, activation waits for the prior session quiet period and reconstructs limiter state from S3.

## Testing

- Matching and mismatching 412 responses.
- Application retries beyond the AWS SDK retry budget.
- Fail-stop after the configured failure interval.
- Full hashi-guardian library test suite.


crates/hashi-guardian/src/s3_client.rs
```diff
@@ -28,8 +28,14 @@ use hashi_types::guardian::VerifiedSessionInfo;
...
 use serde::Serialize;
 use tracing::info;
+use tracing::warn;
 
+/// Maximum attempts the AWS SDK makes before returning one write failure.
 const MAX_RETRY_ATTEMPTS: u32 = 5;
+/// Delay between application-level retries of an immutable S3 log write.
+const S3_WRITE_RETRY_INTERVAL: Duration = Duration::from_secs(10);
+/// Maximum write failure interval before the enclave aborts.
+const MAX_S3_WRITE_FAILURE_INTERVAL: Duration = Duration::from_mins(5);
```
zhouwfang: This is now half of the no-double-live invariant: max alive-but-silent \approx HEARTBEAT_INTERVAL + MAX_S3_WRITE_FAILURE_INTERVAL \approx 6min, which must stay below OTHER_SESSION_QUIET_PERIOD (10 min, as in heartbeat_checks.rs), but nothing ties the constants together. Can we document or const-assert the relation so a future bump can't silently open a double-live window?

crates/hashi-guardian/src/s3_client.rs
```diff
@@ -161,6 +223,17 @@ impl GuardianS3Client {
         Ok(())
     }
 
+    async fn verify_existing_write(&self, key: &str, expected_body: &[u8]) -> GuardianResult<()> {
```
zhouwfang, line 227: It doesn't need the lock-metadata assertion: byte-compare alone decides durable-vs-squat, and going through get_locked_object_bytes means a lock-less squat (or a retention-read perms gap) surfaces as the generic 5-min timeout panic instead of the crisp mismatch one. Consider a plain GET here.

benr-ml (approved):
the new logic looks good to me, but pls address the other questions

crates/hashi-guardian/src/s3_client.rs
mskd12: Makes sense. I replaced throwing an Err for "lock-less squat" or "retention-read perms gap" with a panic.

crates/hashi-guardian/src/s3_client.rs
mskd12: That's an excellent suggestion! Thanks. Added to lib.rs

zhouwfang (approved):

PR #804, opened 2026-07-13
docs: Fix operator runbook config template and key-setup guidance

Bridgerz:
Fixes for issues validators hit during testnet onboarding (reported by ContributionDAO, Staking Facilities, Chainode Tech, Brightlystake, and Michal Zegarek in the operator channel).

> **Rebased twice over overlapping main work** (#806, #807, e1cf8393, 21ba9d40). Superseded pieces were dropped each time; this PR now carries only what remains relevant.

## Bitcoin Core version is a minimum, not a target

§1.5 said "Hashi targets Bitcoin Core v29", which read as an exact-version requirement. Per Brandon: any v29+ works, including the latest release. Now stated as "v29 or newer".

## TLS key generation

Added the `openssl genpkey` command to §2.2 and a file-permissions note (an unreadable TLS key file caused an incomplete registration; the loud-failure code fix is #805). TLS keys remain PEM/DER-only — unaffected by the new multi-format operator-key loading.

## §2.3 vs §2.4 confusion

Both sections showed the same `hashi register` command with differently-named placeholders, and it wasn't stated whether that address is the validator address or the fresh operator key. Unified the placeholder, stated it's the operator key's address, and clarified that one registration tx both registers and delegates — re-running is idempotent.

## Config template polish (on top of #807)

A prose "field order matters" note above the template (describing the post-#807 unknown-field error), and the `[hashi-ids]` template comment now points at the Networks table where the published IDs live.

## Dropped as superseded

- Template section reorder → #807
- backup-pgp-cert inline-only wording → #806 made the field accept paths; original wording restored
- "sui keytool formats not accepted" §2.4 guidance → 21ba9d40 + e1cf8393 added native suiprivkey/Base64 support and documented all formats

Also includes the style-guide audit fixes (no em dashes in prose, "for example" instead of "e.g.").

### DIFF
--- a/design/docs/node-operator-runbook.mdx
+++ b/design/docs/node-operator-runbook.mdx
@@ -178,11 +178,17 @@ bitcoin-cli -signet getnetworkinfo                        # NODE_COMPACT_FILTERS
 
 ### 2.2 TLS key (dedicated)
 
-Generate a fresh Ed25519 TLS key dedicated to your Hashi node, separate from your Sui validator's TLS key, for clean separation of concerns. This key authenticates your Hashi node to other committee members, and the node registers the corresponding public key onchain during registration. Set `tls-private-key` to a PKCS#8 PEM/DER file path or an inline PEM string.
+Generate a fresh Ed25519 TLS key dedicated to your Hashi node, separate from your Sui validator's TLS key, for clean separation of concerns. This key authenticates your Hashi node to other committee members, and the node registers the corresponding public key onchain during registration. Set `tls-private-key` to a PKCS#8 PEM/DER file path or an inline PEM string:
+
+```bash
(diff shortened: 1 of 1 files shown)


jessiemongeon1:
## Style Guide Audit

**2** violation(s) found. All must be fixed before merge.

### `design/docs/node-operator-runbook.mdx` (2 violation(s))

_2 violation(s) (0 regex, 2 claude)_

- **Line 209** — no-italic
  - Current: `*not*`
  - Fix: `Use bold (**not**) or rewrite to avoid italic emphasis`
- **Line 242** — no-italic
  - Current: `*and*`
  - Fix: `Use bold (**and**) or rewrite to avoid italic emphasis`

---

bmwill (approved):

PR #808, opened 2026-07-13
config: support all sui key formats for operator key

bmwill:


jessiemongeon1:
## Style Guide Audit

### `design/docs/node-operator-runbook.mdx` (3 violation(s))

_3 violation(s) (2 regex, 1 claude)_

- **Line 209** — Use "basic" not "simple"
  - Current: `simple`
  - Fix: `basic`
- **Line 209** — Use "might" not "may"
  - Current: `may`
  - Fix: `might`
- **Line 209** — word-preference
  - Current: `The value may be a file path`
  - Fix: `The value might be a file path`

---

dlukegordon (commented):
We also need to update `crates/hashi/src/backup.rs:329-357`.

The key loader now accepts inline suiprivkey and Base64 values, but backup discovery still recognizes only inline PEM. It treats either new inline format as a path, rejects it as missing, and interpolates the complete value into the error at line 356. Automatic backups log that error at crates/hashi/src/backup.rs:109, exposing the private key.

dlukegordon (approved):

PR #810, opened 2026-07-13
Add per-request deadlines to `sui-rpc` unary calls

zhouwfang:
#16


bmwill (commented):
Grpc already has a built in way to express this. Just include the timeout in the tonic::Request

zhouwfang:
> Grpc already has a built in way to express this. Just include the timeout in the tonic::Request

Client-side they're the same mechanism. tonic enforces Request::set_timeout by racing the response future, just like this layer. Channel-level for two reasons: set_timeout is opt-in per call (16 sites), so one forgotten site reintroduces the hang; and it puts grpc-timeout on the wire, where whether the subscription survives depends on the server/proxy (tonic only bounds the accept; envoy/grpc-go cancel the stream). Can inject grpc-timeout centrally instead (exempting the
subscription) if you prefer the built-in protocol.

bmwill:
>reintroduces the hang

can you elaborate more on what hang you're referring to?

zhouwfang:
> > reintroduces the hang
> 
> can you elaborate more on what hang you're referring to?

A server that accepts the connection but never responds. Keepalives only catch dead transports; an accepted
request with no deadline pends forever, and callers' retries only run on Err, which never comes — e.g. the rejoin e2e sat ~30 min against a wedged localnet until Deepak cancelled it; a wedged fullnode would do the same to a real node. Per-call set_timeout makes that protection opt-in per site; the channel layer makes it the default.

bmwill (approved):

PR #813, opened 2026-07-14
docs: Sync config and governance docs with on-chain reality

Bridgerz:
Audit of the design docs against the live testnet deployment and the last week of main, prompted by operator questions about current config values.

## Wrong default

`config.mdx` said `bitcoin_deposit_time_delay_ms` defaults to **1000 ms**; the Move default, and the value live on testnet, is **600000 ms** (10 minutes). The adjacent comment in `btc_config.move` calls the same constant "15 minutes" — fixed both. (Values verified by reading the deployed Hashi object's config VecMap directly.)

## Missing config keys (5 of 12 undocumented)

`config.mdx` covered 7 keys. Added:
- `governance_emergency_pause_threshold_bps` / `governance_emergency_unpause_threshold_bps` — the asymmetric emergency-pause quorums (5% to pause, two thirds to unpause), semantics verified in `emergency_pause.move`
- `mpc_threshold_in_basis_points`, `mpc_max_faulty_in_basis_points`, `mpc_weight_reduction_allowed_delta`, `mpc_nonce_generation_protocol` — with the pinned-per-committee semantics from `mpc_config::pin` (mid-epoch updates never affect the active committee), valid ranges from `is_valid_value`, and the AVID protocol value (`1`) that the recent AVID PRs made selectable

## Missing proposal types (3 of 7 undocumented)

`governance-actions.mdx` listed Upgrade / EnableVersion / DisableVersion / UpdateConfig as "the current set." Added `EmergencyPause`, `UpdateGuardian` (URL only; the guardian BTC key is immutable because rotation would invalidate derived deposit addresses), and `AbortReconfig` (cross-linked to the reconfiguration doc).

## Audited and found consistent (no change)

- `reconfiguration.mdx` vs #492's `ProposalExecuted<T>` event change
- runbook backup docs vs the inline-key backup-discovery fix
- `UpdateConfig` 2/3 threshold claim (all four base proposal types are 6667 bps)

**Flagged, not addressed here:** AVID has zero design-doc coverage (`mpc-protocol.mdx` never mentions nonce generation) — that needs an author with the protocol context. The runbook's curated "Key metrics" table also predates the two new `hashi_mpc_avid_*` metrics; happy to add them if ops wants them watched.

### DIFF
--- a/design/docs/config.mdx
+++ b/design/docs/config.mdx
@@ -33,7 +33,7 @@ unspendable UTXOs.
 | | |
 |---|---|
 | **Type** | `u64` |
-| **Default** | `1000` |
+| **Default** | `600000` (10 minutes) |
 | **Unit** | milliseconds |
 
 The minimum time that must elapse between a deposit being approved by the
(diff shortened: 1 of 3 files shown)


bmwill (approved):

jessiemongeon1:
## Style Guide Audit

All 2 file(s) pass the style guide audit.

PR #814, opened 2026-07-14
docs: Answer the second round of operator onboarding questions

Bridgerz:
Stacked on #804 (same sections of the runbook). Fixes from the second wave of validator-channel feedback, all claims verified against code before documenting:

## Offline signing commands (asked by Coinage x DAIC, answered in-thread by Nodes.Guru)

§2.3 now shows the exact `sui keytool sign --address ... --data ...` and `sui client execute-signed-tx --tx-bytes ... --signatures ...` invocations (flags verified against the current sui CLI), with a note for hardware/multisig custody. §5.3 cross-references instead of repeating.

## Operator address derivation (Coinage x DAIC)

Operators who generated their key with openssl had no documented way to compute its Sui address. §2.4 now covers both paths: `sui keytool generate` prints it, and for an existing PEM there's a one-liner (pubkey DER → blake2b-256 over `0x00 || pubkey`) that I verified derives the identical address to `sui keytool` for a test key.

## Fund the operator with at least 2 SUI (hit by Brightlystake)

`sweep_to_address_balance` bootstraps the account's address balance with **exactly 1 SUI** at startup (`sui_tx_executor.rs:2101`), so an operator holding exactly 1 SUI fails the simulation with `InsufficientCoinBalance` — precisely the error reported. §2.4 now says to fund with at least 2 SUI and explains why.

## Metadata updates are automatic (asked by Brightlystake, Triton.One)

Verified: the register builder diffs config vs the on-chain member record and sends only stale fields, and the node runs it at startup. §2.4 now states that editing the config and restarting is sufficient; no manual re-register.

## "Where do operator steps end?" (Galaxy)

§5.4 gains an explicit **last operator step** marker; steps 4-6 are admin/network.

### DIFF
--- a/design/docs/node-operator-runbook.mdx
+++ b/design/docs/node-operator-runbook.mdx
@@ -212,7 +212,7 @@ sui client execute-signed-tx --tx-bytes <TX_BYTES_BASE64> --signatures <SIGNATUR
 
 For hardware wallets or multisig accounts, produce the signature with your own signing tooling instead of `sui keytool sign`; the broadcast step is the same.
 
-`--operator-address` is the Sui address of the **fresh operator key** you generate in [2.4](#24-operator-key-day-to-day-signing). It is *not* your validator address. Passing it here delegates day-to-day signing in the same transaction, so the validator account key can go straight back to cold storage.
+`--operator-address` is the Sui address of the **fresh operator key** you generate in [2.4](#24-operator-key-day-to-day-signing). It is **not** your validator address. Passing it here delegates day-to-day signing in the same transaction, so the validator account key can go straight back to cold storage.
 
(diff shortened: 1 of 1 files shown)


jessiemongeon1:
## Style Guide Audit

### `design/docs/node-operator-runbook.mdx` (3 violation(s))

_3 violation(s) (1 regex, 2 claude)_

- **Line 268** — "on-chain" is one word → "onchain"
  - Current: `on-chain`
  - Fix: `onchain`
- **Line 215** — no-italic-text
  - Current: `*not*`
  - Fix: `**not**`
- **Line 257** — no-italic-text
  - Current: `*and*`
  - Fix: `**and**`

---

bmwill (approved):

PR #820, opened 2026-07-14
Adopt fastcrypto #988 with `use_legacy=true`

zhouwfang:
Followup PR: MystenLabs/hashi#821

[https://github.com/MystenLabs/fastcrypto/pull/988](<https://github.com/MystenLabs/fastcrypto/pull/988>)  <br>MystenLabs/hashi#193

### DIFF
--- a/crates/hashi/src/constants.rs
+++ b/crates/hashi/src/constants.rs
@@ -23,6 +23,8 @@ pub const BITCOIN_REGTEST_CHAIN_ID: &str =
 /// `initial_pool_size / PRESIG_REFILL_DIVISOR`.
 pub const PRESIG_REFILL_DIVISOR: usize = 2;
 
+pub const PRESIG_USE_LEGACY_EXTRACTION: bool = true;
+
 pub fn is_production_sui_chain(chain_id: &str) -> bool {
     chain_id == SUI_MAINNET_CHAIN_ID || chain_id == SUI_TESTNET_CHAIN_ID
 }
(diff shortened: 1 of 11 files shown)


crates/hashi/src/constants.rs
```diff
@@ -23,6 +23,8 @@ pub const BITCOIN_REGTEST_CHAIN_ID: &str =
 /// `initial_pool_size / PRESIG_REFILL_DIVISOR`.
 pub const PRESIG_REFILL_DIVISOR: usize = 2;
 
+pub const PRESIG_USE_LEGACY_EXTRACTION: bool = true;
```
benr-ml, line 26: i'm assuming we are planning to switch to a protocol config in a follow up pr, to enable it on the next version?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2546,18 +2526,49 @@ impl MpcManager {
...
+            .ok_or_else(|| {
+                MpcError::NotReady("no verified common message for this AVID round".into())
+            })?;
+        let receiver = self.create_avid_nonce_receiver(dealer, batch_index)?;
+        let message = batch_avss_avid::AvssMessage {
+            common: state.common,
+            ciphertext: state.own_ciphertext,
+        };
+        let (_, _, verified_common) = receiver
+            .process_avss_message(&message)
```
benr-ml, line 2569: nit - we verify common every time we call this function. not super expensive, but avoidable, so maybe let's leave a todo to consider changing that in the future (e.g., add serde to VerifiedAvssCommonMessage and store it)

benr-ml (approved):

crates/hashi/src/constants.rs
zhouwfang: @benr-ml (I had listed the followup PR in the PR description: #821.) Actually no config, we can't make any Move change now, the frozen package can't take one. I added a temporary gate on a compiled per-chain activation epoch. Once testnet flips we delete the legacy formula entirely.

PR #821, opened 2026-07-14
Gate presig derivation on a compiled per-chain activation epoch

zhouwfang:
MystenLabs/hashi#193


bmwill (changes requested):
We cannot make any move changes now that we've moved to testnet, so you'll need to find a different way to make this change and we can make note of move changes we'd like to make before we go to mainnet

zhouwfang:
> We cannot make any move changes now that we've moved to testnet, so you'll need to find a different way to make this change and we can make note of move changes we'd like to make before we go to mainnet

The new approach does not make any Move change. Comments in `crates/hashi/src/constants.rs` should be helpful for understanding. WDYT?

crates/hashi/src/constants.rs
```diff
@@ -23,18 +23,45 @@ pub const BITCOIN_REGTEST_CHAIN_ID: &str =
...
 
+/// Epoch at which the privacy-threshold presignature extraction activates;
+/// earlier epochs use the legacy.
+/// All validators must agree on this value for every epoch they serve,
+/// so flipping testnet means shipping a release with a concrete epoch,
+/// chosen at least one epoch ahead and only after the whole fleet
+/// verifiably runs it.
+pub fn presignature_derivation_activation_epoch(chain_id: &str) -> u64 {
+    if chain_id == SUI_TESTNET_CHAIN_ID {
+        u64::MAX
```
bmwill: Yes this works, if you want set this to an epoch number ~2 weeks or so in the future from now and we can tell validators to update to a new version, after which point we can remove all this extra machinery

bmwill (approved):

bmwill:
Also @zhouwfang can you file an issue for us to remember that we need to make it easier for us to add new config params to the epoch-specific configs?

crates/hashi/src/constants.rs
zhouwfang: Set to epoch 20.

PR #823, opened 2026-07-15
docs: refresh TypeScript SDK page from hashi-ts-sdk README

Bridgerz:
Unblocks CI for every open PR (first seen on #822): `design/docs/ts-sdk.mdx` is generated from the external `hashi-ts-sdk` README, whose latest update (testnet support 🎉) made the committed page stale — so the `docs` job's `make is-dirty` now fails everywhere.

This is the verbatim output of `node design/scripts/fetch-sdk-readme.js`.

## Why the automation didn't catch it

The `refresh-sdk-docs` workflow exists for exactly this, but its `git push` to main is now rejected by branch protection (verified by dispatching it: `GH006: Protected branch update failed ... Changes must be made through a pull request`). Its nightly cron has presumably been failing silently since protection was tightened. **Follow-up needed:** convert the workflow to open a PR (for example `peter-evans/create-pull-request`) or grant it bypass — otherwise this manual dance repeats on every SDK release.

PR #826, opened 2026-07-15
docs: refresh TypeScript SDK page from hashi-ts-sdk README (style fixes)

Bridgerz:
Companion to [hashi-ts-sdk#45](https://github.com/MystenLabs/hashi-ts-sdk/pull/45), which fixed the 14 style-guide violations the audit flagged on #823 at their source. Verbatim `fetch-sdk-readme.js` output.

**Merge promptly:** the `docs` CI job fetches the live upstream README at job time, so every open PR's docs check fails from the moment #45 merged until this lands (same dynamic as #823).

Structural follow-up worth a decision from whoever owns the docs pipeline — this is the second manual round-trip in two days:
1. Fix `refresh-sdk-docs` to open a PR instead of pushing to main (its push is rejected by branch protection, so the nightly sync has been failing silently), **or**
2. Pin `fetch-sdk-readme.js` to a README git revision bumped explicitly here, so upstream edits can never break this repo's CI asynchronously.

### DIFF
--- a/design/docs/ts-sdk.mdx
+++ b/design/docs/ts-sdk.mdx
@@ -12,7 +12,7 @@ TypeScript SDK for the [Hashi](https://github.com/MystenLabs/hashi) protocol. Ha
 
 :::warning
 
-**Not production-ready.** This SDK is pre-1.0 and under active development. The API may change without notice and only Sui testnet and devnet are wired up. Do not use it in production environments yet.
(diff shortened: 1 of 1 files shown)


Bridgerz:
Folding this into #822 instead (approval there survives new commits), so both land in one merge. The regen commit is cherry-picked onto that branch verbatim.

PR #835, opened 2026-07-16
[Guardian] Authenticate provisioner init requests with KP signatures

mskd12:
PI submissions are now typed KP-signed requests binding the guardian session, OI config hash, and encrypted share. The relay preserves those signed envelopes, and the enclave verifies each signature plus the OI-pinned signer-fingerprint-to-share assignment before decrypting and commitment-checking the batch. PI HPKE no longer uses the config hash as AAD because the signed request carries the binding.

This is stacked on precursor #836, which pins `CeremonyState` during OI. RotateKps retains its existing state-hash AAD and migration TODO; genesis authorization, multiple certificates per KP, and certificate rotation remain out of scope. Validated with `make fmt` and `cargo run -p proto-build`; broader checks are left to CI per repository guidance.


jessiemongeon1:
## Style Guide Audit

0xsiddharthks (approved):

PR #836, opened 2026-07-16
[Guardian] Read ceremony and KP share state together

mskd12:
This adds a flat, session-free `CeremonyState` built from the ceremony and KP-share log messages or from a live setup response. `GuardianReader` can return the latest ceremony together with its latest KP-share state, rejecting a ceremony without corresponding shares. Verified point reads log the selected object key and writing session for traceability.

Withdraw OI reads and stores the complete state once, and later init phases rely on that pinned snapshot without rereading ceremony or KP-share logs. Provisioning and roster consumers use the same combined reader. Validated with `make fmt`; broader checks are left to CI per repository guidance.


jessiemongeon1:
## Style Guide Audit

0xsiddharthks (approved):

PR #837, opened 2026-07-16
Gate the nonce collection weight on the presig derivation version

zhouwfang:
Followup on [https://github.com/MystenLabs/fastcrypto/pull/988](<https://github.com/MystenLabs/fastcrypto/pull/988>)

MystenLabs/hashi#193


arnab-roy (approved):
LGTM after cross-checking with the spec and the sync on optional extra accumulation.

PR #839, opened 2026-07-16
[onchain] Poll the config snapshot onchain until final_publish

0xsiddharthks:
## Changes

- `watcher.rs`: while the launch is pending, poll the Hashi config and swap it into the snapshot when it changed.
  - Once `guardian_url` and the BTC key are both on-chain the poll retires. 
  - Bounded (10s) so a wedged connection can't hang the checkpoint loop.
- `config.rs`: new `onchain_config_poll_interval_ms` node-config knob for the pending-launch poll, default 5 minutes.
- Derive `PartialEq` on `Config` / `UpgradeCap` for the change check.

### DIFF
--- a/crates/e2e-tests/src/hashi_network.rs
+++ b/crates/e2e-tests/src/hashi_network.rs
@@ -178,6 +178,26 @@ impl HashiNodeHandle {
         }
     }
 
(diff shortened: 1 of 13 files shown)


Bridgerz (approved):
lgtm

PR #841, opened 2026-07-18
Accumulate nonce dealer certs within a stamped time window past the `W−f` floor

zhouwfang:
MystenLabs/hashi#193 #515


packages/hashi/sources/core/mpc_config.move
```diff
@@ -21,12 +15,15 @@ const DEFAULT_MAX_FAULTY_IN_BASIS_POINTS: u64 = 3333;
 
 const VANILLA_NONCE_GENERATION_PROTOCOL: u64 = 0;
 
+const DEFAULT_NONCE_ACCUMULATION_WINDOW_MS: u64 = 700;
```
benr-ml: let's go with 2 sec?

packages/hashi/sources/core/tob.move
```diff
@@ -50,6 +43,7 @@ public struct DealerMessagesHashV1 has copy, drop, store {
 public struct DealerSubmissionV1 has copy, drop, store {
     message: DealerMessagesHashV1,
     signature: CommitteeSignature,
+    timestamp_ms: u64,
```
benr-ml, line 50: can we change this in testnet?

packages/hashi/sources/core/mpc_config.move
benr-ml: also, maybe let's add a doc saying that we wait DEFAULT_NONCE_ACCUMULATION_WINDOW_MS after the last published node before we complete collecting nonces

crates/hashi/src/mpc/types.rs
```diff
@@ -154,6 +156,75 @@ impl MpcConfig {
...
+    }
+
+    pub fn cutoff_ms(&self) -> Option<u64> {
+        match self.state {
+            NonceCollectionState::Window { cutoff_ms } => Some(cutoff_ms),
+            _ => None,
+        }
+    }
+
+    pub fn past_cutoff(&mut self, timestamp_ms: u64) -> bool {
```
benr-ml: nit - prefer this to be immutable since the name does not imply a mutation, nor it's critical for perf

benr-ml (approved):

packages/hashi/sources/core/tob.move
zhouwfang: No, testnet's package is frozen.

packages/hashi/sources/core/mpc_config.move
zhouwfang: @benr-ml One wording nit: the window is anchored once at the cert that crosses W−f, not at the last published cert — later arrivals don't extend the deadline, otherwise stragglers could drag collection out by up to f*DEFAULT_NONCE_ACCUMULATION_WINDOW_MS. No?

crates/hashi/src/mpc/types.rs
zhouwfang: I went further: past_cutoff is gone. It's now try_admit/record.

packages/hashi/sources/core/tob.move
bmwill: We can't even do an upgrade for this, we'll need to make a V2 struct

packages/hashi/sources/core/mpc_config.move
benr-ml: both options are fine imo

packages/hashi/sources/core/tob.move
benr-ml: note that some of the entry functions change in this pr as well - while those can indeed be upgraded since they are private entry, we probably need to be careful to not break backward compatibility of callers.

alternative to consider is to depend on checkpoints directly - fetch certs + their checkpoint inclusion time from the FN.

zhouwfang: Re "pulling checkpoint time from the FN": I had considered it too, but it doesn't survive recovery. Rebuilding a consumed batch needs each cert's stamp to re-derive the cutoff, and once the FN prunes those txs you can't re-date them.

I added `StampedDealerSubmissionV1` that just wraps the existing `DealerSubmissionV1` and adds the Clock stamp. (DKG/key rotation use `DealerSubmissionV1`, while nonce generation uses `StampedDealerSubmissionV1`.)

zhouwfang: I added a gate on package version to let the client run safely on both the current frozen package and the upgraded one, so we can merge this PR now and it will automatically flip when the move upgrade lands.

packages/hashi/sources/core/cert_submission.move
```diff
@@ -43,14 +38,15 @@ entry fun submit_nonce_cert(
     dealer: address,
     messages_hash: vector<u8>,
     cert: CommitteeSignature,
+    clock: &sui::clock::Clock,
```
benr-ml, line 41: how do we make sure all nodes call this function in the same epoch and not the old version?

packages/hashi/sources/core/tob.move
benr-ml: another option btw is to add a dynamic field to DealerSubmissionV1 with the timestamp (not sure which option is cleaner though)

benr-ml (approved):

packages/hashi/sources/core/tob.move
zhouwfang: Can't do it: DealerSubmissionV1 has copy + drop and UID has neither, so the struct can't hold one.

packages/hashi/sources/core/cert_submission.move
zhouwfang: Behavior isn't per-node: sizing and admission read the cert's own timestamp, so every node lands on the same answer from the same data. The gate just picks which shape we fetch and submit, and since one package version writes every cert on a given network, they're all the same shape anyway. You'd need an in-place upgrade to get a mix, which is out of scope here and will be done later.

benr-ml: sorry, i don't follow

zhouwfang: There's only one package deployed per network, so there's no old version to call: testnet has the v1 function (no clock), a fresh devnet publish has the v2 one. Both shapes only coexist after an in-place upgrade, which we're not doing yet.

PR #842, opened 2026-07-18
Fix the leader's spent-UTXO cleanup resubmitting cleaned sets forever

Bridgerz:
## Summary

Root-cause fix for the 7/17 gas-drain incident: at least four validators were each burning ~10 SUI/hour submitting successful **no-op** `withdraw::cleanup_spent_utxos` transactions in a tight loop (~45 tx/min, 16k txs/day from one address, 34 SUI/day).

## Root cause

Three-layer interaction:
1. `utxo_pool::cleanup_spent` emits no event, so a successful cleanup is invisible to the watcher.
2. The watcher never removes entries from the local `utxo_records` mirror (a comment claimed `UtxoSpent` removes them; it only marks them).
3. The leader's task-completion arm respawned the cleanup unconditionally — ungated by leadership, no backoff. First real cleanup succeeds → mirror still shows the records → scan rediscovers the same set → resubmit as a paid no-op, at transaction round-trip speed, forever.

## Fix — final design

One principle: **never pay for a cleanup the chain doesn't currently show as needed, and never write the shared mirror from outside the watcher.**

- The GC task builds a **task-local snapshot** of the on-chain `utxo_records` table (`scrape_utxo_records_snapshot`) — only that table, not the unbounded `spent_utxos` history — scans it for spent-but-uncleaned records (500/GC cap), submits, and returns the count. **The shared mirror is never written**; the watcher remains its sole writer.
- **Freshness is enforced on every response**: the `BitcoinState` wrapper read and each manually-paginated `ListDynamicFields` page must be at or past a monotonic height floor (seeded from the wrapper's `X_SUI_CHECKPOINT_HEIGHT`, folded with the re-sampled watcher cursor per page, raised by each accepted page). Any backward height bails into backoff — heights can only move forward through a scrape.
- **Scheduling**: spawn only from the leader-gated checkpoint arm, gated by an armed-scan flag (armed at boot, on withdrawal ConfirmedOnSui, and after a task that did work or failed), single-flight, with exponential failure backoff (5s → 5min cap, never giving up permanently — a drained wallet heals by top-up). A barren scan disarms until the next withdrawal confirms.

## Verification

- Adversarially reviewed across four passes (state-machine exhaustion, rotation races, failure-path cadence, mixed-fleet deployment): no unbounded gas or RPC path exists; every re-arm leads to a fresh chain read whose empty result disarms. Residuals are bounded: ≤2 no-op txs (~7.4 mSUI) per rotation/boot race episode; failure loops are gas-free (dry-run precedes signing).
- Unit tests cover the scan filter and cap; 74 utxo tests + clippy + fmt green.
- Known follow-up: no e2e exercises the leader GC loop end-to-end (pre-existing gap).

## Deployment

Ships alone against the current package (no Move changes). The fix is **per-process**: every validator must upgrade **and restart** — an unrestarted node keeps draining ~10 SUI/hr once it has held leadership after a withdrawal wave. Restart before topping up gas wallets (for upgraded nodes, top-up order is irrelevant). Serve validator RPC from a height-consistent source; the log signatures to alert on are repeated `Cleaning up spent UTXO(s) pending cleanup` with non-decreasing counts and repeated `stale UTXO pool read` errors.

### DIFF
--- a/crates/hashi/src/onchain/mod.rs
+++ b/crates/hashi/src/onchain/mod.rs
@@ -212,6 +212,14 @@ impl OnchainState {
         self.0.state.write().unwrap()
     }
 
+    /// Prune UTXO records that a cleanup transaction has removed on-chain.
+    /// `cleanup_spent_utxos` emits no event the watcher could react to, so
+    /// the leader that executed the cleanup applies its effect to the local
+    /// mirror through this call.
+    pub(crate) fn remove_cleaned_utxo_records(&self, utxo_ids: &[types::UtxoId]) {
+        self.state_mut().hashi.utxo_pool.remove_cleaned(utxo_ids);
+    }
+
     pub fn subscribe_checkpoint(&self) -> watch::Receiver<CheckpointInfo> {
         self.0.checkpoint.subscribe()
     }
--- a/crates/hashi/src/onchain/mod.rs
+++ b/crates/hashi/src/onchain/mod.rs
@@ -212,28 +212,33 @@ impl OnchainState {
         self.0.state.write().unwrap()
     }
 
-    /// Prune UTXO records that a cleanup transaction has removed on-chain.
-    /// `cleanup_spent_utxos` emits no event the watcher could react to, so
-    /// the leader that executed the cleanup applies its effect to the local
-    /// mirror through this call.
-    pub(crate) fn remove_cleaned_utxo_records(&self, utxo_ids: &[types::UtxoId]) {
-        self.state_mut().hashi.utxo_pool.remove_cleaned(utxo_ids);
-    }
-
-    /// Re-scrape the on-chain UTXO pool into the local mirror. The cleanup
-    /// GC calls this before submitting an orphan-scan batch: the mirror can
-    /// overstate pending cleanups (another leader's cleanup emits no event
-    /// the watcher could apply), and every overstated id becomes a paid
-    /// on-chain no-op. One table scan here is cheaper than paying to
-    /// re-clean records that are already gone.
-    pub(crate) async fn refresh_utxo_pool(&self) -> Result<()> {
+    /// Scrape the on-chain UTXO pool into a task-local snapshot for the
+    /// cleanup GC's use. Deliberately does NOT touch the shared mirror: the
+    /// watcher task is its sole writer, and installing an out-of-band
+    /// snapshot could revert mutations the watcher already applied from
+    /// newer checkpoints (e.g. a `spent_by` lock, whose loss would let coin
+    /// selection re-pick a locked input and pay for an aborted tx).
+    ///
+    /// The GC scans this snapshot instead of the mirror because the mirror
+    /// can overstate pending cleanups (`cleanup_spent_utxos` emits no event
+    /// the watcher could apply, so another leader's cleanup is invisible),
+    /// and every overstated id becomes a paid on-chain no-op. Errors if the
+    /// read is older than the watcher's checkpoint cursor (e.g. a
+    /// load-balanced endpoint serving a lagging fullnode) — the caller
+    /// should back off and retry rather than act on stale data.
+    pub(crate) async fn scrape_utxo_pool_snapshot(&self) -> Result<types::UtxoPool> {
         let client = self.client();
-        let bitcoin_state =
+        let (scrape_height, bitcoin_state) =
             fetch_bitcoin_state(client.clone(), self.hashi_id(), self.package_id_original())
                 .await?;
-        let utxo_pool = scrape_utxo_pool(client, bitcoin_state.utxo_pool).await?;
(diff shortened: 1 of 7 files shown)


bmwill (approved):

Bridgerz:
Both review inputs addressed in `5164e3e`, which supersedes the watermark approach with the stronger design:

**The GC now decides from a task-local chain snapshot and never writes the shared mirror.** `refresh_utxo_pool` (install-into-mirror) is gone; `scrape_utxo_pool_snapshot` returns the pool to the GC task, which scans it and submits. The watcher is back to being the mirror's sole writer, so the snapshot-older-than-cursor scenario (reverted `spent_by` locks, vanishing pending-change records) is impossible rather than guarded — there is no install to guard.

The freshness check survives in the form that still matters: the scrape bails into the existing backoff when its `X_SUI_CHECKPOINT_HEIGHT` is behind the watcher's cursor, so cleanup decisions only ever use data at least as fresh as anything the node has seen. Bailing (rather than skipping and proceeding) matters: falling back to the stale mirror would reintroduce pay-on-stale-belief through the side door.

One residual, considered and accepted: the pool table scrape is paginated, so pages can span checkpoints even when the first response is fresh — a torn view can still produce one idempotent no-op id (or miss one, caught next scan). That's the boot scrape's existing exposure and bounded, so per-page watermarking would be gold-plating.

Also net −67 lines vs. the previous revision: the leader-side mirror prune (`remove_cleaned`) is deleted entirely — mirror hygiene now arrives only via the `UtxoCleaned` event in #843, through the watcher.

crates/hashi/src/onchain/mod.rs
```diff
@@ -212,6 +212,35 @@ impl OnchainState {
...
+            fetch_bitcoin_state(client.clone(), self.hashi_id(), self.package_id_original())
+                .await?;
+        let watcher_height = self.latest_checkpoint_height();
+        if scrape_height < watcher_height {
+            anyhow::bail!(
+                "stale UTXO pool read: scrape at checkpoint {scrape_height} is behind the \
+                 watcher's cursor {watcher_height}"
+            );
+        }
+        scrape_utxo_pool(client, bitcoin_state.utxo_pool).await
```
dlukegordon: The freshness check only covers the preceding BitcoinState request, not the paginated list_dynamic_fields requests that supply the cleanup IDs. A current parent response followed by stale pages can still trigger paid no-op cleanups. Please paginate manually, retain each page’s checkpoint metadata, and reject/retry pages older than the watcher cursor or inconsistent with the first page.

dlukegordon: Also, this calls scrape_utxo_pool, which also loads the complete, ever-growing spent_utxos history, although cleanup only uses utxo_records. Please use a cleanup-specific scraper that fetches only utxo_records.

Bridgerz: Ok take another look

crates/hashi/src/onchain/mod.rs
```diff
@@ -212,6 +212,88 @@ impl OnchainState {
...
+                .ok_or_else(|| anyhow!("response missing X_SUI_CHECKPOINT_HEIGHT header"))?;
+            // Re-sample the cursor: it advances while we paginate, and a
+            // page must not predate anything this node has already seen.
+            let watcher_height = self.latest_checkpoint_height();
+            if page_height < watcher_height {
+                anyhow::bail!(
+                    "stale UTXO pool read: page at checkpoint {page_height} is behind the \
+                     watcher's cursor {watcher_height}"
+                );
+            }
```
dlukegordon, line 312: This still allows checkpoint heights to move backward: if the wrapper is at 105 and the watcher at 100, a page at 102 passes despite being stale relative to the wrapper. Please track a monotonic minimum initialized to scrape_height, then require each page to be at least max(previous_page_height, latest_watcher_height).

Bridgerz: Monotonic floor seeded from the wrapper's height, folded with the resampled watcher cursor per page, raised by each accepted page... good?

dlukegordon (approved):

crates/hashi/src/onchain/mod.rs
```diff
@@ -233,36 +238,56 @@ impl OnchainState {
...
             let mut request = ListDynamicFieldsRequest::default()
                 .with_parent(utxo_records_id)
                 .with_page_size(SCRAPE_PAGE_SIZE)
-                .with_read_mask(FieldMask::from_paths([DynamicField::path_builder()
-                    .value()
-                    .finish()]));
+                .with_read_mask(FieldMask::from_paths([
+                    // `name` is load-bearing even though we only read
+                    // `value`: the fullnode's `should_load_field` does not
+                    // recognize a value-only mask and would return fields
```
bmwill, line 285: if this is true we should fix it in the fullnode. but no harm with this for now

PR #843, opened 2026-07-18
Emit UtxoCleaned from cleanup_spent so every node prunes its mirror

Bridgerz:
## Summary

Stacked on #842. The base fix prunes the local `utxo_records` mirror only on the **leader that executed** the cleanup transaction. Every other validator's mirror keeps the stale records, so each would fire one redundant (idempotent, paid) cleanup on its first leadership turn after any withdrawal wave.

## Changes

- `utxo_pool.move`: emit `UtxoCleaned { utxo_id, spent_epoch }` from the real-cleanup branch of `cleanup_spent`. The no-op branch stays silent, so replayed cleanups emit nothing.
- `hashi-types`: `UtxoCleaned` struct + `HashiEvent` variant + parse arm (module `utxo_pool`, mirrors `UtxoSpent`).
- Watcher: on `UtxoCleaned`, prune via the same `UtxoPool::remove_cleaned` helper (no-ops on the leader that already pruned directly).

## Tests

Move tests in `tests/utxo_spent_tests.move`: a real cleanup emits exactly one event with the right fields; a no-op cleanup emits none.

## Deployment

Takes effect with the next testnet package upgrade. Mixed-version windows are safe in both directions: new binaries against the old package simply see no event (base-PR behavior), and old binaries against the new package ignore unknown events in `HashiEvent::try_parse`.

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -841,6 +841,7 @@ pub enum HashiEvent {
     WithdrawalPresigsReassigned(WithdrawalPresigsReassigned),
     WithdrawalConfirmed(WithdrawalConfirmed),
     UtxoSpent(UtxoSpent),
+    UtxoCleaned(UtxoCleaned),
     ReconfigStarted(ReconfigStarted),
     ReconfigEnded(ReconfigEnded),
 }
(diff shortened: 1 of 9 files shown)


Bridgerz:
Pushed review follow-ups (rebased on the revised #842): the watcher now applies the event's **own** `spent_epoch` unconditionally via a new `UtxoPool::apply_cleaned` — the previous delegation to `remove_cleaned` skipped the tombstone when the local record had never seen its `UtxoSpent` (leaving a window where a spent bridge UTXO was in neither local table), and updated the three comments that still claimed cleanup emits no event.

Two verified-safe notes from the review worth having on record: mixed versions are safe in both directions (old binaries ignore unknown events via the `try_parse` fallthrough; new binaries on the old package just see no events), and a `UtxoCleaned` landing in the *same checkpoint* as the package-upgrade tx is dropped by the package-id snapshot — harmless, since the next cleanup scan re-reads chain state anyway.

bmwill:
I think we no longer really need this since we now track updates without relying on events

PR #844, opened 2026-07-18
Add a hashi_sui_balance metric for the operator gas wallet

Bridgerz:
## Summary

Requested by a validator during the GC gas-drain incident (#842): the gas wallet drained with zero observability, and `sui client gas` reads empty even for a funded wallet because the node sweeps coins into the Sui **address balance** at startup.

## Changes

- New service in `lib.rs` polls `StateService/GetBalance` every 60s for the operator address (the response's `balance` field totals owned coins **plus** address balance) and sets `hashi_sui_balance` (MIST).
- If no operator key is configured, the service logs and exits cleanly.
- Runbook: metric table row + suggested alert (`hashi_sui_balance < 1e9`), since the same wallet signs deposit approvals, withdrawal certs, and MPC/reconfig txs — an empty wallet means dead leadership windows.

### DIFF
--- a/crates/hashi/src/lib.rs
+++ b/crates/hashi/src/lib.rs
@@ -837,6 +837,7 @@ impl Hashi {
         let backup_service = backup_service.start();
         let mpc_service = mpc_service.start();
         let guardian_bootstrap_service = self.clone().start_guardian_bootstrap();
+        let sui_balance_service = self.clone().start_sui_balance_metric();
 
         let service = Service::new()
             .merge(onchain_service)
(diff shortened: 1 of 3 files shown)


Bridgerz:
Pushed review follow-ups: the gauge now initializes to `-1` (never sampled) so the suggested low-balance alert can't false-fire on nodes without an operator key or before the first successful poll — a registered-but-unset IntGauge exports 0, which is indistinguishable from a genuinely drained wallet. The documented alert is gated on `>= 0` accordingly, and the request is built with `with_owner`/`with_coin_type` + `StructTag::sui()` to match the existing call sites. Verified against the pinned SDK that `Balance.balance` totals owned coins **plus** the address balance, which matters because the node sweeps gas into the address balance at startup.

jessiemongeon1:
## Style Guide Audit

bmwill (approved):

PR #845, opened 2026-07-18
Emit LaunchCompleted from finish_publish; refresh config on version proposals

Bridgerz:
## Summary

Stacked on #843 (second in the package-upgrade batch — **not for the current binary release**; #842 ships alone now, this lands with the upcoming package upgrade).

Closes out the two config-staleness gaps from this week's incidents:

1. **`LaunchCompleted` event** — `finish_publish` writes `guardian_url` and the guardian BTC key directly into config with no event, which is the root cause of the 2026-07-16 launch stall (nodes booted pre-launch kept an empty snapshot, derived no deposit addresses, and terminally rejected every deposit). The event carries the pinned guardian fields; the watcher responds by re-fetching the config snapshot and kicking the guardian reconcile so the BTC-key pin lands immediately instead of waiting for an unrelated guardian interaction. The #839 launch poll stays as the fallback for pre-event packages.
2. **Version proposals refresh the snapshot** — the `ProposalExecuted` refresh arm covered only `UpdateConfig | EmergencyPause | UpdateGuardian`, but `types::Config` also carries `enabled_versions` and `upgrade_cap`: executing `EnableVersion`/`DisableVersion`/`Upgrade` left every stable-stream node with stale version gating until a reconnect. This half is pure watcher code and needs no package upgrade — those proposals already emit `ProposalExecuted`.

## Tests

- Move: `finish_publish` emits exactly one `LaunchCompleted` with the pinned fields (full Move suite: 151 passing).
- Mixed-version safety: identical to #843 — old binaries ignore the unknown event via the `try_parse` fallthrough; new binaries on the old package simply never see it and rely on the existing poll.

## Deployment

Rollout order: #842 binary now → package upgrade with #843 + this PR → the `EnableVersion` proposal that activates the new version is itself the first beneficiary of change 2.

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -842,6 +842,7 @@ pub enum HashiEvent {
     WithdrawalConfirmed(WithdrawalConfirmed),
     UtxoSpent(UtxoSpent),
     UtxoCleaned(UtxoCleaned),
+    LaunchCompleted(LaunchCompleted),
     ReconfigStarted(ReconfigStarted),
     ReconfigEnded(ReconfigEnded),
 }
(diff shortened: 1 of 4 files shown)


bmwill:
I think this is not needed now due to the changes in how we track onchain state

PR #847, opened 2026-07-20
watcher: Rescrape the full chain state every 10 minutes

bmwill:
Previously, the watcher only rescraped on-chain state after a reconnect, so the local mirror could drift indefinitely from chain state for writes that emit no event (such as withdraw::cleanup_spent_utxos, whose stale mirror entries caused the recent gas-drain incident).

With this commit, the watcher unconditionally rescrapes the full state every 10 minutes. The rescrape runs on the watcher task itself, so it is serialized with event application and cannot race the checkpoint stream; any checkpoints the scrape already included simply replay as no-ops. The reconnect-rescrape body is extracted into a shared rescrape_hashi_state helper, and a successful periodic rescrape also resets the config-poll clock since the full scrape includes config. A failed periodic rescrape only warns and waits for the next interval, as the reconnect rescrape still guards correctness.

### DIFF
--- a/crates/hashi/src/onchain/watcher.rs
+++ b/crates/hashi/src/onchain/watcher.rs
@@ -37,6 +37,12 @@ use crate::withdrawals::withdrawal_limiter_consumption_amount;
 /// server silently stopped sending checkpoints.
 const CHECKPOINT_STREAM_STALL_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(120);
 
+/// How often to unconditionally rescrape the full on-chain state. The event
+/// stream can drift from chain state (e.g. writes that emit no event, such as
+/// `cleanup_spent_utxos`); a periodic rescrape bounds how long any such drift
+/// can persist.
+const PERIODIC_RESCRAPE_INTERVAL: std::time::Duration = std::time::Duration::from_secs(10 * 60);
+
 #[tracing::instrument(name = "watcher", skip_all)]
 pub async fn watcher(sui_rpc_url: String, state: OnchainState, metrics: Option<Arc<Metrics>>) {
     let subscription_read_mask = FieldMask::from_paths([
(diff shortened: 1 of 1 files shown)


crates/hashi/src/onchain/watcher.rs
```diff
@@ -203,6 +207,30 @@ pub async fn watcher(sui_rpc_url: String, state: OnchainState, metrics: Option<A
...
+    metrics: Option<&Arc<Metrics>>,
+) -> anyhow::Result<()> {
+    let (checkpoint_info, hashi) = super::scrape_hashi(
+        client.clone(),
+        state.hashi_id(),
+        state.package_id_original(),
+    )
+    .await?;
+    state.replace_hashi_state(hashi);
+    state.update_latest_checkpoint_info(checkpoint_info);
```
Bridgerz, line 225: If the scrape lands on a lagging replica, this installs a snapshot older than what the stream already applied. `update_latest_checkpoint_info` is a bare send_replace and replace_hashi_state swaps wholesale, so we rewind and stay rewound (those events won't re-arrive) until the next event or interval. A regressed watch height also weakens the stale UTXO pool read guard in scrape_utxo_pool_snapshot, which is the exact bug class this PR is fighting. 

Suggest: compare the scrape's checkpoint height against the current watch value and skip the whole install when the scrape is behind. (Window existed pre-PR on reconnect, but every-10-min makes it routine.)

crates/hashi/src/onchain/watcher.rs
```diff
@@ -194,6 +184,20 @@ pub async fn watcher(sui_rpc_url: String, state: OnchainState, metrics: Option<A
             if let Some(metrics) = &metrics {
                 metrics.update_onchain_state(&state);
             }
+
+            // Unconditional periodic rescrape. Runs on the watcher task, so
+            // it can't race event application; any checkpoints the scrape
+            // already included are replayed by the stream afterward, which
+            // re-applies them as no-ops. On failure, just wait for the next
+            // interval — the reconnect rescrape still guards correctness.
+            if last_rescrape.elapsed() >= PERIODIC_RESCRAPE_INTERVAL {
```
dlukegordon, line 193: If `WithdrawalSigned` lands during the scrape, the snapshot may already mark it signed, causing replay to skip limiter consumption. Could we use first insertion into `withdrawal_signed_at_ms` as the idempotency guard instead of `txn.is_fully_signed()`? That map survives snapshot replacement, so the event consumes exactly once while actual redelivery remains a no-op.

bmwill: this should be generally ok since we also resnap limits at some cadence as well

dlukegordon (approved):

PR #848, opened 2026-07-20
[Guardian] Make genesis bootstrap KP-authorized

mskd12:
OperatorInit now pins an optional GenesisState separately from InitConfig, and GuardianInfo exposes its optional hash. Key provisioners independently derive that hash and bind it into the normal signed ProvisionerInit submission alongside the session and config hash; the relay and enclave enforce both pins.

On first deploy, threshold PI writes genesis/record.json with the pinned committee after authoritatively verifying the KP-signed submissions. The operator and each KP must pass --do-genesis purely as an explicit intent marker, and callers fail when that intent disagrees with the observed S3 committee state.


jessiemongeon1:
## Style Guide Audit

crates/hashi-guardian-init/src/main.rs
```diff
@@ -56,6 +56,9 @@ enum OperatorCommand {
         /// Path to operator provision YAML config file.
         #[arg(long)]
         config: PathBuf,
+        /// Explicitly bootstrap the first serving committee during PI.
+        #[arg(long)]
+        do_genesis: bool,
```
dlukegordon, line 61: I think we need to update docker/hashi-guardian-local/scripts/provision.sh to use this?

mskd12: Good catch — updated the local genesis provisioning flow to pass `--do-genesis` to both `operator provision` and every `key-provisioner provision` invocation. Included in 938f7e1c.

dlukegordon (approved):

PR #851, opened 2026-07-20
Save encrypted share record after KP ceremony

dlukegordon:


crates/hashi-types/src/guardian/log/envelope.rs
```diff
@@ -245,6 +245,13 @@ impl LogRecord {
         self,
         pub_key: &GuardianPubKey,
     ) -> GuardianResult<(SessionID, UnixMillis, LogMessage)> {
+        self.verify_preserving_envelope(pub_key)?;
+
+        Ok((self.session_id, self.timestamp_ms, self.message))
+    }
+
+    /// Verify this signed record without consuming its S3 envelope.
+    pub fn verify_preserving_envelope(&self, pub_key: &GuardianPubKey) -> GuardianResult<()> {
```
mskd12: Naming is a little weird. How about: `verify` → `into_verified_parts` and `verify_preserving_envelope` → `verify`?

crates/hashi-guardian/src/s3_reader.rs
```diff
@@ -382,21 +396,27 @@ impl GuardianSessionCache {
...
+        Ok(VerifiedLogRecord::new(
+            record.object_key,
+            record.session_id,
+            record.timestamp_ms,
+            record.message,
+            build_pcrs,
+        ))
+    }
+
+    async fn verify_record_preserving_envelope(
```
mskd12: How about `verify_record` → `into_verified_record` and `verify_record_preserving_envelope` → `verify_record`?

mskd12 (changes requested):
Why store signatures alongside the encrypted shares? A corrupt file will be detected at the time of `kp provision` when you compare with the whose share commitment. So let's keep it simple and store just the CeremonyState?

crates/hashi-guardian/src/s3_reader.rs
mskd12: Also, why is it returning what it takes as input? Caller just needs &LogRecord, no?

dlukegordon: Yeah that's dumb, it's just a copied pattern from `verify_record`, will fix.

dlukegordon:
The idea was that the file we save to disk should be the same as what's on s3. So that way, if for some reason the bucket every got wiped, we could copy it over more easily. But I'm fine changing it to be more simple.

mskd12 (approved):

PR #860, opened 2026-07-21
Make presig recovery pending-aware with cert-derived batch sizing

zhouwfang:
MystenLabs/hashi#193

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -40,6 +40,7 @@ pub use crate::mpc::types::MpcError;
 pub use crate::mpc::types::MpcOutput;
 use crate::mpc::types::MpcOutputRecoveryOutcome;
 pub use crate::mpc::types::MpcResult;
+use crate::mpc::types::NonceCertToVerify;
 use crate::mpc::types::NonceGenerationProtocol;
 pub use crate::mpc::types::NonceMessage;
 pub use crate::mpc::types::NonceReconstructionOutcome;
@@ -1095,6 +1096,38 @@ impl MpcManager {
             }
         }
         (certified, weight_sum)
+    }
+
+    pub(crate) async fn verified_nonce_certs<T>(
+        mpc_manager: &Arc<RwLock<Self>>,
+        epoch: u64,
+        certs: Vec<(Address, T)>,
+    ) -> Vec<(Address, T)>
+    where
+        T: NonceCertToVerify,
+    {
+        let mut verified = Vec::with_capacity(certs.len());
+        for (dealer, cert) in certs {
+            let dealer_cert = match cert.to_dealer_certificate(epoch) {
+                Ok(dealer_cert) => dealer_cert,
+                Err(e) => {
+                    tracing::info!("recovery: dropping malformed nonce cert from {dealer:?}: {e}");
+                    continue;
+                }
+            };
+            let mgr = Arc::clone(mpc_manager);
+            let verification = spawn_blocking(move || {
+                mgr.read().unwrap().committee.verify_signature(&dealer_cert)
+            })
+            .await;
+            match verification {
+                Ok(()) => verified.push((dealer, cert)),
+                Err(e) => tracing::info!(
+                    "recovery: dropping nonce cert with invalid signature from {dealer:?}: {e}"
+                ),
+            }
+        }
+        verified
     }
 
     async fn run_dkg_as_dealer(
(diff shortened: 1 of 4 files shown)


crates/hashi/src/mpc/service.rs
```diff
@@ -619,6 +746,19 @@ impl MpcService {
                 info!("sync_if_stale: recovered epoch {epoch} via certs path");
             }
             Err(certs_err) => {
```
dlukegordon, line 756: Recovery failures still fall back to generating a fresh batch 0. If presigs were already allocated, this recreates the original divergence between on-chain assignments and local nonce material. Please only allow this fallback when num_consumed_presigs == 0 and there are no pending assignments; otherwise retry certified recovery.

zhouwfang: Could you look at the most recent version? This had been addressed before you commented.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1076,10 +1076,10 @@ impl MpcManager {
         ))
     }
 
-    fn certified_nonce_dealers_from_certs(
+    pub(crate) fn certified_nonce_dealers_from_certs<T>(
```
dlukegordon, line 1080: These certificates are counted without verifying their signature or signer quorum, but on-chain submission defers verification. A Byzantine dealer can therefore alter the recovered dealer prefix and batch size even though live nonce processing would reject its certificate. Can we reuse the live certificate-validation path before accumulating dealer weight?

crates/hashi/src/mpc/service.rs
```diff
@@ -530,66 +531,194 @@ impl MpcService {
...
+        let recovered_end = boundaries
+            .last()
+            .map_or(0, |&(_, start, size)| start + size as u64);
+        let first_pending = boundaries
+            .iter()
+            .position(|&(_, start, size)| {
+                let end = start + size as u64;
+                pending.iter().any(|&p| p >= start && p < end)
+            })
+            .unwrap_or_else(|| boundaries.len().saturating_sub(1));
```
dlukegordon, line 615: When cursor is zero and batch 0 has no certificates, boundaries is empty but recovery still installs an empty SigningManager. That suppresses genesis fallback and causes refill to start at batch 1, leaving index 0 unusable. Please reject empty recovery state and let cursor-zero recovery generate batch 0.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -5352,6 +5353,75 @@ impl MpcManager {
...
+                         during recovery"
+                    ))
+                })?;
+            let Messages::NonceGeneration(ref nonce) = messages else {
+                return Err(MpcError::ProtocolFailed(format!(
+                    "Retrieved non-nonce message for dealer {dealer:?} during recovery"
+                )));
+            };
+            let mut mgr = mpc_manager.write().unwrap();
+            let epoch = mgr.mpc_config.epoch;
```
dlukegordon, line 5452: After retrieving a certified message, we also need to invalidate dealer_nonce_outputs[(batch_index, dealer)]. Otherwise reconstruction at line 1051 can reuse output derived from the stale hash-mismatching message. The normal party path already performs this invalidation at lines 1809–1815.

crates/hashi/src/mpc/service.rs
```diff
@@ -682,14 +824,21 @@ impl MpcService {
                             "No nonce gen certificates on TOB for epoch {epoch} batch {batch_index}"
                         )
```
dlukegordon, line 845: Phase 2 refetches certificates but does not recheck the required weight before calling presig_count(). A shorter partial fetch can underflow total_weight - consumed. Please reuse the validated Phase 1 certificates, or revalidate the floor and use checked subtraction here.

crates/hashi/src/mpc/signing.rs
```diff
@@ -1330,6 +1397,137 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_new_recovered_retains_pending_batch_and_gates_slots() {
```
dlukegordon, line 1401: This test covers slot masking but not the incident lifecycle. Please add Vanilla and AVID e2e coverage where a withdrawal retains pending indices from an older batch, all nodes restart after the cursor advances, and the same withdrawal completes through Bitcoin confirmation.

crates/hashi/src/mpc/service.rs
bmwill: I think this is the one issue that we probably need to fix before landing, the others likely could be defered

crates/hashi/src/mpc/signing.rs
zhouwfang: I considered this, and it's racy to reproduce reliably. I'll think more about how to test it later.

dlukegordon (approved):

PR #861, opened 2026-07-22
DNS - comments on DKG

benr-ml:


crates/hashi-types/src/committee.rs
```diff
@@ -34,11 +34,14 @@ pub use crate::move_types::DEFAULT_MPC_WEIGHT_REDUCTION_ALLOWED_DELTA;
 pub use crate::move_types::VANILLA_MPC_NONCE_GENERATION_PROTOCOL;
 
 // TODO: Read threshold from on-chain config once it is made configurable.
+// Ben: do we need this?
```
zhouwfang, line 37: This was already fixed in main: https://github.com/MystenLabs/hashi/blob/d4bb24498c7e1c1ca7f244e1ad8ce02b2eae3df6/crates/hashi-types/src/committee.rs#L37-L38

crates/hashi-types/src/committee.rs
```diff
@@ -34,11 +34,14 @@ pub use crate::move_types::DEFAULT_MPC_WEIGHT_REDUCTION_ALLOWED_DELTA;
...
 const THRESHOLD_NUMERATOR: u64 = 2;
 const THRESHOLD_DENOMINATOR: u64 = 3;
 
 /// Matches Move's `threshold::certificate_threshold`.
+// Ben: shouldn't this be dynamic base on f & t?
```
zhouwfang, line 42: This one is for withdrawal/deposit/reconfig-completion certs. Move verifies these on-chain with the same fixed 6667bps, so the Rust side has to mirror it. The t/f-based thresholds live in the MPC cert path: https://github.com/MystenLabs/hashi/blob/d49f7a1148cce2aef9b6c82fd00d65c525e1dc5b/packages/hashi/sources/core/mpc_config.move#L26-L27

crates/hashi-types/src/committee.rs
```diff
@@ -34,11 +34,14 @@ pub use crate::move_types::DEFAULT_MPC_WEIGHT_REDUCTION_ALLOWED_DELTA;
...
 pub fn certificate_threshold(total_weight: u64) -> u64 {
+    // Ben: this should be ceil, not floor (10000 -> 6667)
```
zhouwfang, line 44: Responded in https://github.com/MystenLabs/hashi/pull/861/changes#r3669875417

crates/hashi-types/src/committee.rs
```diff
@@ -293,6 +296,9 @@ impl Committee {
     /// Verify an [CommitteeSignature]. If you also need to verify the weight, you can either
     /// get the weight of the signature with [CommitteeSignature::weight] or use the [Self::verify_signature_and_weight]
     /// function.
+    
+    // Ben: Scary as a public function: do we really need this to be public? all calls should either check a signer or a weighted cert
+    // also, should we have here a reduced weights verify?
```
zhouwfang, line 301: Re "all calls should either check a signer or a weighted cert": the legit raw callers are wrappers that do their own weight gate in a domain committee can't know (e.g. the AVID nonce loop verifies the signature, then separately checks the signers' reduced weight against per-kind quorums. Those weights and thresholds live in the MPC layer, not in committee).

It can't be non-public since the callers are in the hashi crate. I renamed it to `verify_signature_any_weight`  in #898 instead so the missing quorum check is obvious at call sites.

Yes on reduced weights. That check belongs in the MPC layer (committee doesn't know reduced weights). Today only the AVID path checks weight on read; adding it everywhere else is coming in a separate PR.

crates/hashi-types/src/committee.rs
benr-ml: @zhouwfang just making sure i follow - do you mean for protocol-level / non-mpc certs, we use 2/3 by default, right?

crates/hashi-types/src/committee.rs
benr-ml: let's find a safer solution - e.g., why callers cannot pass the weights/thresholds if needed
cc @jonas-lj

jonas-lj: We can add a function, `verify_signature_with_weights`, which takes a new set of weights instead of using the ones in the committee? It's not ideal since its up to the committee to handle the weights here, but it makes the API safer.

crates/hashi-types/src/committee.rs
zhouwfang: Yes, non-MPC certs use the fixed >2/3 quorum mirroring move. Only MPC certs use t/f.

crates/hashi-types/src/committee.rs
zhouwfang: SGTM with two tweaks though: pass the Nodes rather than a weights map, which also lets it verify the weights actually belong to this committee (it checks each signer's node key against the member's encryption key — a map, or just a node-count check, can't tell current from previous when they're the same size). and it returns the achieved weight rather than just ok/err, since AVID only knows the cert kind after verifying.

Once the party loops and recovery paths move onto it, the only caller left outside hashi-types is `AvidCertificate::verify` (signature-only by design), with its weight gate in front. So one call site rather than fully private.

zhouwfang: PTAL at #898 as step 1.

crates/hashi-types/src/committee.rs
benr-ml: got it, let's add a comment above those const and say that those are the threshold for non-mpc certs

crates/hashi-types/src/committee.rs
zhouwfang: Followup: #919. Only caller of verify_signature_any_weight left after that is AvidCertificate::verify, signature-only by design with the weight gate in front. So one call site rather than private, as above.

crates/hashi-types/src/committee.rs
```diff
@@ -544,6 +550,8 @@ pub struct BlsSignatureAggregator<'a, T> {
     bitmap: BitMap,
     signed_weight: u64,
     message: T,
+    // Ben: This struct is very confusing, we either use it with signed_weight OR (signed_reduced_weight, reduced_weights), but never both.
+    // It should be an enum and simpler accessors.
```
zhouwfang, line 554: PTAL at #927

crates/hashi-types/src/committee.rs
zhouwfang: Added in #927.

crates/hashi/src/communication/sui_tob.rs
```diff
@@ -46,6 +46,7 @@ impl From<TobError> for ChannelError {
     }
 }
 
+// Ben: It is not a general channel, but per protocol - maybe call it SuiTobChannelPerProtocol
```
zhouwfang, line 49: PTAL at #928. I renamed it to be SuiTobSessionChannel.

crates/hashi/src/communication/sui_tob.rs
```diff
@@ -172,6 +173,7 @@ impl OrderedBroadcastChannel<CertificateV1> for SuiTobChannel {
         .await
         .map_err(ChannelError::from)?;
         if existing.iter().any(|(d, _)| *d == dealer) {
+            // Ben: check it is the same cert or warn
```
zhouwfang, line 176: Done in #934.

crates/hashi/src/mpc/service.rs
```diff
@@ -321,6 +321,7 @@ impl MpcService {
             .inner
             .mpc_manager()
             .ok_or_else(|| anyhow::anyhow!("MpcManager not initialized for DKG recovery"))?;
+        // Ben: we didnt verify the certs (maybe we should define VerifiedCertificateV1 type to enforce this?)
```
zhouwfang, line 324: Fixed in #919.

crates/hashi/src/mpc/service.rs
```diff
@@ -377,6 +378,7 @@ impl MpcService {
                 .into_iter()
                 .map(|(_, cert)| cert)
                 .collect();
+        // Ben: we didnt verify the certs
```
zhouwfang, line 381: Fixed in #919.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1077,13 +1076,14 @@ impl MpcManager {
...
+    fn certified_nonce_dealers_from_certs(
         &self,
-        certs: &[(Address, T)],
-    ) -> (HashSet<Address>, u32) {
+        certs: &[(Address, hashi_types::move_types::DealerSubmissionV1)],
+    ) -> HashSet<Address> {
         let required_weight = self.required_nonce_weight();
         let mut weight_sum = 0u32;
         let mut certified = HashSet::new();
+        // Ben: we didnt verify the certs
```
zhouwfang, line 1086: Fixed by #935.

crates/hashi/src/mpc/service.rs
```diff
@@ -284,6 +283,7 @@ impl MpcService {
             .map_err(|e| {
                 anyhow::anyhow!("failed to fetch reconfig certs for epoch {epoch}: {e}")
             })?;
+        // Ben: we didnt verify the certs...
```
zhouwfang, line 286: Fixed by #935.

crates/hashi/src/communication/timeout_and_retry.rs
```diff
@@ -13,6 +13,8 @@ use std::time::Duration;
 use sui_sdk_types::Address;
 
 // TODO: Use lower thresholds for unit tests.
+// Ben: should we continue to retry for longer? why stop after 10 retries? i suppose it depends on the context:
+// For dealer needing t+f responses, we should retry for longer, but afterwards wait just a bit more.
```
zhouwfang, line 17: See #940

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1163,6 +1131,7 @@ impl MpcManager {
             .mpc_p2p_broadcast_duration_seconds
             .with_label_values(&[MPC_LABEL_DKG])
             .start_timer();
+        // Ben: join_all? why not wait for t+f + timeout? or at least have fewer retries?
```
zhouwfang, line 1134: Fixed by #940.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1796,7 +1765,7 @@ impl MpcManager {
                 let cert = nonce_cert.clone();
                 let verified = spawn_blocking(move || {
                     let mgr = mgr.read().unwrap();
-                    mgr.committee.verify_signature(&cert)
+                    mgr.committee.verify_signature(&cert) // Ben: no weights?
```
zhouwfang, line 1768: Fixed by #919

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -4718,6 +4687,7 @@ impl MpcManager {
         certificates: &[CertificateV1],
         complaint_cache: &HashMap<DealerOutputsKey, avss::AvssOutput>,
     ) -> MpcResult<ReconstructionOutcome> {
+        // Ben: why not use base_session_id_for_epoch? we need more consistency with session ids
```
zhouwfang, line 4690: Fixed in #956.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -4746,6 +4717,7 @@ impl MpcManager {
                 })?;
             let messages = Messages::Dkg(message.clone());
             let actual_hash = compute_messages_hash(&messages);
+            // Ben: what if we stored a message from a dealer but a different message was certified? we should retrieve the right message
```
zhouwfang, line 4720: Fixed by #945

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -4734,6 +4704,7 @@ impl MpcManager {
             };
             let msg = dkg_cert.message();
             let dealer_address = msg.dealer_address;
+            // Ben: what about the case we see a cert for a message we don't have?
```
zhouwfang, line 4707: @benr-ml It's expected rather than a bug. We stop collecting at t+f and drop the remaining sends, so a node can end up without the message.This is the restart path that tries to rebuild the output locally from stored messages instead of re-running the protocol. When one's missing we give up on that and fall back to the full run_dkg, which pulls it from the cert's signers. Costs a re-run, and if no signer still has it we error out and retry.

Probably worth adding a repair step here. The previous-epoch path already has one.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -4759,7 +4731,7 @@ impl MpcManager {
                 as PartyId;
             let session_id = source_session_id
                 .dealer_session_id(&dealer_address)
-                .to_vec();
+                .to_vec(); // Ben: why?
```
zhouwfang, line 4734: Fixed in #965

crates/hashi/src/mpc/mpc_except_signing.rs
benr-ml: got it. let's at least document that in this function (say that this function succeed only if it the current node already has all dkg messages), or even rename it from reconstruct_dkg_output to reconstruct_dkg_output_locally

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -5576,7 +5476,7 @@ impl MpcManager {
 
     fn base_session_id_for_epoch(&self, epoch: u64, protocol_type: &ProtocolType) -> SessionId {
         if epoch == self.mpc_config.epoch {
-            self.session_id.clone()
+            self.session_id.clone() // Ben: but this one might have used a different protocol type
```
zhouwfang, line 5479: Fixed in #956

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -5891,28 +5792,28 @@ fn process_avss_message(
     nodes: Nodes<EncryptionGroupElement>,
     party_id: u16,
     params: Parameters,
-    session_id: Vec<u8>,
+    session_id: Vec<u8>, // Ben: should be SessionId
```
zhouwfang, line 5795: Fixed in 965

crates/hashi/src/mpc/types.rs
```diff
@@ -83,6 +83,7 @@ pub struct AvidRoundState {
 pub(crate) type HeldAvidEchoes = (batch_avss_avid::AvidVote, Vec<(Address, Messages)>);
 
 // Domain separation constants for RandomOracle
+// Ben: what is 754526047e6e997e6c348e7c3491c57b79e22c3efab204b9f0e72c85249c5959?
```
zhouwfang, line 86: @benr-ml You wrote it in #62.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Done in #979

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -4785,6 +4757,7 @@ impl MpcManager {
                     outputs.insert(dealer_party_id, output);
                 }
                 avss::ProcessedMessage::Complaint(complaint) => {
+                    // Ben: who will trigger recovery afterwards? and why won't it fail again then (we don't check )
```
zhouwfang, line 4760: @benr-ml Rotation path recovers right here and loops, bounded by the cache short-circuit. Restart path falls back to the full run_dkg instead, deliberately, since we don't put key shares on disk.

Will add a comment.

crates/hashi/src/mpc/types.rs
```diff
@@ -129,10 +130,10 @@ impl PresignatureDerivationVersion {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct MpcConfig {
     pub epoch: u64,
-    pub nodes: Nodes<EncryptionGroupElement>,
+    pub nodes: Nodes<EncryptionGroupElement>, // TODO: reduced weights? comment that
```
zhouwfang, line 133: Yes, added in #979.

crates/hashi/src/mpc/types.rs
```diff
@@ -241,7 +244,7 @@ pub struct GetPublicMpcOutputResponse {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[allow(clippy::large_enum_variant)]
-pub enum Messages {
+pub enum Messages { // nit - this is actually one message, rename?
```
zhouwfang, line 247: Rotation is the exception: RotationMessages is a map of one message per share index, so the plural does hold for that variant. Keeping the name.

crates/hashi/src/mpc/types.rs
```diff
@@ -263,7 +266,7 @@ impl Messages {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct SendMessagesRequest {
-    pub messages: Messages,
+    pub messages: Messages, // Ben: rename to message?
```
zhouwfang, line 269: Same as the enum: mirrors the oneof messages in the proto, and rotation carries one per share index. Singular would drift from the proto name.

crates/hashi/src/mpc/types.rs
```diff
@@ -288,7 +291,7 @@ pub struct RetrieveMessagesRequest {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct RetrieveMessagesResponse {
-    pub messages: Messages,
+    pub messages: Messages, // Ben: rename?
```
zhouwfang, line 294: Same as the enum: mirrors the oneof messages in the proto, and rotation carries one per share index. Singular would drift from the proto name.

crates/hashi/src/mpc/types.rs
```diff
@@ -377,7 +381,7 @@ pub enum ComplaintResponse {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct DealerMessagesHash {
     pub dealer_address: Address,
-    pub messages_hash: MessageHash,
+    pub messages_hash: MessageHash, // here it is MessageHash :)
```
zhouwfang, line 384: That one's the odd one out. Everything around it is plural, so I'd rather make the alias MessagesHash than pull the rest singular.

crates/hashi/src/mpc/types.rs
```diff
@@ -171,6 +172,8 @@ pub enum ProtocolType {
     Signing { message_hash: MessageHash },
 }
 
+// Ben: strange that one can call new with any protocol but we only use it for DKG or key rotation, and use nonce_dealer_session_id otherwise.
+// Would be safer to have instead new_for_dkg and new_for_key_rotation, instead of new.
```
zhouwfang, line 176: @benr-ml Yeah, new is really DKG-or-rotation only. Signing is never constructed at all so I'll drop it.
  
Two constructors wouldn't fully close it though, current_session_id passes the type dynamically. I tried a couple of narrowings and neither held up once I wrote the misuse out.

Root of it is that MpcManager holds one protocol_type but actually serves three protocols, nonce derivation bypasses the field entirely. So the protocol isn't static where it needs to be, and typing further downstream can't recover it. Real fix is protocol-typed handles from the caller, or splitting the manager, both much bigger than this. Leaving the shape alone for now.

zhouwfang:
Clear the remaining mechanical nits from the DKG audit: #980

crates/hashi/src/mpc/types.rs
zhouwfang: "Signing is never constructed at all so I'll drop it." -- done in #980.

crates/hashi/src/mpc/types.rs
```diff
@@ -356,12 +359,13 @@ pub enum ProtocolComplaint {
...
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct ComplainRequest {
+pub struct ComplainRequest { 
+    // nit - order of fields better be more consistent, and be from the most general to the most specific (epoch, dealer, batch_index, etc)
     pub dealer: Address,
     pub share_index: Option<ShareIndex>, // Only for key rotation
     pub batch_index: Option<u32>,        // Only for nonce generation
     pub complaint: ProtocolComplaint,
-    pub protocol_type: ProtocolTypeIndicator,
+    pub protocol_type: ProtocolTypeIndicator, // Ben: is this needed? it is implied from complaint
```
zhouwfang, line 368: @benr-ml Right. Not from the complaint variant alone, since Avss covers both DKG and rotation, but the request pins it: share_index only for rotation, batch_index only for nonce generation.

Plan is to derive it now and reject mismatches, with the complaint variant cross-checked too so an inconsistent pair gets rejected rather than silently resolved. No wire change. Dropping the field from the proto is a separate staged rollout, since it's required today and there's no version negotiation on this RPC.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -5617,6 +5517,7 @@ impl MpcManager {
         if epoch == self.mpc_config.epoch {
             Ok(&self.committee)
         } else {
+            // Ben: assert epoch == mpc_config.epoch-1
```
zhouwfang, line 5520: @benr-ml Agree on asserting, but epoch-1 isn't right. Previous committee is the nearest one below target, not current minus 1, and committee epochs skip when a Sui epoch passes without a reconfig. That would fire on a legit complaint. Should be `epoch == self.previous_epoch`. Will work on it tomorrow.

crates/hashi/src/mpc/types.rs
benr-ml: oh :( i should have documented it, but i usually document the preimage if it's the output of a hash funciton, so here it's likely a random value i picked

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Done in #984

crates/hashi/src/mpc/types.rs
zhouwfang: Added todo in #984

crates/hashi/src/mpc/types.rs
zhouwfang: Added todo in #984

PR #863, opened 2026-07-22
[Guardian] Isolate and revoke temporary initialization state

mskd12:
Groups ceremony state, optional genesis state, and the initialization config hash into one temporary initialization capability. Operator initialization installs the bundle, provisioner initialization and operator activation consume an owned snapshot, and activation removes the entire capability after its durable log commits.

Durable values needed by an active enclave remain directly in EnclaveConfig. This removes broad initialization accessors (e.g., `get_secret_sharing_instance()`) while preserving the existing S3 log schema and lifecycle ordering.

### DIFF
--- a/crates/hashi-guardian/src/enclave.rs
+++ b/crates/hashi-guardian/src/enclave.rs
@@ -1,9 +1,9 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Core enclave types: `Enclave` holds all guardian state (immutable config
-//! set during operator/provisioner-init, mutable runtime state, and
-//! initialization state). Lives in the library so external crates
+//! Core enclave types: `Enclave` holds all guardian state (durable config,
+//! mutable runtime state, and temporary initialization state). Lives in the
+//! library so external crates
 //! (integration test harnesses, ops tooling) can construct and drive an
 //! enclave without going through `main`.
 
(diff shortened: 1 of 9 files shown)


crates/hashi-guardian/src/enclave.rs
```diff
@@ -502,26 +473,28 @@ impl Enclave {
...
+                .map(|state| state.ceremony_state.secret_sharing_instance.clone()),
             bucket_info: self
                 .config
                 .s3_logger()
                 .ok()
                 .map(|l| l.bucket_info().clone()),
             encryption_pubkey: self.encryption_public_key().to_bytes().to_vec(),
-            config_hash: self.config_hash(),
-            genesis_state_hash: self.genesis_state_hash(),
+            config_hash: temporary_init_state.as_ref().map(|state| state.config_hash),
```
dlukegordon, line 488: Clearing temporary_init_state makes config_hash None after activation, but the operator post-check still requires Some(expected_config_hash) at hashi-guardian-init/src/operator_activate.rs:395. This makes activation report failure after it has successfully committed. Please either retain config_hash in durable config or update the post-activation check to expect its removal.

mskd12: i got rid of it as the operator already ensures that an enclave's lifecycle stage is as expected..

dlukegordon (approved):

PR #864, opened 2026-07-22
docs: Add a systemd unit for running the standalone binary

Bridgerz:
Bare-metal operators (running the binary directly, not the container) asked for a setup guide. The runbook is already binary-first — the run command is `hashi server --config validator.toml` and all the config/keys/ports guidance is container-agnostic — but it had **no service-management example**, which is the actual gap they hit.

Adds a **Running as a standalone service (systemd)** block under §6.2 Server with a ready-to-use unit file:
- dedicated unprivileged `hashi` user
- `Restart=on-failure` + `RestartSec`
- `AmbientCapabilities=CAP_NET_BIND_SERVICE` so a non-root process can bind the default `:443` `listen-address` (with a note to drop it if fronting a proxy on a high port)
- `enable --now` + `journalctl -u hashi -f`
- notes on key-file permissions, the persistent `db` path, journald JSON logging, and that this is an alternative to containerizing (§1.1), not an addition

Placed inside §6.2 to avoid renumbering §6.3–6.9. Verified: no style-guide violations (no em dashes / `e.g.` / `on-chain` / italics), anchors resolve, code fences balanced.

### DIFF
--- a/design/docs/node-operator-runbook.mdx
+++ b/design/docs/node-operator-runbook.mdx
@@ -662,6 +662,45 @@ hashi server [--config <path>]
 
 Run the Hashi validator daemon. Uses the validator config.
 
+**Running as a standalone service (systemd).** Operators who run the binary directly rather than a container can manage it with systemd. Install the verified binary at `/usr/local/bin/hashi` (confirm its sha256 first, see [1.3](#13-software-dependencies)), place your config at `/etc/hashi/validator.toml`, and run the daemon under a dedicated unprivileged user:
+
+```ini
+# /etc/systemd/system/hashi.service
+[Unit]
+Description=Hashi validator node
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Type=simple
+User=hashi
+Group=hashi
+ExecStart=/usr/local/bin/hashi server --config /etc/hashi/validator.toml
+Restart=on-failure
+RestartSec=5
+LimitNOFILE=65536
+# The node binds to :443 by default (the listen-address field). This grants the
+# non-root process the one privileged-port capability it needs. Drop this line if
+# you set listen-address to a port at or above 1024 and front it with a proxy.
+AmbientCapabilities=CAP_NET_BIND_SERVICE
+
+[Install]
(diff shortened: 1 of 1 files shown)


design/docs/node-operator-runbook.mdx
```diff
@@ -662,6 +662,45 @@ hashi server [--config <path>]
 
 Run the Hashi validator daemon. Uses the validator config.
 
+**Running as a standalone service (systemd).** Operators who run the binary directly rather than a container can manage it with systemd. Install the verified binary at `/usr/local/bin/hashi` (confirm its sha256 first, see [1.3](#13-software-dependencies)), place your config at `/etc/hashi/validator.toml`, and run the daemon under a dedicated unprivileged user:
```
dlukegordon, line 665: The linked verification procedure does not work exactly as expected, build.sh prints the checksum but does not create out/hashi.sha256. Please update instructions.

design/docs/node-operator-runbook.mdx
```diff
@@ -662,6 +662,45 @@ hashi server [--config <path>]
...
+AmbientCapabilities=CAP_NET_BIND_SERVICE
+
+[Install]
+WantedBy=multi-user.target
+```
+
+Enable it, start it, and follow the logs:
+
+```bash
+sudo systemctl enable --now hashi
```
dlukegordon, line 695: We need to add `sudo systemctl daemon-reload` before enable --now, otherwise systemd may not load the newly installed or updated unit.

design/docs/node-operator-runbook.mdx
```diff
@@ -662,6 +662,45 @@ hashi server [--config <path>]
...
+journalctl -u hashi -f
+```
+
+Notes:
+
+- The `hashi` user needs read access to the config and key files (keep key files `chmod 600`) and write access to the `db` path, which must be on persistent storage.
+- The node writes structured JSON logs to stdout, so journald captures everything and there is no separate log file to rotate.
```
dlukegordon: Actually stderr

dlukegordon (approved):

PR #867, opened 2026-07-22
[Guardian] Cleanup and test hardening

mskd12:
## Summary

- verify the exact `kp-shares` S3 snapshot written by a KP certificate rotation instead of reading globally latest state
- share the S3 fetch, version-history, signature, attestation, PCR, decoding, and read-logging path between exact and latest KP-share reads
- replace natural protobuf-to-domain conversion helpers with `TryFrom` implementations, retaining helpers only where Rust's orphan rules prevent the trait implementation
- remove stale guardian TODOs and document why limiter recovery reaches genesis only when no withdrawal has ever succeeded
- strengthen ceremony tests by decrypting returned KP shares, verifying their commitments, and reconstructing the advertised Bitcoin public key
- assert the exact signed S3 records emitted by operator/provisioner initialization, withdrawals, rate-limit failures, and heartbeats

The production behavior change is limited to binding KP certificate rotation to the exact S3 snapshot it just wrote; the remaining changes are cleanup and test hardening.

## Verification

- `make fmt`
- CI

### DIFF
--- a/crates/hashi-guardian/src/ceremony_mode/setup.rs
+++ b/crates/hashi-guardian/src/ceremony_mode/setup.rs
@@ -97,7 +97,9 @@ pub async fn setup_new_key(
 mod tests {
     use super::*;
     use crate::mock_logger_capturing;
-    use hashi_types::guardian::test_utils::mock_kp_certs_roster;
+    use crate::test_utils::decrypt_kp_shares;
+    use crate::test_utils::mock_kp_certs_roster_with_secrets;
+    use hashi_types::guardian::crypto::combine_shares;
     use hashi_types::guardian::LogMessage;
     use hashi_types::guardian::LogRecord;
     use hashi_types::guardian::VersionedLogMessage;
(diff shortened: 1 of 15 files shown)


crates/hashi-guardian/src/ceremony_mode/setup.rs
dlukegordon, line 210: This only checks the persisted roster’s share count, so persisting different ciphertexts from those returned would pass. Can we also assert `shares.encrypted_shares == validated_resp.encrypted_shares` and that the logged sharing instance equals the response instance. Also we need the same thing at rotate.rs:310.

crates/hashi-guardian/src/s3_reader.rs
```diff
@@ -216,28 +216,61 @@ impl GuardianReader {
...
-        Ok(Some(*msg))
+        Ok(Some(msg))
+    }
+
+    /// Read and verify one exact encrypted KP-share snapshot.
+    ///
+    /// The object key binds the writing guardian session and the two sequence
+    /// numbers. This lets callers verify the snapshot produced by one request
+    /// even if a later request has already advanced the latest state.
+    pub async fn read_kp_share_state_log(
```
dlukegordon, line 239: This fixes the concurrent-rotation race but has no direct test coverage. Please add a test with snapshots N and N+1 that requests N and verifies the exact key, history, signature, and build-policy checks.

crates/hashi-guardian/src/ceremony_mode/setup.rs
mskd12: Addressed in be265980. Setup now asserts that the complete logged sharing instance equals the response instance and that the persisted ciphertext roster equals the returned roster. Rotation now asserts that its persisted ciphertext roster equals the returned roster as well. (`RotateKpsResponse` does not expose the new sharing instance, so there is no response instance to compare on that path.)

crates/hashi-guardian/src/s3_reader.rs
mskd12: i think testing this is a little messy.. so i am leaving it as is..

dlukegordon (approved):

PR #868, opened 2026-07-22
[Guardian] Make mutating RPCs cancellation-safe

mskd12:
## Summary

- add a `task_spawner` boundary for every state-changing Guardian RPC
- centralize Tokio task ownership in `Enclave::spawn_task`, so accepted work survives caller cancellation
- compose `spawn_control_task` with a private control lock for lifecycle and control-plane transitions
- keep standard withdrawal on `spawn_task`, using its limiter guard as the narrower serialization boundary
- remove nested task spawning from S3 retry and withdrawal logging helpers
- preserve standard-withdrawal validation, limiter, signing, logging, and response ordering
- test caller cancellation survival and control-task serialization, with comments explaining the simulated RPC behavior

## Verification

- `make fmt`
- CI

### DIFF
--- a/crates/hashi-guardian/src/s3_client.rs
+++ b/crates/hashi-guardian/src/s3_client.rs
@@ -113,21 +113,15 @@ impl GuardianS3Client {
         self.write_at_key(&key, &log, object_lock_duration).await
     }
 
-    /// Retry an immutable log write through the grace period, then abort. The
-    /// worker is detached so caller cancellation cannot abandon an in-flight PUT.
+    /// Retry an immutable log write through the grace period, then abort.
+    /// State-changing RPCs call this from their root-owned task.
     pub async fn write_log_record_or_abort(&self, log: LogRecord) -> GuardianResult<()> {
-        let writer = self.clone();
-        tokio::spawn(async move {
-            writer
-                .write_log_record_or_abort_inner(
-                    log,
-                    MAX_S3_WRITE_FAILURE_INTERVAL,
-                    S3_WRITE_RETRY_INTERVAL,
-                )
-                .await
-        })
+        self.write_log_record_or_abort_inner(
+            log,
+            MAX_S3_WRITE_FAILURE_INTERVAL,
+            S3_WRITE_RETRY_INTERVAL,
+        )
         .await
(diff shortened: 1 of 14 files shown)


crates/hashi-guardian/src/withdraw_mode/committee_update.rs
```diff
@@ -25,9 +25,6 @@ pub async fn update_committee(
     enclave: Arc<Enclave>,
     signed: HashiSigned<CommitteeTransitionRequest>,
 ) -> GuardianResult<u64> {
```
dlukegordon, line 27: This function is public but no longer acquires the control lock. Concurrent direct calls can race and abort the enclave. Please make unlocked handlers private, exposing only the locked task_spawner wrappers.

crates/hashi-guardian/src/s3_client.rs
```diff
@@ -113,21 +113,15 @@ impl GuardianS3Client {
         self.write_at_key(&key, &log, object_lock_duration).await
     }
 
-    /// Retry an immutable log write through the grace period, then abort. The
-    /// worker is detached so caller cancellation cannot abandon an in-flight PUT.
+    /// Retry an immutable log write through the grace period, then abort.
+    /// State-changing RPCs call this from their root-owned task.
     pub async fn write_log_record_or_abort(&self, log: LogRecord) -> GuardianResult<()> {
```
dlukegordon, line 118: Similarly, can we somehow restrict this to only be callable by cancelation safe callers, such as by making it private.

crates/hashi-guardian/src/withdraw_mode/committee_update.rs
mskd12: Answering both the questions here itself. What are you concerned about? External crates only use the S3 readers exported from the `hashi-guardian` crate. So my framework is to be careful about pub/private in those code, whereas for the rest (RPCs, S3 writers), I dont think they matter.

dlukegordon: Good point, I forgot that this is all only accessible from what is is exported from the crate. So disregard please.

dlukegordon (approved):

PR #869, opened 2026-07-22
Auto-heal encryption-key loss via re-registration and fresh-member rejoin

zhouwfang:
MystenLabs/hashi#163

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -5020,21 +5020,28 @@ impl MpcManager {
         p2p_channel: &impl P2PChannel,
         previous_committee_threshold: u64,
     ) -> MpcResult<PublicMpcOutput> {
-        let (previous_committee, epoch) = {
+        let (previous_committee, previous_nodes, epoch) = {
             let mgr = mpc_manager.read().unwrap();
             let previous_committee = mgr
                 .previous_committee
                 .clone()
                 .expect("key rotation requires previous committee");
-            (previous_committee, mgr.previous_epoch)
+            let previous_nodes = mgr.previous_nodes.clone().ok_or_else(|| {
+                MpcError::InvalidConfig("previous_nodes required for public-output quorum".into())
+            })?;
+            (previous_committee, previous_nodes, mgr.previous_epoch)
         };
         let request = GetPublicMpcOutputRequest { epoch };
         let mut futures: FuturesUnordered<_> = previous_committee
             .members()
             .iter()
-            .map(|member| {
+            .enumerate()
+            .map(|(party_id, member)| {
                 let addr = member.validator_address();
-                let weight = member.weight();
+                let weight = previous_nodes
+                    .share_ids_of(party_id as u16)
+                    .map(|ids| ids.len() as u64)
+                    .unwrap_or(0);
                 let req = request.clone();
                 async move {
                     let result = p2p_channel.get_public_mpc_output(&addr, &req).await;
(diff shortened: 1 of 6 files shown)


Bridgerz (approved):

crates/hashi/src/mpc/service.rs
```diff
@@ -746,6 +748,7 @@ impl MpcService {
             Ok(output) => output,
             Err(e) => {
                 error!("sync_if_stale: recover_mpc_state failed for epoch {epoch}: {e}");
+                self.re_register_encryption_key_if_lost().await;
```
dlukegordon: After re-registering, the service remains in the startup loop waiting for a current-epoch SigningManager, so it never handles a pending reconfiguration. This can stall healing when the validator’s weight is required. Check pending_epoch_change in the startup loop and call handle_reconfig when replacement keys are ready.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -5073,16 +5073,20 @@ impl MpcManager {
...
                 .is_some();
-            (is_member, mgr.previous_reconfig_output_threshold)
+            (
+                is_member,
+                mgr.previous_encryption_key.is_some(),
+                mgr.previous_reconfig_output_threshold,
+            )
         };
-        let previous = if is_member_of_previous_committee {
+        let previous = if is_member_of_previous_committee && has_previous_key {
```
dlukegordon, line 5096: This new fallback relies on fetch_public_mpc_output_from_quorum, which compares a reduced threshold against original committee weights. With weight reduction enabled, one peer may incorrectly satisfy quorum. Sum weights from previous_nodes rather than CommitteeMember::weight().

crates/hashi/src/mpc/service.rs
```diff
@@ -781,6 +784,48 @@ impl MpcService {
...
+            .state()
+            .hashi()
+            .committees
+            .current_committee()
+            .cloned();
+        let Some(committee) = committee else { return };
+        let Ok(me) = self.inner.config.validator_address() else {
+            return;
+        };
+        if !self.inner.committee_encryption_key_lost(&committee, me) {
```
dlukegordon: This only detects encryption-key loss, but manager recovery also fails when only the signing key is missing. In that case no replacement is registered despite the updated error promising automatic recovery. Detect definitive loss of either key and register a replacement pair.

crates/hashi/src/mpc/service.rs
```diff
@@ -781,6 +784,48 @@ impl MpcService {
...
+            return;
+        }
+        self.inner.metrics.mpc_encryption_key_lost_total.inc();
+        let target = match self.inner.next_reconfig_epoch().await {
```
dlukegordon: next_reconfig_epoch() may return an epoch whose pending committee has already been snapshotted, so this registration cannot affect that reconfiguration. Check whether the pending committee contains the replacement key; if it does not, target the following epoch.

dlukegordon: This still races with start_reconfig: the pending committee can be snapshotted after the check above but before this registration completes, leaving the replacement keys excluded from target. After registration, refresh the authoritative committee state and verify that target contains the newly registered keys. If it does not, prepare and register keys for target + 1.

zhouwfang: Fixed with verify-after-write. The race itself can't be eliminated, but losing it now costs exactly one detected epoch (the cryptographic floor anyway).

dlukegordon (approved):

PR #870, opened 2026-07-22
Stop a stalled cert wait from freezing the MPC service loop

zhouwfang:
MystenLabs/hashi#163

### DIFF
--- a/crates/hashi/src/communication/sui_tob.rs
+++ b/crates/hashi/src/communication/sui_tob.rs
@@ -203,6 +203,18 @@ impl OrderedBroadcastChannel<CertificateV1> for SuiTobChannel {
                 }
             }
             if self.pending_certs.is_empty() {
+                let (onchain_epoch, pending) = {
+                    let state = self.onchain_state.state();
+                    let committees = &state.hashi().committees;
+                    (committees.epoch(), committees.pending_epoch_change())
+                };
+                if tob_wait_superseded(self.protocol_type, self.epoch, onchain_epoch, pending) {
+                    return Err(ChannelError::Superseded(format!(
+                        "{:?} TOB wait for epoch {} (onchain epoch {onchain_epoch}, \
+                         pending epoch change {pending:?})",
+                        self.protocol_type, self.epoch,
+                    )));
+                }
                 tokio::time::sleep(POLL_INTERVAL).await;
             }
         }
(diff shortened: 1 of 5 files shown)


crates/hashi/src/communication/sui_tob.rs
```diff
@@ -203,6 +203,18 @@ impl OrderedBroadcastChannel<CertificateV1> for SuiTobChannel {
                 }
             }
             if self.pending_certs.is_empty() {
+                let (onchain_epoch, pending) = {
+                    let state = self.onchain_state.state();
+                    let committees = &state.hashi().committees;
+                    (committees.epoch(), committees.pending_epoch_change())
+                };
+                if tob_wait_superseded(self.protocol_type, self.epoch, onchain_epoch, pending) {
+                    return Err(ChannelError::Superseded(format!(
```
Bridgerz, line 241: I think the consumer converts this to a generic BroadcastError, so in the logs it's indistinguishable from a real broadcast failure. Maybe we add a tracing::info! or debug to add context?

Bridgerz (approved):

crates/hashi/src/communication/sui_tob.rs
```diff
@@ -203,6 +203,18 @@ impl OrderedBroadcastChannel<CertificateV1> for SuiTobChannel {
                 }
             }
             if self.pending_certs.is_empty() {
+                let (onchain_epoch, pending) = {
+                    let state = self.onchain_state.state();
+                    let committees = &state.hashi().committees;
+                    (committees.epoch(), committees.pending_epoch_change())
+                };
+                if tob_wait_superseded(self.protocol_type, self.epoch, onchain_epoch, pending) {
```
dlukegordon, line 234: Supersession is checked only after pending_certs is drained. A quorum of buffered certificates can let an aborted DKG/rotation complete and continue post-reconfig setup for a removed committee. Check supersession before returning buffered certificates, or revalidate the pending epoch after the protocol succeeds.

crates/hashi/src/mpc/service.rs
```diff
@@ -141,7 +148,7 @@ impl MpcService {
             self.try_submit_genesis_reconfig().await;
         } else if self.inner.is_in_current_committee() {
             loop {
-                self.sync_if_stale().await;
+                timeboxed("sync_if_stale", self.sync_if_stale()).await;
```
dlukegordon: This times out the entire recovery scan, which always restarts from batch 0 (recover_presigning_state, lines 586-649). A valid scan taking over 900 seconds will restart forever. Could we time out individual fetches or persist/resume recovery progress instead?

crates/hashi/src/mpc/service.rs
```diff
@@ -99,6 +100,12 @@ pub struct MpcService {
     backup_handle: crate::backup::BackupHandle,
 }
 
+async fn timeboxed(label: &str, fut: impl Future<Output = ()>) {
+    if tokio::time::timeout(ARM_TIMEOUT, fut).await.is_err() {
```
dlukegordon: Dropping this future does not cancel any active spawn_blocking jobs inside it. Those jobs can continue mutating or locking the MPC manager after the next attempt begins. Limit this timeout to cancellation-safe operations or add cooperative cancellation/generation checks before committing blocking results.

crates/hashi/src/mpc/service.rs
zhouwfang: I think neither of these two options is robust. I think the robust fix is cooperative pending-abort inside the scan, and removing the external timeout from the sync arms entirely.

bmwill:
@zhouwfang looks like this also needs to be rebased to fix up some merge conflicts.

zhouwfang:
> @zhouwfang looks like this also needs to be rebased to fix up some merge conflicts.

Yes, I have been working on it.

bmwill (approved):

dlukegordon (approved):

PR #871, opened 2026-07-23
Lossless watcher

bmwill:


crates/hashi/src/onchain/apply.rs
```diff
@@ -0,0 +1,1543 @@
...
+                OutputObjectState::DoesNotExist => changes.push(TxChange::Deleted { id }),
+                // Accumulator writes and unknown future states don't
+                // fit the object-contents shape; ignore them.
+                _ => {}
+            }
+        }
+
+        Ok(Some(Self {
+            checkpoint: tx.checkpoint(),
+            transaction_index: tx.transaction_index(),
```
0xsiddharthks: afaik both checkpoint and transaction_index default to 0 instead of erroring out.

Both are `optional uint64`, and prost's bare accessor returns 0 when the field is
absent.

I think the zeros in here will mess the dedup rachet at `mirror.rs:393` (we use `if view.position() <= mirror.applied_position { return Ok(()); }` in here)

We already have `checkpoint_opt()` and `transaction_index_opt()` generated. So I think we should use something like: 

```rust
checkpoint: tx.checkpoint_opt().context("ExecutedTransaction is missing checkpoint")?,
transaction_index: tx.transaction_index_opt()
    .context("ExecutedTransaction is missing transaction_index")?,
```

Also for timestamp_ms , we should not have 0 cause i think the `limiter.apply_consume(seq, timestamp_ms / 1000, …)` at `mirror.rs:493` will be impacted.

so we could either `?` or maybe fallback to `state.latest_checkpoint_timestamp_ms()` ?

crates/hashi/src/onchain/mirror.rs
```diff
@@ -0,0 +1,550 @@
...
+        }
+        tokio::time::sleep(RECONNECT_DELAY).await;
+    }
+}
+
+fn build_client(sui_rpc_url: &str, state: &OnchainState) -> Result<Client> {
+    // A fresh client per attempt: re-subscribing on the same channel can
+    // reuse a wedged h2 connection — the one whose stream just stalled —
+    // and silently hang again.
+    let mut client = Client::new(sui_rpc_url)?;
```
0xsiddharthks: We should have a request timeout here.

All the other sui clients use `crate::sui_rpc_client::new_sui_rpc_client` which adds a 60s `TimeoutLayer`.

crates/hashi/src/onchain/mirror.rs
```diff
@@ -0,0 +1,550 @@
...
+        if let Some(metrics) = metrics {
+            metrics.watcher_unrouted_objects_total.inc();
+        }
+    }
+    mirror.applied_position = view.position();
+    // In-order delivery means everything before this transaction's
+    // checkpoint has been delivered; its own checkpoint counts as
+    // covered too, up to the same-burst stragglers documented on
+    // `wait_until_checkpoint`.
+    advance_watermark(state, mirror, view.checkpoint);
```
0xsiddharthks: Applying one transaction claims its whole checkpoint, which breaks the GC gate this PR adds.

I don't exactly understand the issue here. But I'm doing a back and forth with claude and this is what it's telling me. 🫣:

I'm working on trying to understand this a bit more.
```
The softness is documented on `wait_until_checkpoint` (`onchain/mod.rs:252-259`),
but the justification there — "in the common single-transaction case is exactly
their own" — is precisely the case the new GC gate is *not* in.

`garbage_collection.rs:222-236` waits for `scan_target` (the checkpoint the
confirm landed in), then reads the mirror. If the confirm is at index 5 of
checkpoint C and *any other* Hashi transaction is at index 2 of C, applying index
2 pushes the watermark to C, the wait returns, and
`find_spent_utxos_pending_cleanup` runs over records whose `spent_by` hasn't been
applied yet. It comes back empty, `cleanup_spent_utxos` returns `Ok(0)`, and the
caller disarms the scan — "stranding the spent records until the next confirm",
which is the exact sentence at `withdrawal_transactions.rs:1063` explaining why
`scan_target` was added. The gate narrows the race instead of closing it.

`mpc/service.rs:908` has the same shape — wait for the registration's checkpoint,
then read `committees().get(&target)` — and the same failure, ending the bump loop
an epoch short.

The server's own watermark is already exact: `query_options.proto:97-103` defines
`Watermark.checkpoint` as the checkpoint the scan has *fully covered*, so a frame
carrying tx #2 of C reports at most C−1. Deleting line 418 and leaning on the
frame watermark makes `wait_until_checkpoint` mean what its name says. The cost is
waiting for the next progress frame, which your own comment at `mirror.rs:262-265`
puts at ~25 checkpoints — call it 6s, well inside the 30s `VISIBILITY_TIMEOUT` on
both callers. If you'd rather keep the fast path,
`view.checkpoint.saturating_sub(1)` is correct and costs nothing.
```

0xsiddharthks:
Deployed on devnet: https://github.com/MystenLabs/sui-operations/actions/runs/30391601118

0xsiddharthks (approved):

PR #872, opened 2026-07-23
[Guardian] Add environment-aware S3 retention

mskd12:
## Summary

- add a required retention environment to shared Guardian S3 configuration for both readers and writers
- select two retention classes: Mainnet/Testnet use 10 years long-lived and 30 days short-lived; Devnet uses 90 days long-lived and 7 days short-lived
- map heartbeats and encrypted KP shares to short-lived retention, and all other Guardian logs to long-lived retention
- require unexpired Compliance locks on records that depend on S3 immutability, while allowing signature-verified KP shares to remain readable after their lock expires
- keep the local Guardian Docker flow current by selecting the Devnet retention policy in its generated init config
- correct the Hashi monitor YAML examples to match the shared S3 configuration shape
- add focused expired-lock tests

## Open question

Should retention_environment remain an explicit Guardian S3 setting, or should it eventually be replaced by or derived from a repository-wide Hashi network identifier? A TODO remains on the configuration field so this PR presents the question without deciding the broader network model.

Object Lock expiry permits deletion but does not trigger it. Heartbeats may be removed by a separate lifecycle policy. Stale encrypted KP shares can be explicitly deleted after their recovery window, for example when a KP encryption key is retired or lost.

The signed log payload shapes are unchanged. The future purpose-built operator-init log should record the retention environment solely for historical and audit context—not for reader policy discovery or policy versioning. Readers carry the configured fixed policy; a TODO(testnet-wipe) notes that, once legacy seven-day logs are gone, required reads should also verify that each retain-until deadline covered the configured duration for its log type.

## Verification

- make fmt


crates/hashi-monitor/audit.sample.yaml
```diff
@@ -10,6 +10,7 @@ next_event_delays:
 
 guardian:
```
dlukegordon, line 11: This block does not match S3Config at hashi-types/src/guardian/mod.rs:383-390, so the documented commands fail to parse. Please update both examples to include access_key, secret_key, session_token, and nested bucket_info: { bucket, region }, alongside retention_environment. Also the same goes for README.md:55-57**.**

mskd12: done!

dlukegordon (approved):

PR #873, opened 2026-07-23
[Guardian] Improve error classification

mskd12:
## Summary

- represent lifecycle mismatches with expected/actual lifecycle values and use lifecycle requirements consistently for activated-only operations
- distinguish S3 API/history failures, unrecoverable invalid S3 logs, and potentially recoverable guardian build/PCR policy failures
- keep reusable signature and attestation verification source-neutral, then classify failures at the request or persisted-log boundary
- preserve typed `GuardianReader` errors instead of flattening them into generic input or internal errors
- report invalid Hashi committee certificates as unauthenticated and rejected guardian builds as failed preconditions
- classify limiter sequence mismatches as gRPC `Aborted` and rate limiting as `ResourceExhausted`
- update the Hashi leader to consume those status codes instead of parsing Guardian error messages

## Verification

- `make fmt`
- `git diff --check`

### DIFF
--- a/crates/hashi-guardian/src/withdraw_mode/operator_activate.rs
+++ b/crates/hashi-guardian/src/withdraw_mode/operator_activate.rs
@@ -101,7 +101,10 @@ pub async fn operator_activate(
 ) -> GuardianResult<()> {
     info!("/operator_activate - Received request.");
 
-    enclave.require_lifecycle(WithdrawStage::ProvisionerInitialized.into())?;
+    enclave.require_lifecycle(
+        "operator_activate",
+        WithdrawStage::ProvisionerInitialized.into(),
+    )?;
     info!("Lifecycle stage validated.");
 
     // ---- Validate & build: Nothing in this phase mutates enclave state, so any
--- a/crates/hashi-guardian/src/withdraw_mode/operator_activate.rs
+++ b/crates/hashi-guardian/src/withdraw_mode/operator_activate.rs
@@ -17,7 +17,6 @@ use hashi_types::guardian::RateLimiter;
 use hashi_types::guardian::WithdrawStage;
 use std::sync::Arc;
 use tracing::info;
-use GuardianError::InternalError;
 use GuardianError::InvalidInputs;
 
 /// S3-derived activation state ready for its fail-stop commit.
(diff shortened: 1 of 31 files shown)


crates/hashi-types/src/guardian/errors.rs
```diff
@@ -4,31 +4,92 @@
 use serde::Deserialize;
 use serde::Serialize;
 
+use super::lifecycle::EnclaveLifecycle;
+
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
```
dlukegordon, line 9: Warning: GuardianError is embedded in signed S3 failure logs, so changing variant ordering changes BCS discriminants and may invalidate existing signatures. Removing EnclaveUninitialized may also prevent old logs from deserializing. If existing logs must remain readable, preserve the old variants/order and append new variants, or introduce a versioned persisted error type. A fixture covering logs signed by the previous version would verify compatibility.

If we plan to manually deal with the old s3 logs this is fine.

crates/hashi-guardian/src/withdraw_mode/operator_activate.rs
```diff
@@ -52,16 +51,12 @@ impl OAInstall {
```
dlukegordon, line 51: This conversion flattens typed S3Error and InvalidS3Log failures into InvalidInputs, which then becomes gRPC InvalidArgument. Could we propagate S3/data-integrity errors unchanged and classify only heartbeat-state failures with an appropriate precondition or retryable variant?

mskd12: I think the next PR (#877) addresses this. Let me know if it misses what you had in mind.

dlukegordon: That LGTM

crates/hashi-types/src/guardian/errors.rs
mskd12: Yeah, I feel like manually dealing with old S3 failure logs when we want to read seems appropriate. For mainnet, I wonder if we should just write a string in the S3 logs instead of serializing the full error type (to avoid this issue).

dlukegordon: I agree that would be a good idea

dlukegordon (approved):

PR #874, opened 2026-07-23
hashi: treat a lost committee handoff race as reconfig completion

bmwill:
When several nodes race to finish a reconfig, the first end_reconfig to land closes the reconfig window, and every other node's queued submit_committee_handoff then aborts with ENotReconfiguring. Previously the abort classifier only mapped ENotReconfiguring to the benign "already completed" case when it came from `end_reconfig`, so the same abort from `submit_committee_handoff` classified as a non-retryable Move abort and panicked every losing node (seen as an intermittent failure of test_new_member_joins_key_rotation_after_dkg: eight nodes panicked and the new member's MPC key wait timed out).

With this commit, the classifier matches ENotReconfiguring from any `reconfig` entry point, and the handoff submission loop handles EndReconfigAlreadyCompleted the way the end_reconfig loop already does: wait for the object mirror to observe the epoch change, then return success once the epoch is no longer pending. Also add unit tests for the classifier (the losing-side handoff race, classification through anyhow context chains, the cert-already-set abort, other reconfig aborts staying non-retryable, and non-Move-abort errors), backed by a test-only constructor on TransactionExecutionError.


crates/hashi/src/mpc/service.rs
```diff
@@ -1496,8 +1496,25 @@ impl MpcService {
...
-                    | ReconfigSubmissionErrorKind::EndReconfigAlreadyCompleted => {
+                    ReconfigSubmissionErrorKind::EndReconfigAlreadyCompleted => {
+                        // Another node's end_reconfig closed the reconfig
+                        // window before our handoff landed; the handoff cert
+                        // is already recorded on-chain, so there is nothing
+                        // left to submit.
+                        warn!(
+                            "submit_committee_handoff submission for epoch {epoch} found reconfig already completed; waiting for the watcher to observe it: {e}"
+                        );
+                        self.wait_for_epoch_change_visibility(epoch).await;
```
dlukegordon, line 1524: That method doesn't exist

bmwill: thats what i get for cherry-picking it from another PR i'm working on lol

bmwill: reworked to be based on top of my other PR for now

crates/hashi/src/mpc/service.rs
```diff
@@ -1513,8 +1513,25 @@ impl MpcService {
...
                         return Ok(());
                     }
-                    ReconfigSubmissionErrorKind::NonRetryableMoveAbort
-                    | ReconfigSubmissionErrorKind::EndReconfigAlreadyCompleted => {
+                    ReconfigSubmissionErrorKind::EndReconfigAlreadyCompleted => {
+                        // Another node's end_reconfig closed the reconfig
+                        // window before our handoff landed; the handoff cert
+                        // is already recorded on-chain, so there is nothing
+                        // left to submit.
+                        warn!(
```
Bridgerz, line 1521: Do we want to warn here? Race loss is normal operation

crates/hashi/src/mpc/service.rs
```diff
@@ -1513,8 +1513,25 @@ impl MpcService {
...
+                            "submit_committee_handoff submission for epoch {epoch} found reconfig already completed; waiting for the watcher to observe it: {e}"
+                        );
+                        self.wait_for_epoch_change_visibility(epoch).await;
+                        if self.get_pending_epoch_change() != Some(epoch) {
+                            return Ok(());
```
zhouwfang, line 1526: If the window was closed by abort_reconfig, this returns Ok without the epoch flipping and handle_reconfig then prunes current-epoch state. It is pre-existing on the end_reconfig arm too, so follow-up is fine. I created an issue.

zhouwfang: Done in #880.

PR #877, opened 2026-07-23
[Guardian] Move activation heartbeat retries into guardian-init

mskd12:
## Summary

- model activation heartbeat readiness with structured current-session-not-live and prior-session-still-recent Guardian errors
- treat missing and stale current-session heartbeats uniformly, retrying once per heartbeat interval for a fixed ten-minute window
- wait for prior sessions using the ten-minute quiet period plus a five-minute shutdown/startup buffer
- compute both retry deadlines once so repeated checks cannot extend either wait window
- preserve heartbeat error types through the Guardian activation path and map retryable readiness to gRPC `Unavailable`
- fail immediately for S3, integrity, build, authentication, and other non-heartbeat errors
- remove the local provisioning script message-parsing retry loop now that the CLI owns the bounded retry policy

## Tradeoff

The recent heartbeat scan remains bounded. If the presumed current session last heartbeated outside that scan, it is treated like any other current-session-not-live result and may wait up to ten minutes before failing. A historical lookup can be added later if this uncommon delay becomes operationally significant.

## Follow-up

- Replace serialized `GuardianError` values in failure logs with strings before mainnet.
- Extend and test `docker/hashi-guardian-local` with standby-guardian activation; the current harness covers genesis activation only.

## Verification

- `make fmt`
- `git diff --check`

### DIFF
--- a/crates/hashi-guardian/src/s3_reader/heartbeat_checks.rs
+++ b/crates/hashi-guardian/src/s3_reader/heartbeat_checks.rs
@@ -3,13 +3,15 @@
 
 use super::heartbeat_cursor;
 use super::GuardianReader;
+use crate::HEARTBEAT_INTERVAL;
 use crate::LIVE_SESSION_LATEST_HEARTBEAT_MAX_AGE;
 use crate::OTHER_SESSION_QUIET_PERIOD;
 use hashi_types::guardian::time_utils::now_timestamp_secs;
 use hashi_types::guardian::time_utils::unix_millis_to_seconds;
 use hashi_types::guardian::time_utils::UnixSeconds;
-use hashi_types::guardian::GuardianError::InvalidInputs;
+use hashi_types::guardian::GuardianError::CurrentSessionHeartbeatNotLive;
 use hashi_types::guardian::GuardianError::InvalidS3Log;
+use hashi_types::guardian::GuardianError::PriorSessionHeartbeatStillRecent;
 use hashi_types::guardian::GuardianResult;
 use hashi_types::guardian::LogMessage;
 use hashi_types::guardian::SessionID;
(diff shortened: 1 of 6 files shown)


crates/hashi-types/src/guardian/errors.rs
```diff
@@ -35,13 +37,48 @@ pub enum GuardianError {
         expected: u64,
         actual: u64,
     },
+    /// The standby session has not produced its first visible heartbeat.
```
dlukegordon: Just noting we have the same issue as with the last PR, another reminder to use strings.

crates/hashi-guardian/src/s3_reader/heartbeat_checks.rs
```diff
@@ -115,34 +118,32 @@ fn validate_session_live_and_others_quiet(
     let live_session_info = summary
         .iter()
         .find(|s| s.session_id.as_str() == live_session)
-        .ok_or_else(|| {
-            InvalidInputs(format!(
-                "no heartbeat logs found for session {live_session}"
-            ))
+        .ok_or_else(|| CurrentSessionHeartbeatMissing {
```
dlukegordon: An absent entry may represent a heartbeat older than the limited scan window, not a session that never heartbeated. This causes stale sessions to retry for 15 minutes. Could we look up the session’s latest historical heartbeat when absent and return CurrentSessionHeartbeatStale when one exists?

mskd12: That’s true. I wanted to keep the retry policy simple for now; my latest commit intentionally treats missing and stale current-session heartbeats the same and applies a single bounded 10-minute retry window.

I’m happy to add the historical lookup, but the extra complexity may not be warranted yet. We would only encounter this case if the operator attempts to activate a session as current even though it has no heartbeat in the recent three-prefix scan. In that uncommon case, the consequence is an unnecessary 10-minute wait before failing.

crates/hashi-types/src/guardian/errors.rs
mskd12: Sounds good. I am working on it next. Will create a follow up PR.

dlukegordon (approved):

PR #878, opened 2026-07-23
Filter sub-quorum AVID nonce certs from recovery sizing

zhouwfang:
MystenLabs/hashi#531

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -1110,11 +1110,11 @@ impl MpcManager {
             let dealer = &cert.message().dealer_address;
             if dealer != table_dealer {
                 tracing::warn!(
-                    "Nonce cert keyed under {:?} but signed for dealer {:?}; skipping",
+                    "Nonce cert served under table key {:?} but signed for dealer {:?}; \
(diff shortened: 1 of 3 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1098,6 +1100,77 @@ impl MpcManager {
...
+            let dealer = &cert.message().dealer_address;
+            if dealer != table_dealer {
+                tracing::warn!(
+                    "Nonce cert keyed under {:?} but signed for dealer {:?}; skipping",
+                    table_dealer,
+                    dealer
+                );
+                continue;
+            }
+            if certified.contains(dealer) {
```
benr-ml, line 1116: why is this an issue? it's the normal flow, no?

benr-ml (approved):

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: You're right on the honest path: first-write-wins per dealer means one cert per dealer on chain, so this never fires. I kept it for parity with the replay, which does the same check: `len == expected` compares the two and the certs are frozen, so any disagreement retries forever.

PR #879, opened 2026-07-23
Skip duplicate deposit confirmation retries

dlukegordon:


crates/hashi/src/leader/deposits.rs
```diff
@@ -143,9 +144,29 @@ impl LeaderService {
...
+            .with_label_values(&["utxo_already_used"])
+            .set(utxo_already_used_count as i64);
+
+        let approved_deposit_ids: Vec<Address> = deposit_confirmation_candidates
             .iter()
-            .map(|request| request.id)
+            .map(|r| r.id)
             .collect();
         self.approved_deposit_retry_tracker
             .prune(&approved_deposit_ids);
```
Bridgerz, line 172: i think prune needs the pre-filter id list. a duplicate that gets shadowed by the dedup is still in the onchain queue, but pruning it here wipes its retry state every tick, including the permanent max_retries stop. when the shadow lifts (the winner goes never-retry, or its cert goes stale at reconfig) the shadowed request comes back with a fresh retry budget and we re-run full signature rounds for something that already permanently failed. building approved_deposit_ids before filter_deposit_confirmation_candidates runs should fix it.

crates/hashi/src/leader/deposits.rs
```diff
@@ -492,3 +513,71 @@ fn deposit_request_to_proto(req: &DepositRequest) -> SignDepositConfirmationRequ
...
+) -> (usize, usize) {
+    // Remove requests for spent or active UTXOs and count them.
+    let count_before_filter = candidates.len();
+    candidates.retain(|r| !spent_or_active_utxo_ids.contains(&r.utxo.id));
+    let utxo_already_used_count = count_before_filter - candidates.len();
+
+    // Keep the first request per UTXO and count later duplicates.
+    let count_before_filter = candidates.len();
+    let mut selected_utxo_ids = HashSet::new();
+    candidates.retain(|r| selected_utxo_ids.insert(r.utxo.id));
```
Bridgerz, line 553: keep-first by (created_timestamp_ms, id) picks the winner on the wrong key. confirmability is gated by approved_timestamp_ms + deposit_time_delay_ms further down, so the earliest created request can be one that was just re-approved (delay not elapsed) while the duplicate we drop was immediately confirmable, delaying the user's confirmation by up to the full delay. worst case the winner has approved_timestamp_ms == None and hits the "approval timestamp is missing" warn every tick while masking a confirmable duplicate. picking the winner by earliest approved_timestamp_ms (Some before None) among current epoch certs avoids both.

crates/hashi/src/leader/deposits.rs
```diff
@@ -143,9 +144,29 @@ impl LeaderService {
...
+        );
+        let (utxo_already_used_count, duplicate_utxo_count) =
+            filter_deposit_confirmation_candidates(
+                &mut deposit_confirmation_candidates,
+                &spent_or_active_utxo_ids,
+            );
+
+        self.inner
+            .metrics
+            .approved_deposit_requests_ignored
```
Bridgerz: nit: this gauge only gets set on the happy path while we're leader, so it freezes at its last value on pause/reconfig or when we lose leadership, and a fleet wide sum() counts phantoms. leader_items_in_backoff has the same quirk so not blocking, but since this one's new: worth zeroing both labels on the halt path. also the name reads like a counter, someone will rate() it and get 0.

crates/hashi/src/onchain/mod.rs
```diff
@@ -654,6 +655,19 @@ impl OnchainState {
...
 
+    pub fn find_spent_or_active_utxo_ids(
+        &self,
+        utxo_ids: impl IntoIterator<Item = types::UtxoId>,
+    ) -> HashSet<types::UtxoId> {
+        let mut utxo_ids: HashSet<_> = utxo_ids.into_iter().collect();
+        let state = self.state();
+        let utxo_pool = &state.hashi().utxo_pool;
+        utxo_ids.retain(|id| {
+            utxo_pool.utxo_records().contains_key(id) || utxo_pool.spent_utxos().contains_key(id)
```
Bridgerz: nit: this predicate is a verbatim copy of the DuplicateOrSpentOnSui check in deposits.rs. the two have to stay in lockstep by design (this filter exists to predict exactly that rejection), so maybe a UtxoPool::is_active_or_spent(id) helper next to active_utxos() that both call.

Bridgerz (changes requested):

Bridgerz (approved):
LGTM

PR #880, opened 2026-07-23
Gate `handle_reconfig`'s post-loop registration/backup/prune on `reconfig_target_live`

zhouwfang:
MystenLabs/hashi#163


bmwill:
@zhouwfang is this still needed?

PR #885, opened 2026-07-24
Fix a state-lock re-entry deadlock in committee-transition signing

Bridgerz:
Follow-up to #884: a systematic sweep of the workspace for the same bug class — nested acquisition of the non-reentrant `std::sync::RwLock<State>` behind `OnchainState`, which self-deadlocks the moment the watcher queues a write.

## 1. `withdrawals.rs` — `validate_and_sign_committee_transition()` (reconfig path) — real deadlock

Held the read guard from `let state = onchain.state();` for the entire function, then called `sign_message_proto_at_epoch()`, which re-locks via `self.onchain_state().state()`. Reachable from the `sign_committee_transition` gRPC handler on every committee-handoff signing request that misses the per-epoch signature cache — i.e. at least the first request per epoch on every node, **during reconfig**, exactly when the watcher is busiest writing. Same failure signature as the incident: process alive, node dark.

Per review feedback, the fix is now a shared owned accessor rather than guard scoping: `OnchainState::committee_transition(from_epoch) -> Option<(Committee, Committee)>` centralizes the sparse successor lookup and returns owned clones so no state guard can escape. Both this site and the leader's `collect_committee_transition_signatures` (which duplicated the lookup) now use it.

## 2. `leader/deposits.rs` — `check_halt_deposit_processing()` — consistency cleanup only (NOT a deadlock; original description corrected)

The PR originally claimed the `||`'s first-operand temporary guard outlived the condition and deadlocked against `is_reconfiguring()`. **That was wrong**: per the Rust Reference (Destructors → Temporary scopes), *each operand* of a lazy boolean expression is its own temporary scope, so the guard was dropped before the second operand ran — verified empirically with a `try_write` probe on both editions 2021/2024. Credit to @dlukegordon for catching it. The change is kept purely because evaluating both predicates from one snapshot is more consistent than two independent reads racing the watcher.

## Sweep scope

Six angles over the workspace (read-guard re-entry everywhere, write-guard scopes in the watcher, the `RwLock<MpcManager>` domain, std guards across `.await`, guardian/monitor/screener crates, AB-BA ordering), each candidate adversarially verified. Nothing else survived.

## Validation

- `cargo check -p hashi`, `cargo fmt --check` clean; `cargo test -p hashi --lib` 488 passed (gpg-dependent test skipped per local-env limitation)

### DIFF
--- a/crates/hashi/src/leader/deposits.rs
+++ b/crates/hashi/src/leader/deposits.rs
@@ -285,12 +285,10 @@ impl LeaderService {
     }
 
     fn check_halt_deposit_processing(&mut self) -> bool {
-        // Evaluate both predicates from one scoped snapshot: the state lock
-        // is a non-reentrant `std::sync::RwLock`, and the temporary read
+        // Evaluate both predicates from one scoped snapshot. The temporary
         // guard from the `||`'s first operand lives until the whole condition
-        // finishes evaluating — so calling `is_reconfiguring()` (which
-        // re-locks) in the second operand self-deadlocks as soon as the
-        // watcher has a write queued (readers block behind a waiting writer).
+        // finishes evaluating, so `is_reconfiguring()` in the second operand
+        // would reacquire the same state lock while it is held.
         let halt = {
             let state = self.inner.onchain_state().state();
             state.hashi().config.paused()
--- a/crates/hashi/src/leader/deposits.rs
+++ b/crates/hashi/src/leader/deposits.rs
@@ -285,10 +285,7 @@ impl LeaderService {
     }
 
     fn check_halt_deposit_processing(&mut self) -> bool {
-        // Evaluate both predicates from one scoped snapshot. The temporary
-        // guard from the `||`'s first operand lives until the whole condition
-        // finishes evaluating, so `is_reconfiguring()` in the second operand
-        // would reacquire the same state lock while it is held.
+        // Evaluate both predicates from one consistent state snapshot.
         let halt = {
             let state = self.inner.onchain_state().state();
             state.hashi().config.paused()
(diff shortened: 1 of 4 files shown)


crates/hashi/src/leader/deposits.rs
```diff
@@ -285,8 +285,18 @@ impl LeaderService {
     }
 
     fn check_halt_deposit_processing(&mut self) -> bool {
-        if !(self.inner.onchain_state().state().hashi().config.paused() || self.is_reconfiguring())
-        {
+        // Evaluate both predicates from one scoped snapshot: the state lock
```
dlukegordon: Rust drops temporaries from each lazy Boolean operand before evaluating the next operand, so the original read guard was dropped before is_reconfiguring() ran.

Suggested comment:
```
// Evaluate both predicates from one consistent state snapshot.
```

dlukegordon (approved):

crates/hashi/src/withdrawals.rs
```diff
@@ -838,32 +838,36 @@ impl Hashi {
...
-        if !from_committee
-            .members()
-            .iter()
-            .any(|m| m.validator_address() == validator_address)
-        {
-            anyhow::bail!("not a member of the committee at epoch {from_epoch}");
-        }
+        // Scope the read guard so it drops before signing;
+        // `sign_message_proto_at_epoch` reacquires the same state lock.
+        let transition = {
```
dlukegordon: This sparse transition lookup is duplicated in leader/guardian.rs:347. Consider an owned OnchainState accessor so the lookup rule is centralized and the state guard cannot escape:

```
pub fn committee_transition(
    &self,
    from_epoch: u64,
) -> Option<(Committee, Committee)> {
    let state = self.state();
    let committees = state.hashi().committees.committees();
    let from = committees.get(&from_epoch)?.clone();
    let to = committees.range((from_epoch + 1)..).next()?.1.clone();
    Some((from, to))
}
```

Then both callers can use the owned snapshot before signing or performing RPC.

dlukegordon (changes requested):

dlukegordon (approved):

PR #889, opened 2026-07-24
[withdrawals] Fix CPFP transactions

0xsiddharthks:
## Summary

The CPFP machinery is implemented in the codebase, but it could not fire, for three independent reasons:

- **The boost always computed as zero.** `cpfp_deficit` is `fee_rate × ancestor_weight − ancestor_fee`. The target was the 1 sat/vB floor and a stalled parent had already paid 1 sat/vB, so the shortfall was always nothing and the child paid no boost.
- **Ancestor size was measured off the unsigned transaction.** `build_ancestor_chain` used `build_unsigned_withdrawal_tx`, whose witnesses are empty — 127,060 wu for a 700-input settlement weighing 314,662 wu on chain. Sized against that, a deficit that should be 157,317 sat comes to 16,615: a 1.25 sat/vB package, still stuck.
- **Validators would have refused to sign the rescue.** The fee ceiling was 5× the estimate for the transaction *alone*, but a rescue deliberately overpays to cover its parent — ~163,000 sat against a ~9,700 ceiling. A `TODO` on that check already noted it should evaluate the package.

Each blocks a rescue on its own, so all three had to land together.

## Changes

- Fee floor 1 → 3 sat/vB, overridable via `withdrawal_min_fee_rate_sat_vb` and capped at the high-fee threshold so the clamp cannot invert.
- Ancestor weight computed signed: unsigned weight plus each input's satisfaction weight and the segwit marker/flag.
- Validator fee ceiling evaluated over the whole CPFP package, resolving the `TODO`.
- Coin selection bounded by ancestor package weight (90 kvB). `max_tx_weight` covers only the transaction, so nothing stopped a batch pushing its mempool cluster past Bitcoin's 101 kvB limit (`getmempoolinfo.limitclustersize`) and being rejected outright — which is *how* one stalled settlement blocks everything behind it. The 11 kvB held back leaves headroom for a rescue child.

Economic preferences are exposed to config, protocol constraints are not: the fee floor is a judgement call that may need changing mid-incident, so it is a knob; the package bound encodes what the network will *accept*, so it stays a constant like `max_tx_weight`.

`test_signed_weight_matches_onchain_weight` pins the signed weight of a 700-input settlement to the 314,662 wu actually broadcast. 496 tests pass; clippy, fmt and doc clean.

## Effects on existing stuck tx

[`026dc29c`](https://mempool.space/signet/tx/026dc29cecacf63282f724ae819fdcff2654aaf9ef3f3669da1660c6ddb1aa2a) (78,666 vB at 1.0002 sat/vB) has been unmined for 35+ blocks with withdrawals queued behind it. Its change output is currently unspent on chain, so the next batch that spends it computes a 157,317 sat deficit and lands the package at **3.00 sat/vB** — above the 2.0 sat/vB bar, inside the weight bound, and well under the on-chain `worst_case_network_fee` cap (4,075 sat per request at a 40-request batch).

If a `WithdrawalTransaction` has already been committed against that change output, coin selection skips it (`spent_by` is set) and the rescue instead chains onto *that* transaction's change. `build_ancestor_chain` walks the whole DAG, so the deficit covers the stalled parent and the intermediate child together — a grandchild CPFP rather than a direct one. That works provided the intermediate child is itself broadcastable, i.e. under 22.3 kvB given a 78.7 kvB parent. A full-size batch on top of this particular parent would exceed Bitcoin's cluster limit and be unbroadcastable, which no fee change can unlock; that case needs the ability to abort a committed transaction.

Two caveats for whoever rolls this:

- Needs **≥6 requests in the batch**. Below that the per-request cap binds and every `request_count` in the retry loop fails, so no batch forms. If the guardian limiter is throttling withdrawals out one at a time, check its token balance before concluding the fix did not work.
- The miner fee consumes guardian limiter capacity (`inputs − change`). At a 10 BTC bucket the 157,317 sat deficit is 0.016% of capacity, so this is noise, but it is not zero.

Verified on signet: a CPFP child on a matched 77,524 vB / 1.0 sat/vB transaction, lifting the package to 3.06 sat/vB, got that parent mined in the next block.

## Deferred deliberately

Validators still accept a below-floor fee and do not re-check the package weight, so a faulty leader can still commit a stalling or unbroadcastable transaction. Both are worth fixing and neither is fixed here, because enforcing them mid-rollout is the more dangerous option: a not-yet-upgraded leader would have its commitment refused by upgraded validators, leaving it committed on chain but unsignable with its inputs locked — the same shape as the wedge this PR exists to prevent, and what bit us in #626. They should land once the fleet is uniformly upgraded.

The validator also counts every entry still in `withdrawal_txns` as unconfirmed, so a confirmed-but-not-final ancestor inflates its allowance. That widens only the ceiling; the binding limit is the on-chain per-request cap, which is unchanged.

## Follow-ups

- `unconfirmed_ancestor_weight` counts a shared parent once per input. It fails safe (tighter bound, larger fee) but can overpay enough to trip the per-request cap; deduplicating needs identity on `AncestorTx`.
- Validators check ancestor *depth* but not package *weight*. Not required for correctness — an over-budget batch simply fails to broadcast — but it is the symmetric check.

### DIFF
--- a/crates/hashi/src/withdrawals.rs
+++ b/crates/hashi/src/withdrawals.rs
@@ -1092,10 +1092,8 @@ impl Hashi {
 
     /// The configured fee-rate floor, capped at the high-fee threshold.
     ///
-    /// Leader and validator must derive this identically: the leader builds
-    /// to it and the validator prices its ceiling from it, so an uncapped
-    /// value on either side would put them on different fee policies. It
-    /// also keeps `clamp`, which panics when inverted, well-formed.
+    /// Leader and validator must derive this identically or they price
+    /// fees differently. The cap also keeps `clamp` from inverting.
     fn effective_min_fee_rate(&self) -> FeeRate {
         self.config
             .withdrawal_min_fee_rate()
@@ -1606,17 +1604,14 @@ fn unconfirmed_ancestor_depth(
 /// one [`AncestorTx`] entry with its confirmation count, weight, and fee.
 /// The walk is a BFS over the ancestor DAG, capped at
 /// [`MAX_ANCESTOR_DEPTH`] levels.
-/// Aggregate weight and fee of every unconfirmed ancestor of `records`,
-/// counting each ancestor once.
+/// Aggregate weight and fee of the unconfirmed ancestors of `records`,
+/// counted once each.
 ///
 /// Mirrors `unconfirmed_ancestor_depth`: anything still in
-/// `withdrawal_txns` is treated as unconfirmed, so validation stays
-/// deterministic across the committee with no Bitcoin round-trip. Entries
-/// linger until the finality threshold, so an ancestor already in a block
-/// still contributes a deficit here even though CPFP cannot boost it.
-/// That only widens the fee *ceiling* — the binding limit is the on-chain
-/// per-request cap — but it makes this an upper bound on the leader's
-/// figure rather than a match for it.
+/// `withdrawal_txns` counts as unconfirmed, keeping validation
+/// deterministic without a Bitcoin round-trip. Entries linger until
+/// finality, so this is an upper bound on the leader's figure — it
+/// widens only the ceiling, which the on-chain per-request cap bounds.
 fn unconfirmed_ancestor_package(
     hashi: &Hashi,
     records: &[&UtxoRecord],
@@ -1690,9 +1685,8 @@ fn build_ancestor_chain(
             continue;
         }
 
-        // The ancestor set is a DAG, not a tree: a batch spending two
-        // change outputs of the same parent would otherwise record it
-        // twice and double its weight and fee in the CPFP arithmetic.
+        // A DAG, not a tree: spending two change outputs of one parent
+        // would otherwise record it twice.
         if !seen.insert(wid) {
             continue;
         }
@@ -1977,8 +1971,8 @@ mod tests {
         );
     }
 
-    /// The settlement that stalled on signet: 700 inputs, 71 outputs,
-    /// 314,662 wu on-chain against 127,060 wu unsigned.
+    /// The settlement that stalled on signet: 314,662 wu on-chain
+    /// against 127,060 wu unsigned.
     #[test]
     fn test_signed_weight_matches_onchain_weight() {
         use bitcoin::Amount;
--- a/crates/hashi/src/withdrawals.rs
+++ b/crates/hashi/src/withdrawals.rs
@@ -401,13 +401,10 @@ impl Hashi {
             }
         }
 
-        // 6. Validate fee is reasonable.
-        //
-        // The ceiling is evaluated over the whole CPFP package, not the
-        // transaction alone. A leader spending unconfirmed change must
-        // also cover the shortfall its ancestors owe, so judging the fee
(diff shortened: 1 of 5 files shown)


crates/hashi/src/withdrawals.rs
```diff
@@ -1192,6 +1203,7 @@ impl Hashi {
 
             let params = CoinSelectionParams {
                 max_inputs,
+                min_fee_rate,
                 long_term_fee_rate: configured_long_term_fee_rate,
                 max_fee_per_request: self.onchain_state().worst_case_network_fee(),
```
Bridgerz, line 1216: The CPFP surcharge is charged per-request, so the rescue can't be built when fewer than ~6 withdrawals are pending — which is exactly the drained-queue state a stuck settlement produces. The ~157k sat deficit is folded into total_deduction and check_fees divides it across the batch (div_ceil(n) vs max_fee_per_request ≈ 29,454), so it only clears at n≥6. The retry loop counts down from requests.len(), so with 1–5 pending every iteration fails → UtxoSelectionFailed, and the parent stays stuck until a 6th request happens to arrive. Since the surcharge is a shared network cost rather than a per-user fee, could a rescue batch be allowed to exceed the per-request cap (up to the on-chain ceiling), or pad the batch to reach the count? Otherwise the rescue is unavailable in a chunk of the incident it targets.

0xsiddharthks: You're right, and it's a real gap — but it can't be closed from this side. The per-request cap is enforced on chain, not just here: `withdrawal_queue.move` does

```move
let per_user_miner_fee = miner_fee / request_count;
assert!(per_user_miner_fee <= max_network_fee, EMinerFeeExceedsMax);
```

so letting a rescue batch exceed it client-side just moves the failure from `UtxoSelectionFailed` to a rejected commit. `check_fees` uses `div_ceil` against the same value, which keeps the Rust side strictly stricter than Move — deliberately, so we never sign something the chain refuses.

Padding doesn't work either, since we can't invent withdrawal requests to pad with.

So the fix is on chain: either exempt the CPFP surcharge from the per-user cap (it's a shared network cost, exactly as you say, not a per-user fee), or raise `bitcoin_withdrawal_minimum` so the derived cap is bigger. Both are Move changes, so I'd rather not smuggle them into this PR — I've written the constraint up under Effects so it's explicit rather than buried.

Worth noting the threshold moves with the deficit, so it isn't fixed at 6: it's `ceil(deficit / max_network_fee)`, which for the current stalled settlement is ~6 at a 29,454 sat cap. A smaller stalled parent needs fewer.

0xsiddharthks: Yeah this is an issue, but afaik we can't fix it from the Rust side.

the per-request cap is asserted on-chain (EMinerFeeExceedsMax in withdrawal_queue.move), so exceeding it just turns a selection failure into a rejected commit, and padding needs requests we can't invent.

Maybe lets add it in our list of move changes.
either exempt the CPFP surcharge from the per-user cap or raise bitcoin_withdrawal_minimum.

Bridgerz (approved):

PR #890, opened 2026-07-26
hashi: tag prefetched presig batches so stale refills cannot install

bmwill:
Previously, a presignature refill result was staged as an untagged next_batch, and the install path relabeled whatever was staged as latest.batch_index + 1 without checking what it was generated for. On testnet, a node that rebuilt its SigningManager via sync_if_stale had a stale refill request replayed through the coalescing refill channel, regenerated the already-installed batch 19 from the TOB certs, and staged it; when the real batch-20 generation was delayed by RPC rate limiting, signing crossed the batch boundary and installed the batch-19 duplicate as batch 20. The node then evaluated different nonces than its peers at the same global presig indices, poisoning aggregation network-wide, and reused consumed nonces for new messages, leaking its key shares.

With this commit, the staged batch is a `PrefetchedBatch` tagged with the batch index it was generated for, and three layers refuse the stale path:

- `set_next_batch` discards a result that is not strictly newer than the latest installed batch.
- The install path only promotes a prefetch tagged exactly `latest.batch_index + 1`; a stale tag is dropped so the refill request fetches the right batch, and a future tag is kept while the missing batch is requested explicitly. Signing fails safe with `PoolExhausted` instead of binding wrong nonces.
- `refill_presignatures` skips requests for already-installed batches before running generation, saving the RPC budget whose exhaustion delayed the real batch in the incident.

Each layer logs what it refused (batch indices, size, and pool fingerprint), and accepted prefetches are logged so generation, staging, and install can be correlated. Also add regression tests for the stale stash, the stale relabel, and the future-tagged prefetch.

### DIFF
--- a/crates/hashi/src/mpc/service.rs
+++ b/crates/hashi/src/mpc/service.rs
@@ -930,16 +930,24 @@ impl MpcService {
             .inner
             .signing_manager_for(epoch)
             .ok_or_else(|| anyhow::anyhow!("SigningManager not available for epoch {epoch}"))?;
-        // A request for an already-installed batch can arrive via the
-        // coalescing refill channel (e.g., a pre-rebuild request replayed
-        // after `sync_if_stale` recovered the manager). Regenerating it
-        // would waste RPC budget and stage a batch that `set_next_batch`
-        // discards anyway, so skip it up front.
-        let latest = signing_manager.batch_index();
-        if batch_index <= latest {
+        // Refill requests arrive via a coalescing channel that outlives
+        // manager rebuilds and epoch changes, so a replayed value can name a
+        // batch that is already installed, already staged, or not contiguous
+        // with the installed range. Only the immediately-next batch is
+        // actionable; anything else is skipped before spending generation
+        // work, and the trigger paths re-request the right index on demand.
+        let expected = signing_manager.batch_index() + 1;
+        if batch_index != expected {
             info!(
                 "Skipping presignature refill for batch {batch_index}: \
-                 the latest installed batch is already {latest}"
+                 the next installable batch is {expected}"
+            );
+            return Ok(());
+        }
+        if signing_manager.prefetched_batch_index() == Some(batch_index) {
+            info!(
+                "Skipping presignature refill for batch {batch_index}: \
+                 it is already staged for installation"
             );
             return Ok(());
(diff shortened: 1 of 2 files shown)


crates/hashi/src/mpc/service.rs
```diff
@@ -930,11 +930,24 @@ impl MpcService {
             .inner
             .signing_manager_for(epoch)
             .ok_or_else(|| anyhow::anyhow!("SigningManager not available for epoch {epoch}"))?;
+        // A request for an already-installed batch can arrive via the
+        // coalescing refill channel (e.g., a pre-rebuild request replayed
+        // after `sync_if_stale` recovered the manager). Regenerating it
+        // would waste RPC budget and stage a batch that `set_next_batch`
+        // discards anyway, so skip it up front.
+        let latest = signing_manager.batch_index();
```
dlukegordon: This accepts any batch newer than latest, including an old epoch’s queued request far ahead of a rebuilt manager. That can generate and stage a non-contiguous batch and prune state for the missing batches. Could we require batch_index == latest + 1 here and include the epoch or manager generation in refill requests?

dlukegordon: And this only checks installed batches. Repeated watch notifications for batch N can arrive while N is being generated; after N is staged, latest is still N−1, so N gets generated again and replaces the existing prefetch. Could we also skip requests matching the currently prefetched batch?

benr-ml (approved):

crates/hashi/src/mpc/signing.rs
```diff
@@ -259,7 +301,7 @@ impl SigningManager {
         let state = self.state.read().unwrap();
         let batch_end = state.batches.last().map_or(0, |b| b.end_index());
         match &state.next_batch {
-            Some(next) => batch_end + next.len() as u64,
+            Some(next) => batch_end + next.pool.len() as u64,
```
dlukegordon: This counts a future-tagged prefetch as contiguous capacity. For example, batch 20 staged after batch 0 makes the caller believe batch 1’s indices are available, suppressing the proactive refill. Could we include next.pool.len() only when next.batch_index == latest.batch_index + 1?

dlukegordon (approved):

PR #893, opened 2026-07-27
Nest Move package versions under `packages/hashi`

zhouwfang:
MystenLabs/hashi#892


bmwill:
I wonder if it would make more sense to have the following structure:

packages/hashi/v1
packages/hashi/v2
packages/hashi/vx

zhouwfang:
> I wonder if it would make more sense to have the following structure:
> 
> packages/hashi/v1 packages/hashi/v2 packages/hashi/vx

@bmwill Grouping reads better, yes, but what stops me is that it moves the live path on every freeze.

publish.rs, the CLI default and infra's deploy_script all name the live dir. Miss the infra one when v3 is cut and it still points at packages/hashi/v2, a real frozen package by then, so it builds and publishes fine and devnet quietly runs the previous version.

`packages/hashi` as the live dir can't go stale that way. If you want the grouping, `packages/hashi` + `packages/hashi-frozen/v1` gets it without moving the live path.

bmwill:
>is that it moves the live path on every freeze.

yes that is intentional, you're referencing tooling which predates this directory structure which is easily remedied.

jessiemongeon1:

zhouwfang:
> ## Scope Review: Community Tool Documentation
> This PR adds documentation for a community tool, which is outside the scope of the Hashi docs.
> 
> The Hashi documentation covers the core Hashi protocol, its design, and its official components. Community-built tools, integrations, and third-party projects are not included in the Hashi docs.
> 
> **This PR does not meet the requirements for inclusion in the Hashi documentation.**
> 
> If you believe this is an error, please open an issue in this repo to discuss.

Do you mean the one line change in `design/docs/node-operator-runbook.mdx`? It is required as a consequence of this PR.

zhouwfang:
> > is that it moves the live path on every freeze.
> 
> yes that is intentional, you're referencing tooling which predates this directory structure which is easily remedied.

@bmwill Done. Please take another look.

crates/e2e-tests/src/lib.rs
```diff
@@ -150,6 +153,7 @@ impl TestNetworksBuilder {
             bitcoin_builder: BitcoinNodeBuilder::new(),
             onchain_config_overrides,
             external_guardian: None,
+            move_package_dir: DEFAULT_MOVE_PACKAGE_DIR.to_string(),
```
bmwill, line 156: We have to always start by publishing V1 and then do an upgrade to V2. all of our move type mirroring infrastructure will be dependent on this. If we want to streamline this we can easily do the upgrade before we move the UpgradeCap into the contract though.

bmwill (commented):
We'll also want some infrastructure to check the set of "known" package versions against the set of "allowed" package versions (that is onchain) and if the node doesn't have knowledge of the allowed version it should likely crash loudly.

crates/e2e-tests/src/lib.rs
bmwill: Although we may just want to always test upgrades through the onchain voting mechanism for consistency.

crates/hashi/src/cli/mod.rs
```diff
@@ -631,7 +631,7 @@ pub struct PublishOpts {
     pub sui_rpc_url: String,
 
     /// Path to the Move package directory
-    #[clap(long, short = 'p', default_value = "packages/hashi")]
+    #[clap(long, short = 'p', default_value = "packages/hashi/v2")]
     pub package_path: std::path::PathBuf,
```
bmwill, line 635: publish should likely always default to publishing v1 for now

PR #894, opened 2026-07-27
[cli] scrape only governance state for governance commands

0xsiddharthks:
## Changes

- Add `ScrapeScope { Full, GovernanceOnly }`, threaded through `State::scrape` / `scrape_hashi`.
- Group the three Bitcoin collections into `types::BitcoinCollections` and store them as `Option` on `types::Hashi`, behind `bitcoin()` (panics, for the always-`Full` validator paths) and `try_bitcoin()`.
- Add `OnchainState::new_reader`: scrapes at a scope and starts no watcher, since the CLI is one-shot and was dropping the watcher's `Service` immediately anyway. Both constructors share `scrape_into_state`.
- `GovernanceOnly` also skips the `BitcoinState` lookup, which only exists to hand the collections their table ids and the mirror its Bitcoin-side routing.
- `scrape_hashi` returns the `MirrorSeed` as an `Option` for the same reason: a governance scrape routes none of the Bitcoin containers and folds none of their serving heights into the replay floor, so a mirror bootstrapped from it would silently miss every deposit, withdrawal and UTXO write below that floor. `OnchainState::new` and the mirror's re-bootstrap are the only callers that need one, and both scrape `Full`.
- `HashiClient::new` is now governance-scoped; `new_with_bitcoin_state` is the opt-in for `deposit list/status` and `withdraw list/status`, the only CLI commands that read this state. Their fetchers return `Result` so a scope mistake is a CLI error, not a panic.
- Point the `launch` preflight at the governance scope too.

### DIFF
--- a/crates/hashi/src/onchain/mod.rs
+++ b/crates/hashi/src/onchain/mod.rs
@@ -48,25 +48,13 @@ const BROADCAST_CHANNEL_CAPACITY: usize = 100;
 /// the gRPC decode limit; the SDK still pages through every entry.
 const SCRAPE_PAGE_SIZE: u32 = 1000;
 
-/// How much of the on-chain state a scrape loads.
-///
-/// The Bitcoin-side collections (deposit queue, withdrawal queue, UTXO pool)
-/// live in dynamic-field tables paged [`SCRAPE_PAGE_SIZE`] at a time, so they
-/// dominate the cost of a scrape — a 70k-entry withdrawal queue alone is ~70
-/// extra round-trips, enough to trip a fullnode's rate limiter. Governance
-/// reads (config, committees, proposals) never touch them.
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+/// How much of the on-chain state a scrape loads. The Bitcoin collections are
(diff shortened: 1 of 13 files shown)


crates/hashi/src/onchain/mod.rs
```diff
@@ -164,32 +226,26 @@ impl OnchainState {
 
         let (sender, _) = broadcast::channel(BROADCAST_CHANNEL_CAPACITY);
         let (checkpoint, _) = watch::channel(checkpoint);
-        let (state_watermark, _) = watch::channel(seed.floor);
+        // No seed means no floor to claim; 0 makes a waiter block rather
+        // than see coverage no mirror is maintaining.
+        let (state_watermark, _) = watch::channel(seed.as_ref().map_or(0, |seed| seed.floor));
```
dlukegordon: Using 0 for "no mirror coverage" makes wait_until_checkpoint(0) return immediately and reports nonexistent coverage. Could we represent the watermark as Option<u64> and keep it None when there is no mirror seed, making waits block or return an error until coverage exists?

dlukegordon (approved):

PR #896, opened 2026-07-28
[watcher] Close the object mirror's remaining silent-gap paths

0xsiddharthks:
Follow-ups from the #871 review, all in the mirror's ingest path. Each one is a place the mirror could go wrong without saying so.

- **`PackageWrite` demanded an object nothing reads.** It shared the `ObjectWrite` arm, so an upgrade transaction required the package object from the object set only for pass 3 to discard it (the upgrade comes from the root's `UpgradeCap`). A miss would have failed the frame with an error pointing at the read mask. Skipped now, with `AccumulatorWrite` given its own arm so the catch-all no longer absorbs known states.
- **Unknown `output_state` was dropped silently.** No log, no unrouted entry, no metric — the one path where the mirror could stop applying writes without a signal. Those ids ride the tripwire now. `OutputObjectState` is `#[non_exhaustive]`, so the wildcard also covers whatever a future SDK adds.
- **The tripwire's package set was blind to intermediate upgrades.** It only ever held the original id and the current one from the `UpgradeCap`, so a type introduced by, say, v2 of a three-version package looked foreign and never incremented `hashi_watcher_unrouted_objects_total` — exactly the "we added state and forgot to route it" case it exists to catch. Seeded from the full `PackageVersions` history at bootstrap and re-bootstrap.
- **A re-bootstrap could leave the watermark over-claiming.** The scrape installs state wholesale and can regress it, but `advance_state_watermark` is monotonic, so a scrape served below the previous coverage kept reporting state the new mirror did not have. It resets now, warning when it moves backwards.
- **`withdrawal_signed_at_ms` leaked across re-bootstraps.** The wholesale replace swallows the removal transition of any withdrawal that confirmed during the gap. Pruned against the in-flight bag.
- **Object set decoded before the status check**, so a failed transaction paid for a decode of objects nothing reads and could fail the frame on them. Status first.

Plus an unused parameter and the split `impl TxView`.

New tests cover the package-write skip, the unknown-state tripwire, the failed-transaction short circuit, and the package-history seeding.

## Not included

- **The handoff committee round-trip** (`apply.rs`, `move_types::Committee::from`). The scrape path passes `convert_move_committee_handoff` the raw on-chain committee; apply passes one rebuilt from the enriched type, and `convert_move_committee_member` substitutes `fallback_encryption_public_key()` when the on-chain key won't parse. Move only asserts `length == 32`, not validity. I couldn't tell which path is intended — the same reconstruction is already used when signing (`withdrawals.rs`, `leader/guardian.rs`), so the scrape may be the odd one out. Left alone rather than guessed at; happy to follow up either way.
- **A bound on the replay loop.** #871 fixed the client half by moving to `new_sui_rpc_client`, but `replay_transactions` still has no overall deadline: `LedgerTip` short of target sleeps and retries with no cap, and the live loop is never reached, so `STREAM_STALL_TIMEOUT` can't fire. A stuck transaction index would hang the watcher with a flat `hashi_watcher_state_watermark`. Left out as out of scope here.

### DIFF
--- a/crates/hashi/src/onchain/apply.rs
+++ b/crates/hashi/src/onchain/apply.rs
@@ -50,13 +50,10 @@ pub(super) struct TxView {
     /// advance and the withdrawal duration metrics.
     pub timestamp_ms: u64,
     pub changes: Vec<TxChange>,
-}
-
-impl TxView {
-    /// The transaction's position in the chain's total order.
-    pub(super) fn position(&self) -> (u64, u64) {
-        (self.checkpoint, self.transaction_index)
-    }
+    /// Changed objects whose `output_state` was absent or unrecognized.
+    /// Neither a write nor a removal can be derived from them, so they
+    /// ride the unrouted tripwire rather than being dropped silently.
+    pub unknown_output_states: Vec<Address>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


0xsiddharthks:
Closing cause all the changes have been implemented in 
- #897

PR #898, opened 2026-07-29
Add weight-checked certificate verification to `Committee`

zhouwfang:
#918

### DIFF
--- a/crates/hashi-types/src/committee.rs
+++ b/crates/hashi-types/src/committee.rs
@@ -16,6 +16,7 @@ use fastcrypto::traits::KeyPair;
 use fastcrypto::traits::Signer;
 use fastcrypto::traits::ToFromBytes;
 use fastcrypto::traits::VerifyingKey;
+use fastcrypto_tbls::nodes::Nodes;
 use serde::Deserialize;
 use serde::Serialize;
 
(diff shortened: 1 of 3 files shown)


crates/hashi-types/src/committee.rs
```diff
@@ -290,10 +290,9 @@ impl Committee {
             .map_err(SignatureError::from_source)
     }
 
-    /// Verify an [CommitteeSignature]. If you also need to verify the weight, you can either
-    /// get the weight of the signature with [CommitteeSignature::weight] or use the [Self::verify_signature_and_weight]
-    /// function.
-    pub fn verify_signature<T: IntentMessage>(
+    /// Verify the aggregate signature only — no minimum signer weight, so a
+    /// single-member "certificate" passes. Pair with a weight gate.
+    pub fn verify_signature_any_weight<T: IntentMessage>(
```
benr-ml, line 295: do we need this as a public function?

benr-ml (commented):
those changes are obviously not 100% backward compatible, e.g. an dealer might create a non-weighted cert while new parties might reject it.

zhouwfang:
> those changes are obviously not 100% backward compatible, e.g. an dealer might create a non-weighted cert while new parties might reject it.

@benr-ml I grew this PR based on https://github.com/MystenLabs/hashi/pull/861#discussion_r3677071944. It's the API itself now rather than just the rename.

On backward compat: you're right, but no quorum enforcement is added here. The only adoption is the AVID call site, which already did both checks. It lands in the follow-up that migrates the party loops and recovery.

benr-ml (approved):

PR #901, opened 2026-07-29
Optimize withdrawals

bmwill:


crates/hashi/src/withdrawals.rs
```diff
@@ -128,6 +193,67 @@ fn safe_withdrawal_flow_max_inputs(request_count: usize, configured_max_inputs:
...
+    anyhow::ensure!(
+        input_count <= max_inputs,
+        "Commitment has {input_count} inputs for {request_count} requests, \
+         exceeding the flow cap of {max_inputs}",
+    );
+
+    // The counts above bound the input side, but the change-output count is
+    // otherwise unbounded, so an absurdly shaped commitment could still
+    // describe a transaction too heavy to relay.
+    let tx_weight = estimated_withdrawal_tx_weight(input_count, outputs)?;
```
dlukegordon, line 290: This still permits an arbitrary number of trailing change outputs. Each output adds a PTB command, and each change output creates a pending UTXO object, so validators can certify a commitment that exceeds Sui limits. Can we enforce outputs.len() <= request_count + 1, matching the single-change-output shape produced by coin selection?

crates/e2e-tests/src/e2e_flow.rs
```diff
@@ -2593,4 +2593,288 @@ mod tests {
...
+    ///    handful of available inputs.
+    /// 4. Assert commit, sign, and confirm stay under all Sui limits.
+    /// 5. Mine blocks, wait for confirmation, and run cleanup.
+    #[tokio::test]
+    async fn test_drain_mode_max_request_batch() -> Result<()> {
+        init_test_logging();
+        info!("=== Starting Drain Mode Max Batch Test ===");
+
+        let num_withdrawals: usize = hashi::utxo_pool::CoinSelectionParams::MAX_WITHDRAWAL_REQUESTS;
+        let num_deposits: usize = 8;
```
dlukegordon: The production ceiling allows 16 inputs at 298 requests, but this test exercises only 8. Could we use 16 deposits/selected inputs instead?

crates/hashi/src/withdrawals.rs
```diff
@@ -128,6 +193,67 @@ fn safe_withdrawal_flow_max_inputs(request_count: usize, configured_max_inputs:
...
+    outputs: &[OutputUtxo],
+) -> anyhow::Result<Weight> {
+    let input_weight = hashi_bitcoin::SCRIPT_PATH_2OF2_TXIN_WEIGHT * input_count as u64;
+    let output_weight: u64 = outputs
+        .iter()
+        .map(|o| hashi_bitcoin::output_weight_for_witness_program(&o.bitcoin_address))
+        .collect::<anyhow::Result<Vec<_>>>()?
+        .iter()
+        .sum();
+    Ok(Weight::from_wu(
```
dlukegordon: This estimate omits the CompactSize-encoded input and output counts, underestimating the 298-request shape by 16 WU. Please include both count weights, ideally by sharing the varint_weight calculation used by TransactionBuilder::weight.

crates/hashi/src/withdrawals.rs
bmwill: >Can we enforce outputs.len() <= request_count + 1, matching the single-change-output shape produced by coin selection?

we do want to allow for additional change outputs so that we have the flexibility of splitting utxos in the future, although we could cap it to some small number

jessiemongeon1:

crates/hashi/src/withdrawals.rs
```diff
@@ -1191,6 +1363,22 @@ impl Hashi {
             })
             .collect();
 
+        // Drain versus consolidate: size the batch cap by comparing the
+        // queue depth to the available pool. `candidates` counts every
+        // unlocked UTXO — the same set coin selection draws from.
+        let batch_request_cap = withdrawal_batch_request_cap(requests.len(), candidates.len());
```
dlukegordon, line 1369: `candidates.len()` includes every unlocked UTXO, but `select_coins` later excludes candidates based on mempool depth and ancestor-package limits. Ineligible UTXOs can therefore force consolidation mode even when the usable pool is smaller than the queue. Can we calculate this count using the same eligibility criteria as coin selection?

bmwill: I think in practice this shouldn't really matter much

dlukegordon (approved):

PR #902, opened 2026-07-29
[Guardian] Persist failure log errors as strings

mskd12:
## Summary

- store withdrawal and committee-update failure errors as rendered strings
- remove Serde support from `GuardianError` so the runtime enum cannot be embedded in persisted formats accidentally
- verify that signed failure-log JSON contains string error values

## Compatibility

This intentionally changes the failure-log wire shape before mainnet. Existing testnet failure records use the old structured error representation and will need to be handled in a custom way if they need to be read.

Stacked on #877.

## Verification

- `make fmt`
- `git diff --check`

### DIFF
--- a/crates/hashi-guardian/src/s3_reader/limiter_recovery.rs
+++ b/crates/hashi-guardian/src/s3_reader/limiter_recovery.rs
@@ -203,7 +203,7 @@ mod tests {
         let msg = WithdrawalLogMessage::Failure {
             request_data: StandardWithdrawalRequestWire::from(request_data),
             request_sign,
-            error: GuardianError::RateLimitExceeded,
+            error: GuardianError::RateLimitExceeded.to_string(),
         };
         VerifiedLogRecord {
             object_key: "withdraw/failure.json".to_string(),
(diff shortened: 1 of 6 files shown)


dlukegordon (approved):

dlukegordon:
Post-merge review note: `Display` output is now wire format, and it's inconsistent. Every `GuardianError` variant renders as `VariantName: detail` except `RateLimitExceeded`, which renders as "Rate limit exceeded" (`crates/hashi-types/src/guardian/errors.rs`). Since these strings are now inside the signed BCS preimage and persisted forever, the inconsistency gets baked into mainnet records. Worth normalizing to `RateLimitExceeded` before mainnet — same compatibility window that justified this PR's own breakage.

dlukegordon:
Post-merge review note: nothing documents that `Display` is now load-bearing. `error: String` (now at `crates/hashi-types/src/guardian/log/messages/withdrawal.rs` and `committee_update.rs`) has no doc comment, and `errors.rs` doesn't mention that `Display` output is persisted in signed logs. Someone "improving" an error message later won't realize they're changing a persisted format (harmless today since nothing parses it, but the intent should be recorded). A one-line doc comment on the fields and/or above the `Display` impl would do it.

PR #904, opened 2026-07-29
[Guardian] [1/3] Unify signing and log validation APIs

mskd12:
Unify Guardian and KP signed envelopes around intent-based sign and verify_signature APIs.

Separate response timestamps into GuardianResponse while keeping them inside the signed payload, and represent signed logs as GuardianSigned<LogEntry> alongside explicit unsigned records.

Centralize LogRecord validation, keep the actual S3 key comparison at the read boundary, and move VerifiedLogRecord to the reader that establishes attestation and PCR trust.

Local check: make fmt

### DIFF
--- a/crates/hashi-types/src/guardian/log/record.rs
+++ b/crates/hashi-types/src/guardian/log/record.rs
@@ -772,7 +772,7 @@ mod tests {
             message: &signed.data.message,
         };
         let signed_bytes = bcs::to_bytes(&(
-            GuardianSigningIntentType::LogMessage,
+            GuardianSigningIntentType::LogEntry,
             payload,
             signed.data.timestamp_ms,
         ))
--- a/crates/hashi-types/src/guardian/log/record.rs
+++ b/crates/hashi-types/src/guardian/log/record.rs
@@ -1,8 +1,8 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//! The `LogRecord` envelope written to S3: it wraps a `super::schema::LogMessage`
-//! with the session id, timestamp, and (for signed logs) the guardian signature.
+//! The `LogRecord` written to S3: it wraps a `super::schema::LogMessage` with
+//! its routing context and, for signed logs, a guardian signature.
 //! The object key and lock duration are derived from the wrapped message.
 
 use super::ObjectKeyPattern;
@@ -17,8 +17,6 @@ use crate::guardian::GuardianResult;
 use crate::guardian::GuardianSignKeyPair;
 use crate::guardian::GuardianSignature;
 use crate::guardian::GuardianSigned;
-use crate::guardian::GuardianSigningIntent;
-use crate::guardian::GuardianSigningIntentType;
 use crate::guardian::SessionID;
 use crate::guardian::UnixMillis;
 use crate::guardian::now_timestamp_ms;
@@ -30,7 +28,7 @@ use std::time::Duration;
 
 /// The data authenticated by a signed log record.
 ///
-/// Field order is part of the deployed BCS signing format and must not change.
+/// Field order defines the BCS signing format.
 #[derive(Debug, Serialize)]
 pub struct LogEntry {
     schema_version: u64,
@@ -39,13 +37,7 @@ pub struct LogEntry {
     /// intended key with the actual key returned by S3.
     object_key: String,
     message: VersionedLogMessage,
-}
-
-/// A log entry authenticated by its embedded Nitro attestation instead of a
-/// Guardian signature.
-#[derive(Debug)]
-pub struct UnsignedLogEntry {
-    data: LogEntry,
+    /// Milliseconds since Unix epoch.
     timestamp_ms: UnixMillis,
 }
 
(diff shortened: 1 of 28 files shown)


crates/hashi-types/src/guardian/signing.rs
```diff
@@ -92,89 +65,138 @@ pub(crate) fn verify_intent<T: Serialize + SigningIntent>(
...
 
-impl SigningIntent for RotateKpsResponse {
-    const INTENT: IntentType = IntentType::RotateKpsResponse;
+impl GuardianSigningIntent for LogEntry {
+    const INTENT: GuardianSigningIntentType = GuardianSigningIntentType::LogEntry;
 }
 
-impl SigningIntent for ProvisionerRotateCertResponse {
-    const INTENT: IntentType = IntentType::ProvisionerRotateCertResponse;
+impl GuardianSigningIntent for SetupNewKeyResponse {
```
dlukegordon: Response types can be signed directly without a timestamp, making accidental replay possible. Please implement `GuardianSigningIntent` only for `GuardianResponse<ResponseType>`, so all signed responses must include an authenticated timestamp.

crates/hashi-types/src/guardian/signing.rs
```diff
@@ -92,89 +65,138 @@ pub(crate) fn verify_intent<T: Serialize + SigningIntent>(
...
 
-impl SigningIntent for GuardianInfo {
-    const INTENT: IntentType = IntentType::GuardianInfo;
+/// A signed, timestamped response produced by the Guardian.
+pub type GuardianSignedResponse<T> = GuardianSigned<GuardianResponse<T>>;
+
+/// Guardian-signed wrapper - adds a signature to any signable payload.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct GuardianSigned<T> {
+    pub data: T,
```
dlukegordon: Nothing connects `verify_signature()` to reading `data`. If a caller forgets the verify call, the code still compiles and silently uses unverified data. The old `verify(self) -> T` made that impossible, and public access also makes `into_data_unchecked()` meaningless. Please make `data` private, expose an explicitly unchecked accessor, and restore a consuming `verify(self, key) -> Result<T>` method. The same applies to `KpSigned`.

crates/hashi-types/src/guardian/log/record.rs
```diff
@@ -197,11 +249,79 @@ impl LogRecord {
...
+        match self {
+            Self::Signed(signed) => {
+                let timestamp_ms = signed.data.timestamp_ms;
+                let (session_id, message) = signed.into_data_unchecked().into_current()?;
+                Ok((session_id, timestamp_ms, message))
+            }
+            Self::Unsigned(unsigned) if unsigned.message.is_allowed_unsigned() => Err(
+                InvalidS3Log("expected signed log record but message is unsigned".into()),
+            ),
+            Self::Unsigned(_) => Err(InvalidS3Log("missing log signature".into())),
```
dlukegordon, line 328: These unsigned-record errors duplicate the messages in `validate`, so they can drift apart. Please route the unsigned case through a shared helper so each error is defined in one place.

crates/hashi-types/src/guardian/signing.rs
```diff
@@ -92,89 +65,138 @@ pub(crate) fn verify_intent<T: Serialize + SigningIntent>(
...
-            signature,
         }
     }
 
-    /// Verify signature and extract payload
-    /// Checks intent byte to ensure signature is for the correct type
-    pub fn verify(self, pub_key: &VerificationKey) -> CryptoVerificationResult<T> {
-        verify_intent(&self.data, self.timestamp_ms, &self.signature, pub_key)?;
-        Ok(self.data)
+    pub fn into_response(self) -> T {
```
dlukegordon: `response` is public, but this also provides `into_response()`, and callers mix both forms. Please pick one access pattern. Making the field private would make `into_response()` the consistent API.

crates/hashi-types/src/guardian/signing.rs
mskd12: That was the original intent behind the old API, and I agree it is safer when every caller can follow it. In practice, some legitimate callers intentionally do not verify the Guardian envelope—for example, Hashi nodes authenticate the Guardian over TLS and validate the returned Bitcoin signatures, while parts of the proxy consume Guardian responses only for liveness and routing.

As a result, consuming `verify(self) -> T` could no longer be the universal gateway to the payload. This PR now keeps `data` private: verified callers use `verify_signature()` to borrow the authenticated payload or `verify_into_data()` to consume it, while callers intentionally skipping verification must explicitly use an `_unchecked` accessor. This preserves the safer default while making trust-boundary exceptions visible.

crates/hashi-types/src/guardian/log/record.rs
mskd12: Seems like a nit. i decided to skip as it is not worth the complexity of a helper.

crates/hashi-types/src/guardian/signing.rs
mskd12: Agreed on using one access pattern. Since `GuardianResponse` is just a timestamped data container and does not enforce a verification boundary, I kept `response` public and removed `into_response()`. Callers now access `.response` directly; `GuardianSigned::data` remains private because that is the actual signature boundary.

dlukegordon (approved):

PR #906, opened 2026-07-30
Make Bitcoin deposit checks one-shot

dlukegordon:
This PR makes Bitcoin deposit checks return the current result immediately instead of waiting up to five seconds and retrying inside the monitor. Missing, unconfirmed, and mempool transactions are retried by the leader on the next Bitcoin block.

This reduces the impact of the deposit DoS vector by making fake requests occupy worker slots for roughly tens of milliseconds, rather than five seconds each.

I removed much of the old pending-deposit code because the leader already handles retries. The monitor’s separate queue duplicated that behavior and kept retrying requests after their callers had timed out.

### DIFF
--- a/crates/hashi/examples/testnet_monitor.rs
+++ b/crates/hashi/examples/testnet_monitor.rs
@@ -143,20 +143,38 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 // Parse the OutPoint
                 match bitcoin::OutPoint::from_str(input) {
                     Ok(outpoint) => {
-                        info!("Confirming deposit for OutPoint: {}", outpoint);
+                        info!("Checking deposit for OutPoint: {}", outpoint);
 
-                        // Spawn a task to confirm the deposit
+                        // Spawn a task to check the deposit at the current Bitcoin tip.
                         let client = monitor_client.clone();
                         let confirmations = args.confirmations;
                         tokio::spawn(async move {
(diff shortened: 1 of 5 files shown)


0xsiddharthks:
@dlukegordon I have 2 concerns about the downstream effects of this PR.
basically, i think since this PR removes the 5 second stall (which was kinda acting like a speed limiter), we have a few places where I think we can still stall downstream:

Here are the two places I found: (both are O(N^2) with the deposit queue)

- **The leader copies its whole to-do list every time one deposit finishes.** `process_unapproved_deposit_requests` opens with `self.pending_unapproved_deposit_requests.clone()` (`leader/deposits.rs:99`), and `handle_completed_unapproved_deposit_task` calls it again on every completion (`:254`). So draining N requests means N full clones of an N-element vec, plus the O(N) `retain` at `:234`.

- **The monitor rebuilds a metric from scratch on every observation.** `RecordDepositObservation` calls `rebuild_confirmation_metrics()` per message (`monitor.rs:658`), which walks the entire observation cache. N observations, each O(N). This one also runs on the monitor's event loop, which is what pulls `ConfirmDeposit` off a 100-slot channel, so if it falls behind the channel backs up and `confirm_deposit` starts hitting its 5s timeout on the send. Which is the bug we just fixed.

dlukegordon:
Pushed

0xsiddharthks (approved):

PR #907, opened 2026-07-30
Optimize mpc signing flow

bmwill:


crates/hashi/src/mpc/signing.rs
```diff
@@ -106,6 +123,55 @@ pub struct SigningManager {
...
+    peer_cooldowns: PeerCooldowns,
+    /// Peers that contributed a provably bad partial signature this epoch
+    /// (identified against the RS-recovered polynomial in
+    /// [`try_finalize_signature`]). Their partials are excluded from polling
+    /// and merging for the rest of the epoch: one bad share inside the first
+    /// `threshold` slots fails plain aggregation for every input, so a
+    /// misbehaving peer would otherwise force every signature through the
+    /// (more partials, more rounds) recovery path indefinitely. The set
+    /// resets on reconfig with the manager itself.
+    bad_share_peers: Mutex<HashSet<Address>>,
```
zhouwfang: The exclusion is epoch-permanent. A peer that heals stays excluded until reconfig, and there's no cap on blamed weight, so in a mixed-version divergence both sides blame each other and signing can wedge for the epoch. Time-bound it?

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -64,6 +64,78 @@ fn next_signing_chunk(unsigned: &[u64], chunk_size: usize) -> Vec<u64> {
...
 
+/// Delay before retrying a chunk attempt that made no on-chain progress.
+/// Sized for the common cause: committee members whose mirrors have not
+/// indexed the `WithdrawalTransaction` (or its latest chunk) yet and answer
+/// "not found on-chain" — a few seconds of watcher lag, not a hard failure.
+const CHUNK_RETRY_DELAY: Duration = Duration::from_secs(5);
+
+/// Consecutive chunk attempts without on-chain progress before the signing
+/// task gives up and leaves resumption to a later checkpoint tick.
+const MAX_STALLED_CHUNK_ATTEMPTS: u32 = 3;
```
zhouwfang: 15s of slack, but the logs had "not found on-chain" lasting minutes. Give up on elapsed time instead of attempt count?

zhouwfang (commented):
Initial comments

crates/hashi/src/mpc/signing.rs
bmwill: Yeah we could time bound it, but the question is for how long and does it really buy us much? right now it is time bounded by ~24h at most, and its in-memory state so its cleared on restart as well. We could clear it roughly every 1h but i wouldn't want to clear it any more frequent than that.

crates/hashi/src/leader/withdrawal_transactions.rs
bmwill: sorry not sure what you'd prefer instead?

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -64,6 +64,78 @@ fn next_signing_chunk(unsigned: &[u64], chunk_size: usize) -> Vec<u64> {
...
+
+/// Number of committee members asked for a chunk's MPC signatures up front.
+/// Any single synced member can serve the whole chunk — each response
+/// carries full aggregated signatures per input, and the leader keeps the
+/// first valid one — so asking the whole committee buys no redundancy and
+/// multiplies into an all-to-all partial-signature storm on the member side
+/// (every asked member polls every other member for partials).
+const MPC_SIGNING_FANOUT: usize = 3;
```
zhouwfang: I don't think "one healthy member suffices per chunk" holds. handle_get_partial_signatures_request only reads the cache, so a member has partials solely because we asked it to sign. With a 3-member wave only those 3 generate anything, they can't reach threshold, and we stall the full 5s before asking all nodes anyway. Net +5s per chunk, same storm. Either size the wave by weight, or drop the hedge to 1s and widen geometrically like hedged_retrieve.

crates/hashi/src/mpc/signing.rs
zhouwfang, line 531: Blamed and cooled peers aren't filtered here, so they stay in peers_remaining on every later sign() and peers_exhausted never becomes true. Fast-fail is gone and each call burns the full 30s deadline re-aggregating. The scrub in blame_bad_shares only covers the batch in flight.

crates/hashi/src/mpc/signing.rs
```diff
@@ -38,6 +40,21 @@ use crate::mpc::types::SigningResult;
...
+const PARTIAL_SIGS_CALL_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Transport retries within one round's poll of a single peer (on top of the
+/// first attempt). Kept minimal for the same reason as the call timeout.
+const PARTIAL_SIGS_CALL_RETRIES: usize = 1;
+
+/// How long a peer whose poll hard-failed (connect/TLS/timeout) is skipped
+/// before being probed again. Bounds the cost of dead peers to one probe per
+/// cooldown window instead of one per round per concurrent signing task.
+const PARTIAL_SIGS_PEER_COOLDOWN: Duration = Duration::from_secs(30);
```
zhouwfang: Cooldown is 30s and the signing deadline is also 30s, so one bad round means the next call polls zero peers and just times out. Also worth not cooling on Status::unavailable. That's a node still reconciling after an epoch flip, which is a lot of the fleet right after reconfig.

crates/hashi/src/leader/withdrawal_transactions.rs
bmwill: yep, i'll just revert this commit and leave the existing setup as-is

crates/hashi/src/mpc/signing.rs
```diff
@@ -742,68 +838,204 @@ async fn try_finalize_signature(
...
+                        {
+                            let st = &mut pending[i];
+                            for eval in sigs {
+                                // First writer wins per share index: a
+                                // duplicate would fail aggregation outright
+                                // (`aggregate_signatures` rejects duplicate
+                                // indices), and recording the accepted
+                                // value's actual sender is what keeps blame
+                                // attribution honest.
+                                if let std::collections::hash_map::Entry::Vacant(slot) =
```
zhouwfang: A peer that answers first with garbage for every index pushes us into TooManyErrors, so we never reach Done and blame_bad_shares never runs: it poisons every chunk and is never blamed or cooled. Checking eval.index \in share_ids_of(peer) would make attribution authoritative and close it.

crates/hashi/src/leader/withdrawal_transactions.rs
zhouwfang: What about tracking last_progress? Instant instead of a counter. Reset it when unsigned drops, give up after 60/90s stale. Attempts vary wildly in duration (fast "not found" failures burn all 3 in ~15s; slow ones take ~3min), so the count doesn't mean much on its own.

Same clock could bound the whole loop too, so one big txn can't hold the max_concurrent=1 slot for hours.

crates/hashi/src/mpc/rpc/p2p_channel.rs
```diff
@@ -58,6 +58,16 @@ impl RpcP2PChannel {
     }
 }
 
+/// Preserves the one status distinction callers care about: `unavailable`
+/// means the peer is up but not ready yet (e.g. reconciling after an epoch
+/// change) and should be retried shortly rather than treated as failed.
+fn map_status(status: tonic::Status) -> ChannelError {
+    match status.code() {
+        tonic::Code::Unavailable => ChannelError::Unavailable(status.to_string()),
```
zhouwfang: By "don't cool on unavailable", I meant the app-level not-ready case, not the gRPC code. tonic also maps connection-refused, REFUSED_STREAM and 503/504 to Unavailable, so dead peers now never get cooled and never increment mpc_partial_sig_poll_failures_total, which is the metric we'd use to find them.

crates/hashi/src/mpc/signing.rs
```diff
@@ -742,68 +894,227 @@ async fn try_finalize_signature(
...
+        deadline: Instant,
+        metrics: &Metrics,
+    ) -> bool {
+        let threshold = self.config.threshold;
+        let now = Instant::now();
+        let blamed = self.bad_share_peers.snapshot(now);
+        let mut peer_ids: HashMap<Address, Vec<Address>> = HashMap::new();
+        for st in pending.iter() {
+            for peer in &st.peers_remaining {
+                if !blamed.contains(peer) && !self.peer_cooldowns.is_cooling(peer, now) {
```
zhouwfang: Blamed peers are filtered out of peer_ids but not removed from st.peers_remaining, so a session in flight when another task blames someone never hits peers_exhausted and waits out the full deadline.

crates/hashi/src/mpc/signing.rs
```diff
@@ -742,68 +894,227 @@ async fn try_finalize_signature(
...
-        match result {
-            Ok(response) => {
-                for (signing_id, sigs) in response.partial_sigs {
-                    if let Some(&i) = index.get(&signing_id)
-                        && pending[i].peers_remaining.remove(&peer)
-                    {
-                        pending[i].partials.extend(sigs);
-                        progressed = true;
+        }
+        if peer_ids.is_empty() {
```
zhouwfang: The all-cooling fallback fires more than intended. Healthy peers leave peers_remaining as soon as they answer, so near the end of most calls only cooling ones are left and we re-poll them anyway. Keying it on "nobody answered last round" would keep it to the blip case.

crates/hashi/src/mpc/signing.rs
```diff
@@ -742,68 +902,241 @@ async fn try_finalize_signature(
...
+                                // if the owner sends an index twice.
+                                if st.partials.iter().any(|e| e.index == eval.index) {
+                                    continue;
+                                }
+                                st.partials.push(eval);
+                                progressed = true;
+                            }
+                        }
+                    }
+                    if pending
```
dlukegordon: This can stop the round after a fast empty response even when the current threshold shares already failed aggregation and we need extra shares for RS recovery. That can repeatedly cancel slower peers with useful shares until the deadline. Could we track the next required share count after aggregation fails and only exit once that target is reached?

crates/hashi/src/mpc/signing.rs
```diff
@@ -742,68 +902,241 @@ async fn try_finalize_signature(
...
-        .collect();
-    let mut progressed = false;
-    while let Some((peer, result)) = in_flight.next().await {
-        match result {
-            Ok(response) => {
-                for (signing_id, sigs) in response.partial_sigs {
-                    if let Some(&i) = index.get(&signing_id)
-                        && pending[i].peers_remaining.remove(&peer)
+        }
+        if peer_ids.is_empty() && !session_has_responses {
```
dlukegordon: This fallback effectively disables cooldowns when every peer hard-fails, causing repeated all-to-all polling during an outage. Since NotReady already handles reconciling peers separately, could we remove this fallback for hard failures and sleep until the earliest cooldown expires?

crates/hashi/src/leader/withdrawal_transactions.rs
```diff
@@ -526,16 +600,110 @@ impl LeaderService {
...
+            }
+            let checkpoint_height = inner.onchain_state().latest_checkpoint_height();
+            if !super::LeaderService::node_is_leader(inner, checkpoint_height) {
+                info!(
+                    withdrawal_txn_id = %txn_id,
+                    "No longer the leader; stopping the chunk-signing loop"
+                );
+                return Ok(());
+            }
+            if inner.onchain_state().epoch() != txn.signing.epoch {
```
dlukegordon: There’s an epoch race here: after refresh, txn can belong to the new epoch, so this check passes, but members was captured before the loop from the old committee. Could we pin the starting epoch and exit on any change, or refresh the member list together with txn?

dlukegordon (approved):

PR #908, opened 2026-07-30
[Guardian] [3/3] Make S3 reader verification more robust

mskd12:
Stacked on #931.

## Summary

- Move validate-then-consume extraction onto `LogRecord::validate_into_entry` and centralize verified-record reads in `GuardianReader`.
- Make verified session information own record verification, flatten session caching into `GuardianReader`, and keep verified fields behind accessors.
- Make allowlisted builds the default for historical reads while exposing current-build requirements through explicitly named methods and keeping `BuildPolicy` private.
- Update guardian-init and guardian consumers to use the narrower reader APIs.
- Simplify latest ceremony and committee selection, centralize log-directory prefixes on their message types, and move heartbeat/withdrawal cursor construction onto `S3HourScopedDirectory`.
- Move `s3_reader.rs` to `s3_reader/mod.rs`.

## Local validation

- `make fmt`

### DIFF
--- a/crates/hashi-guardian-proxy/src/widlog.rs
+++ b/crates/hashi-guardian-proxy/src/widlog.rs
@@ -20,7 +20,6 @@
 use crate::metrics::ProxyMetrics;
 use aws_sdk_s3::error::DisplayErrorContext;
 use hashi_types::guardian::log::S3_DIR_WITHDRAW;
-use hashi_types::guardian::GuardianError::InvalidS3Log;
 use hashi_types::guardian::LogMessageV1;
 use hashi_types::guardian::LogMessageV2;
 use hashi_types::guardian::LogRecord;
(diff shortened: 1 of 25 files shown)


crates/hashi-guardian-proxy/src/widlog.rs
```diff
@@ -169,14 +168,6 @@ fn parse_success_seq(key: &str) -> Option<u64> {
...
     let record: LogRecord = serde_json::from_slice(bytes)?;
-    if let LogRecord::Unsigned(entry) = &record {
-        let message = if entry.message().is_allowed_unsigned() {
-            "expected signed log record but message is unsigned"
-        } else {
-            "missing log signature"
-        };
-        return Err(InvalidS3Log(message.into()).into());
-    }
     let entry = record.into_entry_unchecked();
```
dlukegordon, line 171: This now accepts withdrawal records with no signature because into_entry_unchecked() also extracts unsigned records. Can we restore the previous LogRecord::Unsigned check before extracting the entry?

dlukegordon (approved):

crates/hashi-guardian-proxy/src/widlog.rs
mskd12: The code only allows one record type (OIAttestation) to be unsigned - so i don't think this check matters.

mskd12: Upon second thoughts, I have realized that the LogRecord deserializer can be strengthened, e.g., the above check should ideally happen inside it. Thinking about how to fix it..

PR #912, opened 2026-07-30
Track pending Bitcoin deposits without repeated RPC polling

dlukegordon:


crates/hashi/src/btc_monitor/monitor.rs
```diff
@@ -855,341 +1167,399 @@ impl Monitor {
...
+        })
+        .await
+        {
+            Ok(tx_info) => tx_info,
+            Err(corepc_client::client_sync::Error::JsonRpc(jsonrpc::error::Error::Rpc(ref e)))
+                if e.code == -5 =>
+            {
+                require_core_checkpoint(&bitcoind_rpc, tip, "after transaction lookup").await?;
+                return Ok(DepositStatus::NotFound);
             }
```
0xsiddharthks: Core returns `-5` for "-txindex off" and "txindex still indexing" too, not just "No such mempool or blockchain transaction". Same code, different message.

Pre-PR that was harmless (re-checked every block). Now it caches `NotFound` for a confirmed deposit, nothing re-discovers it, and that node refuses to sign until kyoto restarts or a reorg. A few validators reindexing at the same time = no quorum on every deposit discovered in that window.

Match the message so the other `-5`s fall through to the error arm and get retried:

```suggestion
            // -5 also covers "-txindex off" and "txindex still indexing";
            // only the ready-index message is a durable not-found.
            Err(corepc_client::client_sync::Error::JsonRpc(jsonrpc::error::Error::Rpc(ref e)))
                if e.code == -5
                    && e.message
                        .starts_with("No such mempool or blockchain transaction") =>
            {
                require_core_checkpoint(&bitcoind_rpc, tip, "after transaction lookup").await?;
                return Ok(DepositStatus::NotFound);
            }
```

(`getindexinfo` → txindex `synced` before caching would also work. Optional: re-discover passive entries every ~100 blocks as a safety net for any future scan bug.)

crates/hashi/src/btc_monitor/monitor.rs
```diff
@@ -855,341 +1167,399 @@ impl Monitor {
...
+    let bitcoind_tip = get_best_block(&bitcoind_rpc)
+        .await
+        .map_err(|e| anyhow::anyhow!("Failed to read Bitcoin Core tip: {e}"))?;
+    if bitcoind_tip != tip.hash {
+        return Err(anyhow::anyhow!(
+            "Bitcoin Core tip is {bitcoind_tip}, but captured Kyoto tip is {}",
+            tip.hash,
+        )
+        .into());
+    }
```
0xsiddharthks: This requires bitcoind's best block to be *exactly* kyoto's tip. The check fires right when kyoto connects a block (`set_tip` → tracker notify → leader reload → task), which is exactly when the two are most likely a block apart, and both directions fail.

The retry only works today because kyoto re-emits `FiltersSynced` after every block, so `process_synced` bumps the sequence a second time a few seconds later. That's accidental, and a validator whose kyoto peers aren't its RPC node and whose bitcoind lags by a few seconds never passes this at all.

Equality isn't needed: bitcoind on kyoto's chain at or past `tip` is enough (unspent at a descendant ⇒ unspent at `tip`; spent at a descendant is a real spend). `require_core_checkpoint` already says that, and the two `bitcoind_tip` comparisons below still pin both `gettxout` answers to the captured tip:

```suggestion
    // Bitcoin Core only needs to be on Kyoto's chain at or past `tip`: unspent
    // at a descendant implies unspent at `tip`, and a spend at a descendant is
    // real. Exact equality fails whenever the two nodes are a block apart.
    require_core_checkpoint(&bitcoind_rpc, tip, "before gettxout").await?;
```

0xsiddharthks (approved):
Went through the reorg / kyoto-restart / generation paths and the leader side. Design is sound: a fake request costs one discovery and then goes passive, and I couldn't get a stale discovery to overwrite a scan result or a block to go unscanned while an entry is passive. The `synced` gate on `confirm_deposit` is what makes the catch-up window safe.

One property to keep in mind: `NotFound`/`InMempool` are permanent until a kyoto restart or reorg, nothing re-discovers them. A wrong passive status is a stuck deposit with no self-healing. Two spots where that bites, inline:

1. `-5` from `getrawtransaction` isn't only "no such tx" — I'd fix this one before merging.
2. `check_unspent_at_tip` requires bitcoind tip == kyoto tip exactly (fragile).

PR #913, opened 2026-07-30
ci: gate Move PRs on package upgrade compatibility

Bridgerz:
Adds a **required-check-friendly, hermetic** CI gate that verifies the current `packages/hashi` source is a **compatible upgrade** of the deployed Sui package — running only when Move files change, with **no network call in CI**.

## How it works
The deployed package's compiled bytecode is **checked in** as a snapshot (`crates/hashi/tests/move_upgrade_snapshots/testnet/v1/` — 33 `.mv` modules + `manifest.json`), kept as test data under the crate (outside `packages/` so Move tooling never touches it). The check (`crates/hashi/tests/move_upgrade_compat.rs`):
1. Builds the current source (reusing `hashi::publish::build_package`, **environment-agnostic** → modules at `0x0`).
2. Loads the checked-in snapshot modules.
3. Rebases the built modules onto the snapshot's runtime address (mirroring the adapter's `substitute_package_id`), then runs `move_binary_format::Compatibility::upgrade_check().check()` per module — the exact logic the validator runs for an `Upgrade` command under the default `Compatible` policy.

Includes a **negative test** (removes a struct field from a synthetic module, asserts rejection — proves the gate catches breaks) and a positive test. `move-binary-format`/`move-core-types` are dev-deps pinned to the workspace's sui rev.

### Why snapshot instead of `sui client upgrade` or querying chain
- `sui client upgrade`'s built-in check needs a wallet-held `UpgradeCap`; ours is custodied in a shared object, so that path isn't available. We reproduce the validator's check directly.
- A checked-in bytecode snapshot keeps CI hermetic — no dependency on a fullnode being reachable inside a required correctness gate. The chain is queried **once** at snapshot-creation time; the regenerate recipe is in `crates/hashi/tests/move_upgrade_snapshots/README.md`.

## Workflow (`move.yml`)
`detect` (paths-filter) → `check-formatting`/`test-move-packages`/`move-compat` gated on Move changes → `move-compat-gate` (`if: always()`). No trigger-level `paths:`, so a single required check works for every PR: green on Rust-only PRs (nothing to check), red only when Move changed and the compat job didn't succeed.

## Verification
Because it's hermetic, the full gate runs and passes **locally**, not just in CI:
```
synthetic_identical_module_is_compatible ... ok
synthetic_incompatible_change_is_rejected ... ok
checking compatibility against runtime address fcea10ca...7f4360: 33 snapshot module(s) vs 33 freshly-built module(s)
OK: current source is a compatible upgrade of the deployed package
test result: ok. 3 passed; 0 failed
```
CI green: `Detect`, `Formatting`, `Build & Test`, `Upgrade compatibility`, `Move compat gate`, `lint`.

## For whoever configures branch protection
- Mark **`Move compat gate`** (not `Upgrade compatibility`) as the required status check.
- `TODO(mainnet)`: add `move_upgrade_snapshots/mainnet/vN/` + the mainnet object id once mainnet deploys (regen command in the README). Regenerate the snapshot after every deployed upgrade.


.github/workflows/move.yml
```diff
@@ -35,3 +61,44 @@ jobs:
...
+  # See crates/hashi/tests/move_upgrade_compat.rs.
+  move-compat:
+    name: Upgrade compatibility
+    needs: detect
+    if: needs.detect.outputs.move == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - run: cargo install --git https://github.com/Mystenlabs/suiup.git --locked
```
dlukegordon: The checker uses `move-binary-format` from fixed Sui commit `77a42de…`, but this installs the latest `suiup` and Sui compiler. Please pin the installer and compiler versions so bytecode generation and compatibility checking do not change independently.

crates/hashi/tests/move_upgrade_compat.rs
```diff
@@ -0,0 +1,429 @@
...
+    std::fs::write(&client_yaml, contents).context("writing client.yaml")?;
+
+    Ok((dir, client_yaml))
+}
+
+/// Directory holding the checked-in bytecode snapshot of the deployed package.
+///
+/// Defaults to `<CARGO_MANIFEST_DIR>/tests/move_upgrade_snapshots/testnet/v1`;
+/// override with `HASHI_COMPAT_SNAPSHOT_DIR`.
+fn snapshot_dir() -> PathBuf {
```
dlukegordon: Hardcoding `testnet/v1` allows this gate to keep checking an obsolete package after an upgrade. Please derive or validate the snapshot network/version/package IDs against `packages/hashi/Published.toml`.

.github/workflows/move.yml
```diff
@@ -35,3 +61,44 @@ jobs:
...
+  # succeed. This lets a single required check work for every PR.
+  move-compat-gate:
+    name: Move compat gate
+    needs: [detect, move-compat]
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - name: Evaluate gate
+        run: |
+          if [ "${{ needs.detect.outputs.move }}" == "true" ] && [ "${{ needs.move-compat.result }}" != "success" ]; then
```
dlukegordon: This fails open when `detect` fails or is cancelled because the missing `move` output is treated like `false`. Please first require `needs.detect.result == 'success'`, then evaluate whether Move changed.

crates/hashi/tests/move_upgrade_compat.rs
```diff
@@ -0,0 +1,429 @@
...
+        compat.check(old_module, &new_module).map_err(|e| {
+            anyhow::anyhow!(
+                "module `{name}` is NOT a compatible upgrade of the deployed version: {e:?}. \
+                 A compatible (default `UpgradeCap` policy) upgrade may not change public \
+                 function signatures, struct/enum layouts or abilities, etc."
+            )
+        })?;
+    }
+
+    Ok(())
```
dlukegordon, line 196: Newly added modules are accepted unconditionally, but current Sui validators reject new modules containing `init`. Please inspect the remaining new modules for `init` and add positive/negative tests for that behavior.

.github/workflows/move.yml
```diff
@@ -10,17 +15,38 @@ env:
...
+    runs-on: ubuntu-latest
+    outputs:
+      move: ${{ steps.filter.outputs.move }}
+    steps:
+      - uses: actions/checkout@v7
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            move:
```
dlukegordon, line 47: The filter excludes the compatibility implementation and fixtures, so changes to them skip this job and receive a green gate. Please include the test, snapshots, `crates/hashi/src/publish.rs`, `crates/hashi/Cargo.toml`, and `Cargo.lock`.

crates/hashi/tests/move_upgrade_compat.rs
```diff
@@ -0,0 +1,429 @@
...
+    let mut mv_paths: Vec<PathBuf> = std::fs::read_dir(&dir)
+        .with_context(|| format!("reading snapshot directory {}", dir.display()))?
+        .map(|entry| entry.map(|e| e.path()).map_err(anyhow::Error::from))
+        .collect::<Result<Vec<_>>>()?
+        .into_iter()
+        .filter(|p| p.extension().and_then(|e| e.to_str()) == Some("mv"))
+        .collect();
+    mv_paths.sort();
+
+    anyhow::ensure!(
```
dlukegordon, line 447: A nonempty snapshot can still be incomplete, causing omitted modules to bypass compatibility checking. Please parse `manifest.json` and require its count/list to exactly match filenames and deserialized module names.

.github/workflows/move.yml
```diff
@@ -8,30 +13,124 @@ permissions:
...
+      - run: cargo install --git https://github.com/Mystenlabs/suiup.git --tag "$SUIUP_PIN" --locked
+      - run: suiup install "sui@$SUI_PIN" -y; suiup install "mvr@$MVR_PIN" -y
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+      - run: sui --version; mvr --version
+      # Runs the snapshot-based gate plus the network-free machinery self-tests.
+      # SUI_BINARY / HASHI_COMPAT_SNAPSHOT_DIR can be overridden here if the
+      # defaults ever need to change.
+      - name: Run upgrade compatibility check
+        run: cargo test -p hashi --test move_upgrade_compat -- --nocapture
```
dlukegordon: Can we add --locked pls

crates/hashi/tests/move_upgrade_snapshots/README.md
```diff
@@ -0,0 +1,95 @@
...
+mainnet deployment is picked up automatically once its entry lands. The
+`HASHI_COMPAT_SNAPSHOT_DIR` environment variable is a dev escape hatch that
+checks exactly one directory instead.
+
+## When to regenerate
+
+Regenerate whenever a **new package version is deployed on chain** (i.e. an
+`Upgrade` is executed). Create a new `v<N+1>/` directory for it (bump the
+version rather than overwriting, so the history of deployed bytecode is
+preserved), update `manifest.json`, and point the test's default at the new
```
dlukegordon: Snapshot selection is now derived from Published.toml, so there is no test default to update. Please replace this with instructions to update the corresponding Published.toml version and add its v<N> snapshot directory.

dlukegordon (approved):

PR #916, opened 2026-07-31
test(e2e): deployed bytecode snapshot -> current source upgrade

Bridgerz:
## What this adds

An e2e-test harness seam that publishes the **checked-in Move bytecode snapshot** of the deployed testnet package into a local Sui net as v1, then upgrades it to the current `packages/hashi` source through the ordinary governance flow. This is the real "deployed bytecode → current source upgrade works end-to-end" test.

It's a strictly stronger claim than the static compatibility gate on the base branch: that gate only normalizes-and-diffs the modules; this actually publishes the deployed bytecode and swaps in the current source against a running validator.

### Stacked on #913

Base branch is `move-compat-ci-gate` (#913). This PR reuses the same checked-in snapshot that #913's compat gate introduced (`crates/hashi/tests/move_upgrade_snapshots/testnet/v1/`). Review/merge #913 first.

## Changes

- **`crates/e2e-tests/src/snapshot.rs`** (new): reusable snapshot module.
  - `load_snapshot_publish(dir) -> Publish`: read every `*.mv`, derive the runtime id from the bytecode, rebase each module self-address `runtime -> 0x0`, re-serialize at each module's original binary version, deps `[0x1,0x2,0x3]`.
  - `publish_snapshot(dir, client, key) -> PublishOutput`: mirrors `publish.rs`, calls `hashi::publish::publish_package`.
  - `default_snapshot_dir()`: the checked-in testnet v1 snapshot.
- **`crates/e2e-tests/src/lib.rs`**: `TestNetworksBuilder::with_v1_from_snapshot(dir)` seam. When set, `build()` publishes v1 from the snapshot instead of a source build; everything else (genesis, DKG, committee, launch) is unchanged. Default behavior (source publish) is unchanged when the seam isn't used.
- **`crates/e2e-tests/tests/upgrade_tests.rs`**: new `snapshot_v1_upgrades_to_current_source` test — stands up a 4-node net with v1-from-snapshot, waits for DKG, runs `execute_full_upgrade` to the current source, asserts effects success, new package id, all nodes' watchers pick up the new version, and the v2-only `upgrade_canary::version()` is callable.
- **`crates/e2e-tests/Cargo.toml`**: add `move-binary-format` / `move-core-types` (pinned to the same sui rev `77a42de4…` as the hashi crate's dev-deps), needed to deserialize/rebase/re-serialize the snapshot.

## Mechanism (folded from a proven PoC)

The on-chain `module_map` (and thus the snapshot) stores modules self-addressed at the package's runtime id. A fresh `publish` on Sui assigns a brand-new object id and expects incoming modules to be `0x0`-addressed (what a source build of an upgradeable package emits) — so we rebase `runtime -> 0x0` (the inverse of the compat test's `substitute_self_address`). Re-serialization uses `serialize_with_version(module.version, ..)` to preserve the original binary version (not `serialize()`, which forces `VERSION_MAX`). The upgrade to v2 goes through the existing governance flow, whose digest is computed by `sui move build`, so no manual upgrade-digest is needed.

## Verify (local + hermetic; needs only the `sui` CLI)

- `cargo build -p e2e-tests` / `cargo test -p e2e-tests --no-run`: clean.
- `cargo fmt --check`: clean.
- `snapshot_v1_upgrades_to_current_source`: **PASSED** (47.6s). Real run:
  - v1 published from snapshot: `0x96e59206…73bf61bf` (fresh id, not the testnet runtime id)
  - DKG / committee formed, 4/4 nodes voted the upgrade proposal
  - upgrade execute+publish+finalize: effects success → new package `0xfdd64032…bef2d8cf`
  - all 4 nodes' watchers logged "package upgraded; version map extended … version=2"
  - v2-only `upgrade_canary::version()` call: effects success
- Regression: the pre-existing `test_upgrade_v1_to_v2` (source-publish path) still **PASSED** (67.2s), confirming the default behavior is unchanged.

### DIFF
--- a/crates/e2e-tests/Cargo.toml
+++ b/crates/e2e-tests/Cargo.toml
@@ -27,10 +27,9 @@ sui-crypto.workspace = true
 sui-sdk-types.workspace = true
 sui-transaction-builder.workspace = true
 # Needed by src/snapshot.rs to deserialize/rebase/re-serialize the checked-in
-# Move bytecode snapshot before publishing it. Pinned to the same sui rev as the
-# hashi crate's move-binary-format/move-core-types dev-deps.
-move-binary-format = { git = "https://github.com/MystenLabs/sui.git", rev = "77a42de4be99383b98518541e0c6002d720cafc8" }
-move-core-types = { git = "https://github.com/MystenLabs/sui.git", rev = "77a42de4be99383b98518541e0c6002d720cafc8" }
+# Move bytecode snapshot before publishing it.
+move-binary-format.workspace = true
+move-core-types.workspace = true
 base64ct.workspace = true
 serde_yaml.workspace = true
 serde.workspace = true
--- a/crates/e2e-tests/Cargo.toml
+++ b/crates/e2e-tests/Cargo.toml
@@ -26,6 +26,11 @@ sui-rpc.workspace = true
 sui-crypto.workspace = true
 sui-sdk-types.workspace = true
 sui-transaction-builder.workspace = true
+# Needed by src/snapshot.rs to deserialize/rebase/re-serialize the checked-in
+# Move bytecode snapshot before publishing it. Pinned to the same sui rev as the
+# hashi crate's move-binary-format/move-core-types dev-deps.
+move-binary-format = { git = "https://github.com/MystenLabs/sui.git", rev = "77a42de4be99383b98518541e0c6002d720cafc8" }
+move-core-types = { git = "https://github.com/MystenLabs/sui.git", rev = "77a42de4be99383b98518541e0c6002d720cafc8" }
 base64ct.workspace = true
 serde_yaml.workspace = true
 serde.workspace = true
--- a/crates/e2e-tests/src/snapshot.rs
+++ b/crates/e2e-tests/src/snapshot.rs
@@ -1,0 +1,197 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Publish a checked-in Move *bytecode snapshot* into a local Sui network.
+//!
+//! Hashi keeps a checked-in bytecode snapshot of the deployed package
+//! (`crates/hashi/tests/move_upgrade_snapshots/<network>/v<version>/`) so the
+//! upgrade-compatibility CI gate can run hermetically. This module reuses that
+//! same snapshot for a *stronger* test: publish the deployed v1 bytecode into a
+//! fresh local net and then upgrade it to the current source, proving the real
+//! "deployed bytecode → current source" upgrade end-to-end (not just the static
+//! compatibility check).
+//!
+//! ## Mechanism
+//!
+//! 1. Read every `*.mv` in the snapshot directory.
+//! 2. Rebase each module's self-address from the package's *runtime* id to
+//!    `0x0`. This is the **inverse** of the compat test's
+//!    `substitute_self_address`: the on-chain `module_map` (and thus the
+//!    snapshot) stores modules self-addressed at the runtime id, while a fresh
+//!    `publish` on Sui assigns a brand-new object id and expects the incoming
+//!    modules to be self-addressed at `0x0` (what a source build of an
+//!    upgradeable package emits).
+//! 3. Re-serialize at each module's *original* binary version (via
+//!    `serialize_with_version`, NOT `serialize()` — the latter forces
+//!    `VERSION_MAX`, which could exceed what the local net accepts).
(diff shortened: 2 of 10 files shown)


crates/e2e-tests/src/snapshot.rs
```diff
@@ -0,0 +1,197 @@
...
+use move_binary_format::CompiledModule;
+use move_core_types::account_address::AccountAddress;
+use sui_crypto::ed25519::Ed25519PrivateKey;
+use sui_rpc::Client;
+use sui_sdk_types::Address;
+
+/// Default snapshot directory, resolved relative to the `e2e-tests` crate:
+/// `../hashi/tests/move_upgrade_snapshots/testnet/v1`. This is the same
+/// snapshot the compat CI gate checks against.
+pub fn default_snapshot_dir() -> PathBuf {
```
dlukegordon: This hardcodes testnet/v1, so after the next deployment the test will continue exercising stale bytecode while still passing. Can we derive the snapshot version from packages/hashi/Published.toml, as the compatibility gate does.

crates/e2e-tests/Cargo.toml
```diff
@@ -26,6 +26,11 @@ sui-rpc.workspace = true
 sui-crypto.workspace = true
 sui-sdk-types.workspace = true
 sui-transaction-builder.workspace = true
+# Needed by src/snapshot.rs to deserialize/rebase/re-serialize the checked-in
+# Move bytecode snapshot before publishing it. Pinned to the same sui rev as the
+# hashi crate's move-binary-format/move-core-types dev-deps.
+move-binary-format = { git = "https://github.com/MystenLabs/sui.git", rev = "77a42de4be99383b98518541e0c6002d720cafc8" }
+move-core-types = { git = "https://github.com/MystenLabs/sui.git", rev = "77a42de4be99383b98518541e0c6002d720cafc8" }
```
dlukegordon: These Sui dependencies duplicate the revision already used by hashi, allowing the two crates to drift independently. Please define both under workspace dependencies and use .workspace = true here and in crates/hashi/Cargo.toml.

crates/e2e-tests/src/snapshot.rs
```diff
@@ -0,0 +1,197 @@
...
+    *addr = AccountAddress::ZERO;
+    Ok(())
+}
+
+/// Read the snapshot's runtime package id.
+///
+/// Every snapshot module is self-addressed at the deployed package's
+/// runtime/original id. We derive it from the first module's bytecode
+/// (matching `MovePackage::original_package_id`) rather than trusting the
+/// manifest, then assert every module agrees.
```
dlukegordon: The comment says every module is checked here, but this function only reads the first module; agreement is checked later by rebase_runtime_to_zero. Please update the comment or move the per-module check into this function.

crates/e2e-tests/tests/upgrade_tests.rs
```diff
@@ -221,3 +222,102 @@ async fn test_upgrade_v1_to_v2() -> Result<()> {
...
+    // ── Upgrade the deployed bytecode to the current source ─────────────
+    let new_package_id = upgrade_flow::execute_full_upgrade(&mut networks).await?;
+    info!("upgraded snapshot v1 -> current source: new package {new_package_id}");
+    assert_ne!(
+        new_package_id, hashi_ids.package_id,
+        "upgrade should mint a new package id"
+    );
+
+    // ── All nodes' watchers must pick up the new package version ─────────
+    info!("waiting for all nodes to detect the new package version...");
```
dlukegordon: This watcher-convergence loop duplicates the one at lines 66-93. Can we extract to a helper?

crates/e2e-tests/tests/upgrade_tests.rs
```diff
@@ -221,3 +222,102 @@ async fn test_upgrade_v1_to_v2() -> Result<()> {
...
+            for (i, node) in networks.hashi_network.nodes().iter().enumerate() {
+                let latest = node.hashi().onchain_state().package_id();
+                info!("node {i}: package_id={latest:?}");
+            }
+            anyhow::bail!("timeout: not all nodes detected the new package version");
+        }
+        tokio::time::sleep(Duration::from_millis(500)).await;
+    }
+
+    // ── v2-only canary module must be callable post-upgrade ─────────────
```
dlukegordon, line 321: The canary proves new bytecode is callable, but not that state initialized by the deployed snapshot works with the upgraded package. Please add a post-upgrade create_deposit_and_wait assertion, like the existing upgrade test, to exercise the real stateful path.

dlukegordon (approved):

PR #921, opened 2026-07-31
test(e2e): sign a withdrawal end-to-end against the deployed v1 bytecode snapshot

Bridgerz:
Stacked on #916 (which is stacked on #913). **Review/merge order: #913 → #916 → this.**

## What

Publishes the checked-in deployed testnet package **bytecode snapshot** as v1 (via `with_v1_from_snapshot`, from #916) and drives a full **deposit → withdrawal** cycle. The withdrawal forces the committee to generate presignatures and threshold-sign a Bitcoin transaction end to end.

## Why

Every other e2e harness publishes the *current* `packages/hashi` source, so the node binary always runs against a package built from the same tree. This test runs the **current binary against the actually-deployed (older) Move bytecode** — the real rollout-window configuration, where the binary is rolled out *before* the on-chain package is upgraded. #916 proves the upgrade *mechanism*; this proves the deployed-package config can still **sign**, which #916 never exercises (DKG + upgrade only, no withdrawal).

## What it does / doesn't cover

- ✅ Full presig-generation + threshold-signing lifecycle against the deployed v1 bytecode.
- ✅ Reusable withdrawal harness: extracts `extract_witness_program` / `wait_for_withdrawal_confirmation` out of `e2e_flow` into public `test_helpers`, adds `create_withdrawal_and_wait` (mirrors `create_deposit_and_wait`).
- ⏳ Forward note: once the version-gated stamped nonce-cert path re-lands (the reverted #841 work), a v1-only snapshot chain is exactly the *bare* cert path. The test carries a marked assertion to also pin `!supports_stamped_nonce_certs()` at that point.

## Test

`snapshot_v1_signs_withdrawal_end_to_end` — passes locally (deposit → withdrawal reaches `WithdrawalConfirmed`, hBTC burned), ~60s.

### DIFF
--- a/crates/e2e-tests/src/test_helpers.rs
+++ b/crates/e2e-tests/src/test_helpers.rs
@@ -16,6 +16,7 @@ use futures::StreamExt;
 use hashi::sui_tx_executor::SuiTxExecutor;
 use hashi_types::bitcoin::BitcoinAddress;
 use hashi_types::move_types::DepositConfirmed;
+use hashi_types::move_types::WithdrawalConfirmed;
 use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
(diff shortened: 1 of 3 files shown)


crates/e2e-tests/src/test_helpers.rs
```diff
@@ -198,6 +199,133 @@ pub async fn create_deposit_and_wait(
...
+        .with_signer(user_key.clone().into());
+    let withdrawal_request_id = executor
+        .execute_create_withdrawal_request(amount_sats, destination_bytes)
+        .await?;
+    info!("Withdrawal request created: {withdrawal_request_id}");
+
+    // Mine in the background so the signed BTC transaction confirms and the
+    // leader's block-driven loop reports the withdrawal on Sui.
+    let _miner = BackgroundMiner::start(&networks.bitcoin_node);
+    let confirmed = wait_for_withdrawal_confirmation(
```
dlukegordon: This waits for any withdrawal confirmation, not necessarily the request created above. Pass `withdrawal_request_id` to the waiter and only return when the event’s `request_ids` contains it.

crates/e2e-tests/src/test_helpers.rs
```diff
@@ -198,6 +199,133 @@ pub async fn create_deposit_and_wait(
...
+        .finish()]);
+    let mut subscription = sui_client
+        .subscription_client()
+        .subscribe_checkpoints(
+            SubscribeCheckpointsRequest::default().with_read_mask(subscription_read_mask),
+        )
+        .await?
+        .into_inner();
+
+    while let Some(item) = subscription.next().await {
```
dlukegordon: The timeout is only checked after the stream returns an item, so a stalled stream can hang forever. Wrap `subscription.next()` in a Tokio timeout using a fixed deadline.

crates/e2e-tests/src/test_helpers.rs
```diff
@@ -198,6 +199,133 @@ pub async fn create_deposit_and_wait(
...
+    let mut executor = SuiTxExecutor::from_config(&hashi.config, hashi.onchain_state())?
+        .with_signer(user_key.clone().into());
+    let withdrawal_request_id = executor
+        .execute_create_withdrawal_request(amount_sats, destination_bytes)
+        .await?;
+    info!("Withdrawal request created: {withdrawal_request_id}");
+
+    // Mine in the background so the signed BTC transaction confirms and the
+    // leader's block-driven loop reports the withdrawal on Sui.
+    let _miner = BackgroundMiner::start(&networks.bitcoin_node);
```
dlukegordon, line 350: Mining starts before the checkpoint subscription is created, so a fast confirmation could be missed. Create the subscription before starting the miner, then wait for the matching event.

dlukegordon (approved):

PR #923, opened 2026-07-31
feat(versioning): active-version resolution + graceful degrade

Bridgerz:
**PR 1 of the dynamic package-version management framework** (PR 2 = the decode rewire will stack on this; #841 re-lands on top). Branches off `main` — independent of the #913/#916/#921 test-infra stack.

## Why

The reverted #841 exposed that the node has no principled version gating: a single hardcoded per-feature flag (`enabled_versions.contains(&2)`, forward-broken at `{3}`/`{2,3}`) and **no notion of which package versions this binary understands** — so a binary running against a chain ahead of it would silently misparse/mis-sign rather than fail safe. For a BTC threshold signer, "halt when I don't understand the chain" beats "act on a misread."

## What

- **`SUPPORTED_PACKAGE_VERSIONS`** (`constants.rs`, currently `&[1]`) — the versions whose semantics this build implements.
- **`onchain::version`** — a pure `resolve_version_support(enabled, published, supported)` returning `Active(v)` / `Unsupported{..}` / `NotReady`. Active = `max(enabled ∩ published ∩ supported)`; an enabled-but-**unpublished** version is excluded so the ABI is never switched on before its package is live. Exhaustively unit-tested (9 cases).
- **Graceful degrade** — a combined `OnchainState::autonomous_halt_reason()` (subsumes the existing `config.paused()` check, reads one snapshot) gates `node_is_leader` and the deposit-processing halt. On `Unsupported` the node stops all leader-driven mutations, sets `hashi_package_version_unsupported=1`, and `/ready` returns 503 — signalling for a binary upgrade. `NotReady` (pre-genesis/incomplete scrape) deliberately does **not** halt.
- **Metrics** — `hashi_package_version_active` + `hashi_package_version_unsupported`.

The on-chain `assert_version_enabled` remains the ultimate backstop (a stale binary's writes abort on-chain).

## Scope note

Gates **leader-driven** mutations (the bulk). **MPC-participation gating is intentionally deferred to the #841 re-land**, which rewrites that code: today the MPC path isn't `paused()`-gated either, a stale binary's MPC writes already abort via `assert_version_enabled`, and the loud metric/readiness signal prompts the upgrade. Gating MPC participation also has committee-availability tradeoffs mid-rollout that deserve deliberate design.

## Follow-up

An e2e "chain ahead of binary" test (upgrade the chain past a test-only supported set via #916's snapshot harness) would exercise the halt end-to-end; kept out of this PR to avoid coupling to the test-infra stack. The resolver logic is fully unit-covered here.

## Test

`cargo test -p hashi --lib onchain` (37 pass, incl. 9 resolver cases); `cargo clippy --all-targets -D warnings` clean; `cargo fmt --check` clean.

### DIFF
--- a/crates/hashi/src/onchain/mod.rs
+++ b/crates/hashi/src/onchain/mod.rs
@@ -77,11 +77,22 @@ mod apply;
 mod mirror;
 mod route;
 pub mod types;
+pub mod version;
 mod watcher;
 
 fn parse_encryption_public_key(bytes: &[u8]) -> Option<crate::mpc::EncryptionGroupElement> {
     let array: [u8; 32] = bytes.try_into().ok()?;
     crate::mpc::EncryptionGroupElement::from_byte_array(&array).ok()
+}
+
+/// Why a node must not initiate autonomous on-chain mutations right now.
+#[derive(Debug, Clone, Copy)]
+pub enum HaltReason {
+    /// The bridge is governance-paused (`config.paused()`).
+    Paused,
+    /// This binary supports no live on-chain package version — typically the
+    /// chain has upgraded past this build. Fields are for diagnostics/logging.
+    BinaryUnsupported { supported_max: u64, live_max: u64 },
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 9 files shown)


crates/hashi/src/onchain/mod.rs
```diff
@@ -52,13 +52,24 @@ mod apply;
...
 
+/// Why a node must not initiate autonomous on-chain mutations right now.
+#[derive(Debug, Clone, Copy)]
+pub enum HaltReason {
+    /// The bridge is governance-paused (`config.paused()`).
+    Paused,
+    /// This binary supports no live on-chain package version — typically the
+    /// chain has upgraded past this build. Fields are for diagnostics/logging.
+    BinaryUnsupported { supported_max: u64, live_max: u64 },
+}
```
bmwill, line 96: I'm torn between hard crashing or not when we detect this. maybe we could log loudly in the leader loop when we detect this?

bmwill (commented):
Can we also add some logic in our tx executor to skip trying to execute any hashi transactions (with an error) if there is not a supported package version we know about?

We should be able to check the onchain allowlist against the const you have.

crates/hashi/src/onchain/mod.rs
bmwill: ah i see you already do this

bmwill (approved):

PR #924, opened 2026-07-31
feat(versioning): self-describing decode of version-divergent dynamic-field values

Bridgerz:
**PR 2 of the dynamic package-version framework**, stacked on #923 (PR 1).

## What

Makes the node's decode of on-chain dynamic-field values **version-aware** wherever a slot's stored type can change across package upgrades — starting with the TOB nonce-cert buckets (`EpochCertsV1` → `StampedEpochCertsV1`).

- **`onchain::versioned_decode`** — reads a dynamic field's on-chain `value_type` and identifies its layout (`TobCertLayout::{Bare, Stamped}`) instead of assuming a single Rust type. Decode follows what the chain actually reports.
- **Gated in `fetch_epoch_certs`** — adds `value_type` to the read mask and checks the layout before decoding. The bucket structs are BCS-identical, so this single gate covers every cert protocol type; a `Stamped` bucket is rejected with a clear "upgrade required" error instead of misparsing its stamped linked-table nodes downstream.

## Why value-type dispatch

Decoding by the field's actual on-chain type — rather than a global "are we on v2 yet" flag — is **self-describing and transition-safe**: during an upgrade a mix of pre- and post-flip bucket layouts is each decoded correctly, with no dependency on a flag staying in lockstep with per-bucket reality. It complements PR 1: the active-version gate halts *writes* when the chain is ahead of the binary; this makes *reads* fail loud rather than wrong.

## How the nonce-cert window work plugs in

The reverted #841 re-lands on top: its `StampedEpochCertsV1` maps to `TobCertLayout::Stamped`, and the re-land fills in the node-level stamped decode (against real data) and declares its types at `PACKAGE_VERSION = 2`. `MoveType::PACKAGE_VERSION` remains the per-type version marker used on the write side.

## Test

`cargo test -p hashi --lib onchain` (40 pass, incl. 3 new dispatch cases); `clippy --all-targets -D warnings` and `fmt --check` clean.

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -89,15 +89,13 @@ impl PackageVersions {
 ///    with the version that introduced them; existing types keep theirs
 ///    forever — never renumber (a Move type's defining address never
 ///    changes across upgrades).
-/// 3. To identify tags: use [`matches`] where a [`PackageVersions`] is in
-///    hand (event routing); use [`MODULE_NAME`] for name-keyed dispatch
-///    where the chain already reports a value's concrete type — see
-///    `hashi::onchain::versioned_decode`, and read its module doc before
-///    adding an address check there.
+/// 3. To identify tags, use [`matches`]: the full tag — defining package
+///    address included — must match the mirror, so a same-name type from a
+///    foreign package is rejected rather than trusted.
 /// 4. If the new type replaces what a dynamic-field slot holds (a v2 type
-///    in the same bucket), wire its name into the `versioned_decode`
-///    dispatch. Readers fail loudly on names they do not implement; they
-///    never guess a layout.
+///    in the same bucket), wire it into the dispatch in
+///    `hashi::onchain::versioned_decode`. Readers fail loudly on types they
+///    do not implement; they never guess a layout.
 ///
 /// The package-wide version the *tree* ships as is a separate axis: see the
 /// `PACKAGE_VERSION` doc in `packages/hashi/sources/core/versioning.move`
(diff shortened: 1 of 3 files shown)


crates/hashi-types/src/move_types/mod.rs
```diff
@@ -879,6 +913,23 @@ pub struct EpochCertsV1 {
...
+impl MoveType for EpochCertsV1 {
+    const MODULE: &'static str = "tob";
+    const NAME: &'static str = "EpochCertsV1";
+}
+
+/// Marker for the Move hashi::tob::StampedEpochCertsV1 type, which the v2
+/// package introduces (the stamped nonce-cert line of work). Identification
+/// only: this build recognizes the type but does not decode its stamped
+/// linked-table nodes; the mirror gains fields when that support lands.
+pub struct StampedEpochCertsV1;
```
bmwill, line 922: was this type already added?

crates/hashi/src/onchain/versioned_decode.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// `tob::StampedEpochCertsV1` — nodes carry a timestamp. Decoding it
+    /// requires binary support for the stamped types (the nonce dealer-cert
+    /// window work); a build without them identifies the layout but must not
+    /// attempt to decode its nodes.
+    Stamped,
+}
+
+impl TobCertLayout {
+    /// Identify the bucket layout from its on-chain value type.
+    pub fn from_struct_tag(tag: &StructTag) -> Result<Self> {
```
bmwill: i'd probably also include a PackageVersion in here to match how we handle this for MoveType. In fact you may want to just reuse the logic from MoveType trait instead of replicating this here.

crates/hashi-types/src/move_types/mod.rs
Bridgerz: Not yet — it's a forward marker for the type the v2 upgrade introduces (@zhouwfang's incoming 953). It's identification-only: no fields, no decode. With your matches() suggestion applied it's now load-bearing — PACKAGE_VERSION = 2 resolves the v2 address once published, and until then a stamped tag fails as a clean unknown-type error rather than a misdecode.

bmwill (approved):

PR #929, opened 2026-08-01
Give the nonce equivocation guard a store fallback

zhouwfang:
MystenLabs/hashi#193 MystenLabs/hashi#531

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -360,7 +360,7 @@ impl MpcManager {
             _ => None,
         };
         let existing =
-            self.get_dealer_messages(request.messages.protocol_type(), &sender, batch_index);
+            self.accepted_dealer_messages(request.messages.protocol_type(), &sender, batch_index)?;
         if let Some(existing_messages) = existing {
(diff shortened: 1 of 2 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -5823,6 +5827,49 @@ impl MpcManager {
...
+                .map_err(|e| MpcError::StorageError(e.to_string()))?
+                .map(Messages::Dkg),
+            ProtocolTypeIndicator::KeyRotation => self
+                .public_messages_store
+                .get_rotation_messages(epoch, dealer)
+                .map_err(|e| MpcError::StorageError(e.to_string()))?
+                .map(Messages::Rotation),
+            ProtocolTypeIndicator::NonceGeneration => {
+                let Some(batch_index) = batch_index else {
+                    return Ok(None);
```
benr-ml: why not an error? it should never happen

benr-ml (approved):

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Yes, it is unreachable.

PR #931, opened 2026-08-03
[Guardian] [2/3] Preserve versioned log records

mskd12:
Stacked on #904.

## Summary

- Preserve each record’s exact `LogEntry` and `VersionedLogMessage` through validation and `VerifiedLogRecord` instead of normalizing verified logs to the writer-facing schema.
- Remove the broad V1-to-V2 conversion APIs. Readers accept `LogMessageV1` and `LogMessageV2` explicitly at their consumption boundaries; only the genuinely different legacy KP-share payload keeps a narrow conversion.
- Make `LogRecord::validate` non-consuming and keep unchecked entry extraction explicit.
- Co-locate verified session and record types with the S3 reader so verified records can retain their exact versioned entries.
- Update guardian, proxy, and monitor consumers to handle versioned entries explicitly.

## Local validation

- `make fmt`

### DIFF
--- a/crates/hashi-guardian/src/s3_reader.rs
+++ b/crates/hashi-guardian/src/s3_reader.rs
@@ -15,7 +15,6 @@ use crate::s3_client::GuardianS3Client;
 use crate::s3_client::HistoryCheck;
 use crate::s3_client::LockCheck;
 use hashi_types::guardian::s3_utils::S3HourScopedDirectory;
-use hashi_types::guardian::time_utils::UnixMillis;
 use hashi_types::guardian::time_utils::UnixSeconds;
 use hashi_types::guardian::BuildPcrs;
 use hashi_types::guardian::CeremonyLogMessage;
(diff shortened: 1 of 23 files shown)


crates/hashi-guardian/src/s3_reader.rs
```diff
@@ -309,13 +264,17 @@ impl GuardianReader {
...
             .await?;
         let record = self.cache.verify_record(&self.s3, record).await?;
-        self.enforce_build_policy(build_policy, &record.build_pcrs)?;
-        let session_id = record.session_id;
-        let LogMessage::KpShareState(msg) = record.message else {
-            return Err(InvalidS3Log(format!("expected a kp-shares log at {key}")));
+        self.enforce_build_policy(build_policy, record.build_pcrs())?;
+        let session_id = record.entry().session_id().clone();
+        let msg = match record.into_entry().into_message() {
+            VersionedLogMessage::V1(LogMessageV1::KpShareState(msg)) => (*msg).try_into()?,
```
dlukegordon: If V1 KP-share conversion fails, this returns InvalidInputs, even though the bad data came from S3. Please map the conversion error to InvalidS3Log, as the previous conversion code did.

dlukegordon (approved):

crates/hashi-guardian/src/s3_reader.rs
mskd12: Fixed. V1 KP-share conversion failures are now mapped to InvalidS3Log, preserving the previous S3 error classification.

PR #915, opened 2026-07-30
[hashi-monitor][2/2] Add Sui polling for active testnet

mskd12:
## Summary

- add checkpoint-bounded Sui v2 `ListEvents` polling with exact timestamp bracketing, resumable cursors, progress logs, retries, and smaller-range fallback
- decode withdrawal approvals and timestamped `DepositApproved` events according to the security policy established in #933
- search a configurable withdrawal-predecessor lookback while auditing deposits over the complete derived Sui range
- align monitor YAML and AWS credential resolution with Guardian Init, and include the active-testnet PCRs, deployment configuration, and runbook
- read only successful Guardian withdrawal records and log progress across the hourly S3 partitions
- support hosted Bitcoin JSON-RPC headers, bounded batching, rate-limit retries, progress logs, and per-pass confirmation caching
- make continuous mode catch up immediately and make batch runs fail if source boundaries are not reached
- use one UTC RFC 3339 format for CLI timestamps and monitor fields, including readable `late_by` durations and explicit diagnostics for unfinished Guardian hourly partitions
- log E1, E2, and E3 timestamps plus signed consecutive-event durations when a withdrawal flow completes
- identify missing withdrawal and deposit events by WID or Bitcoin outpoint, while keeping expected intermediate states out of warning-level logs
- tune the active-testnet predecessor lookback and withdrawal timing bounds from observed traffic

Builds on the security-critical state-machine, finding-classification, deposit-selection, reapproval, and temporary Object Lock bypass behavior merged in #933.

## Validation

- `make fmt`
- `cargo check`
- active-testnet batch audit completed across Sui, Guardian S3, and Bitcoin Signet
- continuous catch-up retained a late Guardian event, reported its liveness finding, confirmed Bitcoin, and completed the withdrawal
- transient Sui deadline retry and range splitting observed live
- seven-day continuous active-testnet catch-up launched on a persistent remote host

PR #933, opened 2026-08-03
[hashi-monitor][1/2] Define security-critical monitor behavior

mskd12:
## Summary

- retain structurally valid late events after reporting their timing findings so later validation still runs
- return zero or more `MonitorFinding` values from ingestion, preserving every violated expectation without representing successful mutation as an operational error
- classify predecessor findings as safety and successor findings as liveness, logging safety at `ERROR` and liveness at `WARN` without terminating the monitor
- audit Sui `DepositApproved` events against their Bitcoin predecessors
- accept legitimate reapprovals for the same outpoint and retain the earliest approval for the strictest timing check
- identify deposits by their complete Bitcoin outpoint `(txid, vout)`
- add the temporary process-wide `HASHI_SKIP_S3_OBJECT_LOCK_CHECK` escape hatch for expired legacy testnet locks while retaining signature, PCR, signed-key, and S3 history validation

The predecessor classification carries a TODO to perform a focused Sui history search before reporting a missing predecessor, since a too-small configured lookback can otherwise cause a false safety finding.

The Object Lock escape hatch is explicitly tagged `TODO(testnet-wipe)` and should be removed after the planned testnet wipe.

This is the security-critical base of the monitor testnet-readiness stack. PR #915 is stacked on top with Sui and Bitcoin RPC integration and the remaining operational changes.

## Local verification

- `make fmt`


crates/hashi-monitor/README.md
```diff
@@ -11,7 +11,9 @@ Audits the cross-system bridge flow on two parallel tracks.
 
 ### Deposits (BTC → Sui)
 - **E1**: Deposit confirmed on Bitcoin.
-- **E2**: `DepositConfirmed` on Sui.
+- **E2**: `DepositApproved` on Sui. The monitor checks the committee's
```
benr-ml: just
```suggestion
- **E2**: `DepositApproved` on Sui. 
```

benr-ml (approved):

PR #937, opened 2026-08-03
Pick the key-generation cert bucket from committee history

zhouwfang:
MystenLabs/hashi#918

### DIFF
--- a/crates/hashi/src/communication/sui_tob.rs
+++ b/crates/hashi/src/communication/sui_tob.rs
@@ -159,12 +159,24 @@ pub async fn fetch_key_generation_certificates(
     onchain_state: &OnchainState,
     epoch: u64,
 ) -> Result<Vec<(Address, CertificateV1)>, TobError> {
-    let rotation =
-        fetch_certificates(onchain_state, epoch, None, ProtocolType::KeyRotation).await?;
-    if !rotation.is_empty() {
-        return Ok(rotation);
+    let earliest_committee_epoch = onchain_state.earliest_committee_epoch();
+    let protocol_type = key_generation_protocol(earliest_committee_epoch, epoch);
+    let certificates = fetch_certificates(onchain_state, epoch, None, protocol_type).await?;
+    if certificates.is_empty() {
+        tracing::warn!(
+            "epoch {epoch}: no key-generation certificates in the {protocol_type:?} bucket \
+             (earliest committee epoch {earliest_committee_epoch:?})"
+        );
     }
-    fetch_certificates(onchain_state, epoch, None, ProtocolType::Dkg).await
+    Ok(certificates)
+}
+
+fn key_generation_protocol(earliest_committee_epoch: Option<u64>, epoch: u64) -> ProtocolType {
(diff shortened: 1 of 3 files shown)


crates/hashi/src/communication/sui_tob.rs
```diff
@@ -159,12 +159,24 @@ pub async fn fetch_key_generation_certificates(
...
 ) -> Result<Vec<(Address, CertificateV1)>, TobError> {
-    let rotation =
-        fetch_certificates(onchain_state, epoch, None, ProtocolType::KeyRotation).await?;
-    if !rotation.is_empty() {
-        return Ok(rotation);
+    let earliest_committee_epoch = onchain_state.earliest_committee_epoch();
+    let protocol_type = key_generation_protocol(earliest_committee_epoch, epoch);
+    let certificates = fetch_certificates(onchain_state, epoch, None, protocol_type).await?;
+    if certificates.is_empty() {
+        tracing::warn!(
```
benr-ml, line 166: why warn? strange that it warns but returns OK

crates/hashi/src/onchain/mod.rs
```diff
@@ -385,6 +385,27 @@ impl OnchainState {
         self.state().hashi.committees.epoch()
     }
 
+    pub(crate) fn earliest_committee_epoch(&self) -> Option<u64> {
+        self.state()
+            .hashi
+            .committees
+            .committees()
```
benr-ml, line 392: so this stores **all** committees? how would we sync new nodes?

crates/hashi/src/communication/sui_tob.rs
zhouwfang: Fair, it probably should be an error. The trouble is, absent and empty come from the same listing, so neither is more reliable, and a lagging one can return a truncated prefix that neither would catch. I'd rather fix that properly than two thirds of it.

crates/hashi/src/onchain/mod.rs
zhouwfang: Yeah, all of them, and a new node pages the lot at boot. It matters here because I derive DKG-vs-rotation from the map's minimum, so the oldest committee has to stay forever. Prune it and that one starts reading as DKG. Committee.config is probably the better home, but it's a Move change so a separate PR.

benr-ml: it sounds unnecessarily expensive - why do we care about more than 2-3 committees back? for the current epoch I only care about the current committee and the last one

zhouwfang: Right, only this read needs the full history: the rest stops at three, and it degrades silently rather than erroring if you go tighter. The fix is to stamp the bit per epoch instead of deriving it, then cap. Separate PR.

benr-ml (approved):

crates/hashi/src/communication/sui_tob.rs
zhouwfang: Followup: #961

PR #940, opened 2026-08-03
Collect dealer signatures until the quorum, not until a deadline

zhouwfang:
MystenLabs/hashi#918

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -87,6 +87,7 @@ use hashi_types::committee::Bls12381PrivateKey;
 use hashi_types::committee::BlsSignatureAggregator;
 use hashi_types::committee::Committee;
 use hashi_types::committee::MemberSignature;
+use hashi_types::committee::ReducedWeight;
 use rand::seq::SliceRandom;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
@@ -109,6 +110,9 @@ const HEDGED_RETRIEVE_INITIAL_ROUND_SIZE: usize = 2;
 const HEDGED_RETRIEVE_ROUND_GROWTH_FACTOR: usize = 2;
 const HEDGED_RETRIEVE_ROUND_TIMEOUT: Duration = Duration::from_secs(1);
 const PREVIOUS_MESSAGE_REPAIR_ATTEMPT_TIMEOUT: Duration = Duration::from_secs(20);
+/// How long a dealer keeps folding straggler signatures after its quorum is
+/// already met.
+const DEALER_STRAGGLER_GRACE: Duration = Duration::from_secs(10);
 
 type AvidEchoAndVote = (
     BLS12381Signature,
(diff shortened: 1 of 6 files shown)


benr-ml (commented):
I don't understand how it solves the issue - we send messages to all parties and want to make sure that all the slow ones will eventually receive them - why not continue retrying until we complete the protocol? e.g., for avss/dkg, the dealer should send the message and retry until it has t+f votes. it should not fail after 1 min but continue retrying, no?

zhouwfang:
@benr-ml Deadline is gone. testnet says it was cutting nearly every round.

Now it collects until t+f, then waits a bit past that before publishing. On "Retry for longer, but afterwards wait just a bit more", I'd done the first half and skipped the second, and it matters: a bare t+f cert leaves complaint recovery no slack. Did both AVID legs too.

Still guessing how long that extra wait should be, so added metrics to pick it off real data.

One AVID thing: optimistic only publishes if everyone confirms, but we now stop at W−f plus a 35s grace instead of waiting for all of them. A slow last node drops us to the pessimistic leg: extra crypto round plus another fan-out. Better than blocking for minutes I think, but the grace decides how often, and I've only run AVID once on devnet so nothing to size it from yet.

benr-ml:
> @benr-ml Deadline is gone. testnet says it was cutting nearly every round.
> 
> Now it collects until t+f, then waits a bit past that before publishing. On "Retry for longer, but afterwards wait just a bit more", I'd done the first half and skipped the second, and it matters: a bare t+f cert leaves complaint recovery no slack. Did both AVID legs too.
> 
> Still guessing how long that extra wait should be, so added metrics to pick it off real data.
> 
> One AVID thing: optimistic only publishes if everyone confirms, but we now stop at W−f plus a 35s grace instead of waiting for all of them. A slow last node drops us to the pessimistic leg: extra crypto round plus another fan-out. Better than blocking for minutes I think, but the grace decides how often, and I've only run AVID once on devnet so nothing to size it from yet.

I don't follow - why does it need to wait after t+f for dkg/key-rotation? the only steps in which we should wait a bit longer are: first phase of the batch avid protocol (to reduce the overhead of the second phase), and when collecting the presig certs (to increase the number of nonce we generate per batch).

also, wdym by
> a bare t+f cert leaves complaint recovery no slack

zhouwfang:
@benr-ml Both fair. Overstated the slack thing: t+f already tolerates f, nothing's missing. Dropped the wait. Had Confirm backwards too. Now it goes for unanimity with a 90s window, still arming at W−f so one dead node doesn't cost us the full retry budget. PTAL.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2884,25 +2918,28 @@ impl MpcManager {
...
+                mgr.mpc_config.threshold as u32 + mgr.mpc_config.max_faulty as u32,
+                mgr.address,
+            )
+        };
+        let decided_weight = dealer_data.vote_quorum_weight.max(min_confirm_weight);
+        collect_dealer_signatures(
+            &mut aggregator,
+            requests,
+            decided_weight,
+            AVID_CONFIRM_GRACE,
```
benr-ml: i am assuming this is the first phase (collecting https://github.com/MystenLabs/fastcrypto/blob/main/fastcrypto-tbls/src/threshold_schnorr/batch_avss_avid.rs#L114-L116). if so maybe call it BATCH_AVSS_VOTES_GRACE

benr-ml (approved):

PR #949, opened 2026-08-05
[Monitor] [1/2] Add Sui and Bitcoin polling

mskd12:
This is the chain-source half of the monitor stack.

This PR:
- scans only transactions emitting `WithdrawalPickedForProcessing` or `DepositConfirmed` through Sui v2 `ListTransactions`
- obtains the timestamp for `DepositConfirmed` from its containing checkpoint and restores rejection of conflicting duplicate confirmed deposits
- resolves timestamp boundaries to checkpoints and resumes through server scan/page limits
- replaces the legacy Bitcoin Core client with provider-compatible JSON-RPC batching, caching, pacing, and rate-limit retries
- requires six Bitcoin confirmations

Monitor stack:
- [1/2]: this PR
- [2/2]: #950

Local verification:
- `make fmt`
- `cargo check`
- live 10-minute testnet batch audit via the complete stack in #950


crates/hashi-monitor/src/rpc/btc.rs
```diff
@@ -1,31 +1,110 @@
...
+    confirmations: u64,
+}
+
+#[derive(Deserialize)]
+struct RawBlockHeader {
+    time: u64,
+}
+
+impl RawTransactionInfo {
+    fn sufficiently_confirmed_block_hash(self, txid: Txid) -> anyhow::Result<Option<BlockHash>> {
```
benr-ml, line 65: why result<option<>>? does it ever return error?

benr-ml (approved):
overall looks good to me, but @bmwill or others better review the sui and btc integrations at some point

crates/hashi-monitor/src/rpc/btc.rs
mskd12: So it currently returns:

- Ok(None): successful lookup, but the transaction is not yet sufficiently confirmed (unmined or mined with 1-5 confirmations) 
- Ok(time): successful lookup with 6 confirmations
- Err: all other RPC errors

The reason we distinguish the first case is to record the time at which we checked Bitcoin and use that to throw MissingEvent if needed.

mskd12: It can return an error: `Ok(None)` represents a successful lookup where the transaction is unmined or has fewer than six confirmations, while parsing an invalid block-hash string returns `Err`. The outer lookup also propagates transport and malformed-response errors. I added a TODO beside the retry loop to classify and retry transient non-rate-limit errors separately.

PR #951, opened 2026-08-05
[Guardian] Validate log records during deserialization

mskd12:
## Summary

- use one owned wire representation for log-record serialization and deserialization
- share `LogEntry` construction between writers and readers, validating canonical routing and attestation-session binding at construction
- reject invalid signed/unsigned envelope combinations during deserialization while keeping public-key-dependent authentication in `LogRecord::validate`

## Local checks

- `make fmt`
- `cargo check`

### DIFF
--- a/crates/hashi-types/src/guardian/log/record.rs
+++ b/crates/hashi-types/src/guardian/log/record.rs
@@ -1,12 +1,14 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//! S3 log-record wire format and record-local validation.
+//! S3 log-record wire format and validation.
 //!
 //! A [`LogRecord`] carries a [`VersionedLogMessage`] and its S3 routing
 //! context. Signed records carry a Guardian signature over their [`LogEntry`].
 //! The one permitted unsigned entry carries the OI attestation that establishes
-//! the Guardian signing key and must be authenticated separately.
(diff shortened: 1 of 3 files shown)


crates/hashi-types/src/guardian/log/record.rs
```diff
@@ -61,47 +64,35 @@ pub enum LogRecord {
...
-                message,
-                signature: &signature,
-            }
-            .serialize(serializer),
+        LogRecordWire {
+            schema_version: data.schema_version,
+            object_key: data.object_key.clone(),
+            session_id: data.session_id.clone(),
+            timestamp_ms: data.timestamp_ms,
+            message: serde_json::to_value(&data.message).map_err(S::Error::custom)?,
```
dlukegordon, line 92: Serializing the message through serde_json::to_value sorts nested object keys because preserve_order is disabled, breaking deployed JSON byte stability and potentially causing S3’s byte-for-byte retry check to panic. Can serialize it directly as before?

mskd12: Hmm, we can do that. But your concern is theoretical, no? Concretely, what can go wrong with this?

dlukegordon: I was originally thinking this was non deterministic serialization. But it actually is deterministic. So you can disregard.

dlukegordon (approved):

PR #952, opened 2026-08-05
[guardian-proxy] Push metrics to Mimir from in-process

0xsiddharthks:
## Summary

Nothing scrapes the guardian proxy: it runs on Fargate and the scrapers live in GKE, so `guardian_proxy_*` never reached Mimir. The proxy now pushes its own registry over Prometheus remote write. A collector sidecar was the alternative, but every container in an ECS task shares the task role, and this one can read the guardian's withdrawal log.

Verified on devnet (sui-operations run 33692347929): `guardian_proxy_widlog_*` lands in Mimir with the full `le` ladder. Pairs with MystenLabs/sui-operations#8549, which sets the env vars.

## Changes

- `remote_write.rs`: gather the registry every `MIMIR_PUSH_INTERVAL_SECS` (default 60, max 300) and POST it as prompb over snappy with basic auth. Off unless `MIMIR_URL` is set.
- Retry the same body with backoff (1s to 16s) on 5xx, 429 and network errors; other 4xx and 3xx are not retried. Every attempt is cut off at the next scheduled tick, so a stalled push never outlives the snapshot after it.
- Label merge follows Prometheus: a metric's own label wins over an external label of the same name.
- No redirects, which would turn the POST into a bodyless GET.
- Validate the config at startup: http(s) URL, non-empty credentials, interval in 1..=300, and `MIMIR_EXTERNAL_LABELS` with no repeated or encoder-reserved names (`__*`, `le`).
- Hand-rolled prompb: `prometheus-reqwest-remote-write` emits histogram buckets under the bare metric name, which breaks `histogram_quantile`.


crates/hashi-guardian-proxy/src/remote_write.rs
```diff
@@ -0,0 +1,321 @@
...
+        .context("remote-write client build")?;
+    let authorization = basic_auth(&config.username, &config.password)?;
+
+    tokio::spawn(async move {
+        info!(url = %config.url, interval =? config.interval, "Started the metrics push task.");
+        let mut ticker = tokio::time::interval(config.interval);
+        ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        loop {
+            ticker.tick().await;
+            if let Err(e) = push_once(&client, &config, &authorization, &registry).await {
```
dlukegordon, line 52: If this request gets a 5xx, we drop it and send a new snapshot on the next tick. If the proxy restarts before a successful push, counter changes can be lost.

Can we retry the same request with exponential backoff for network errors and 5xx responses? Normal 4xx responses should not be retried.

crates/hashi-guardian-proxy/src/config.rs
```diff
@@ -84,10 +105,40 @@ impl Config {
...
+    raw.split(',')
+        .map(str::trim)
+        .filter(|entry| !entry.is_empty())
+        .map(|entry| {
+            let (name, value) = entry.split_once('=').with_context(|| {
+                format!("MIMIR_EXTERNAL_LABELS entry {entry:?} must be name=value")
+            })?;
+            let (name, value) = (name.trim(), value.trim());
+            anyhow::ensure!(
+                is_label_name(name) && !value.is_empty(),
```
julio-ml: parse_external_labels validates each entry independently but does not reject duplicate keys or names added later by the encoder. Values such as `network=a,network=b` or `__name__=foo` produce repeated labels; `le=foo` does the same for histogram series. Remote Write forbids repeated label names, so Mimir skips the invalid series and returns 400. Please reject duplicate keys and collisions with generated labels at startup.

crates/hashi-guardian-proxy/src/config.rs
```diff
@@ -73,6 +80,20 @@ impl Config {
...
             parse_kp_roster(&std::env::var("AUTHORIZED_KP_FINGERPRINTS").unwrap_or_default())?;
+        let remote_write = match std::env::var("MIMIR_URL").ok().filter(|u| !u.is_empty()) {
+            None => None,
+            Some(url) => Some(RemoteWriteConfig {
+                url,
+                username: std::env::var("MIMIR_USERNAME")
+                    .unwrap_or_else(|_| "incoming_metrics".to_string()),
+                password: std::env::var("MIMIR_PASSWORD")
+                    .context("MIMIR_PASSWORD must be set when MIMIR_URL is")?,
+                interval: Duration::from_secs(parse_env_u64("MIMIR_PUSH_INTERVAL_SECS", 60)?),
```
julio-ml: `MIMIR_PUSH_INTERVAL_SECS=0` is accepted here, but `tokio::time::interval(Duration::ZERO)` panics. Because this binary installs an aborting panic hook, that configuration terminates the entire proxy despite metrics being intended not to take it down. Please require a strictly positive interval during config parsing.

crates/hashi-guardian-proxy/src/remote_write.rs
0xsiddharthks: Done in ee177cc6a: a 5xx, 429 or network error now resends the same body with exponential backoff (1s→16s) until the next push is due, then the fresh snapshot takes over; any other 4xx is not retried.

crates/hashi-guardian-proxy/src/config.rs
0xsiddharthks: Done in ee177cc6a: `parse_external_labels` (moved next to the encoder in `remote_write.rs`) now rejects repeated names, `__`-prefixed names and `le` at startup.

crates/hashi-guardian-proxy/src/config.rs
0xsiddharthks: Done in ee177cc6a: `MIMIR_PUSH_INTERVAL_SECS` must be ≥ 1, enforced at startup.

dlukegordon (commented):
Additional review findings.

crates/hashi-guardian-proxy/src/remote_write.rs
```diff
@@ -0,0 +1,525 @@
...
+) -> Vec<TimeSeries> {
+    let mut out = Vec::new();
+    for family in families {
+        let name = family.name();
+        for metric in family.get_metric() {
+            // External labels win; a series may not carry a label name twice.
+            let mut common: Vec<Label> = metric
+                .get_label()
+                .iter()
+                .filter(|l| !external_labels.iter().any(|(k, _)| k == l.name()))
```
dlukegordon: External labels currently replace metric labels. If `outcome` is configured externally, every existing request-counter child collapses to the same label set and timestamp, so Mimir rejects conflicting duplicate-timestamp samples. Please follow Prometheus's merge behavior and keep metric labels when both sources define the same name, or reject the collision.

crates/hashi-guardian-proxy/src/remote_write.rs
```diff
@@ -0,0 +1,525 @@
...
+        .context("remote-write client build")?;
+    let authorization = basic_auth(&config.username, &config.password)?;
+
+    tokio::spawn(async move {
+        info!(url = %config.url, interval =? config.interval, "Started the metrics push task.");
+        let mut ticker = tokio::time::interval(config.interval);
+        ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        loop {
+            ticker.tick().await;
+            let deadline = Instant::now() + config.interval;
```
dlukegordon: This deadline can extend past the ticker's actual next scheduled tick after a late `Skip` tick, and an in-flight request is not bounded by it. A stalled send can therefore keep the old snapshot alive after a fresh push is due. Please derive the deadline from the instant returned by `ticker.tick()` and wrap each attempt with `timeout_at(deadline, ...)`.

crates/hashi-guardian-proxy/src/remote_write.rs
```diff
@@ -0,0 +1,525 @@
...
+    pub username: String,
+    pub password: String,
+    pub interval: Duration,
+    pub external_labels: Vec<(String, String)>,
+}
+
+/// Spawn the push task. A failing push is retried until the next one is due,
+/// then dropped for the fresh snapshot; metrics never take the proxy down.
+pub fn start(config: RemoteWriteConfig, registry: Registry) -> Result<()> {
+    let client = reqwest::Client::builder()
```
dlukegordon: Reqwest follows redirects by default, and 301/302/303 convert this POST into a bodyless GET. If that GET returns 2xx, the snapshot is silently discarded as successful. Please disable redirects for the Remote Write client, or only follow redirects that preserve the POST method and body.

crates/hashi-guardian-proxy/src/config.rs
```diff
@@ -75,6 +83,29 @@ impl Config {
...
+            None => None,
+            Some(url) => {
+                let interval_secs = parse_env_u64("MIMIR_PUSH_INTERVAL_SECS", 60)?;
+                // `tokio::time::interval` panics on zero, and a panic aborts this binary.
+                anyhow::ensure!(
+                    interval_secs > 0,
+                    "MIMIR_PUSH_INTERVAL_SECS must be at least 1"
+                );
+                Some(RemoteWriteConfig {
+                    url: url.parse().context("MIMIR_URL must be a valid URL")?,
```
dlukegordon: `Url::parse` accepts non-HTTP schemes, but reqwest rejects them only during send, where the error is treated as transient forever. Please require `http` or `https` while parsing `MIMIR_URL` so a typo fails startup.

crates/hashi-guardian-proxy/src/remote_write.rs
0xsiddharthks: Fixed in 5dcff8498, metric labels win now like in Prometheus. Added a test with two `outcome` children to make sure they stay separate series.

crates/hashi-guardian-proxy/src/remote_write.rs
0xsiddharthks: Good catch. The deadline now comes from the instant `tick()` returns, and each attempt is wrapped in `timeout_at(deadline)`. 5dcff8498.

crates/hashi-guardian-proxy/src/remote_write.rs
0xsiddharthks: Disabled redirects on the client, so a 3xx now fails the push as permanent. Added 301/303 to the classification test. 5dcff8498.

crates/hashi-guardian-proxy/src/config.rs
0xsiddharthks: Now requires http/https at parse time, and reqwest builder errors are treated as permanent. Also capped the interval at 300s, a huge value overflowed the deadline math and would have aborted the proxy. 5dcff8498.

dlukegordon (approved):
All review comments have been addressed.

PR #953, opened 2026-08-05
Accumulate nonce dealer certs within a stamped time window past the `W−f` floor

zhouwfang:
Re-merging #841


packages/hashi/sources/core/tob.move
```diff
@@ -52,6 +45,18 @@ public struct DealerSubmissionV1 has copy, drop, store {
     signature: CommitteeSignature,
 }
 
+public struct StampedDealerSubmissionV1 has copy, drop, store {
```
Bridgerz, line 48: Is there a reason we have these structs versioned already? We only add struct versions when we need to upgrade an existing struct.

packages/hashi/sources/core/tob.move
```diff
@@ -52,6 +45,18 @@ public struct DealerSubmissionV1 has copy, drop, store {
...
+    submission: DealerSubmissionV1,
+    timestamp_ms: u64,
+}
+
+public struct StampedEpochCertsV1 has store {
```
Bridgerz, line 53: same as above

packages/hashi/sources/hashi.move
```diff
@@ -223,6 +223,18 @@ public(package) fun epoch_certs(
...
+public(package) fun epoch_certs_stamped(
+    self: &mut Hashi,
+    key: hashi::tob::TobKey,
+    ctx: &mut TxContext,
+): &mut hashi::tob::StampedEpochCertsV1 {
+    let epoch = key.epoch();
+    if (!self.tob.contains(key)) {
+        self.tob.add(key, hashi::tob::create_stamped(epoch, key.protocol_type(), ctx));
+    };
+    self.tob.borrow_mut(key)
```
Bridgerz, line 235: This borrow_mut doesn't check the existing entry's concrete type. commit_upgrade auto-enables the new version in the same tx that publishes, so any batch mid-accumulation at upgrade time has a bare bucket, and every post-upgrade submit_nonce_cert for it will abort here. If it hadn't reached the W-f floor yet it never will, and presig refill pins on that index until reconfig. Same race the other way while any pre-upgrade binary is still submitting: first submitter fixes the bucket layout and the other side aborts for the rest of the batch.

Can we make the v2 submit layout tolerant? Append matching the existing bucket type, and only create stamped buckets fresh. That fixes both directions and we wouldn't need to time the upgrade around in-flight batches.

packages/hashi/sources/core/cert_submission.move
```diff
@@ -63,10 +59,16 @@ entry fun destroy_all_certs(
     protocol_type: ProtocolType,
 ) {
     hashi.versioning().assert_version_enabled();
+    let is_nonce_generation = protocol_type.is_nonce_generation();
     let key = hashi::tob::tob_key(epoch, batch_index, protocol_type);
     let current_epoch = hashi.committee_set().epoch();
-    let epoch_certs: hashi::tob::EpochCertsV1 = hashi.tob_mut().remove(key);
-    hashi::tob::destroy_all(epoch_certs, current_epoch);
+    if (is_nonce_generation) {
+        let epoch_certs: hashi::tob::StampedEpochCertsV1 = hashi.tob_mut().remove(key);
```
Bridgerz, line 66: This picks the remove type from protocol_type instead of what's actually stored, so it aborts on every nonce bucket created before the upgrade (those are EpochCertsV1 and stay that way). As long as v1 stays enabled this is just a GC footgun: bare buckets need the v1 entry, stamped ones need this one, and since nothing in Rust calls destroy_all_certs, whoever runs GC manually has to know which package id to use per bucket. But if we ever disable v1, any bare buckets still around become undestroyable without a v3. Probing the bucket's actual type and routing to destroy_all / destroy_all_stamped fixes both, and it's the same pattern fix as the submit path.

packages/hashi/sources/core/mpc_config.move
```diff
@@ -41,6 +41,8 @@ public(package) fun is_valid_value(key: &std::string::String, value: &config_val
         value.is_u64() && (*value).as_u64() <= MAX_BPS
     } else if (k == &KEY_NONCE_GENERATION_PROTOCOL) {
         value.is_u64() && (*value).as_u64() <= 1
+    } else if (k == &KEY_NONCE_ACCUMULATION_WINDOW_MS) {
```
Bridgerz, line 44: On the real deployment path (v1 chain upgraded to v2) this arm is dead code. update_config rejects keys that aren't already in the stored config, and nothing inserts nonce_accumulation_window_ms at upgrade time, so governance can never set or tune the window on the upgraded chain. It's pinned to the compiled default forever. Same root cause as the e2e override warn/no-op, the window e2e only passes because it asserts the default. We need insert-if-absent for known keys in update_config, or a migration step that seeds the key at upgrade.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1084,38 +1149,62 @@ impl MpcManager {
...
         &self,
         certs: &VerifiedNonceCerts<T>,
-    ) -> (HashSet<Address>, u32) {
-        let required_weight = self.required_nonce_weight();
-        let mut weight_sum = 0u32;
+    ) -> (HashSet<Address>, NonceCollectionWindow) {
+        self.certified_nonce_dealers_in_window(certs, self.nonce_collection_window())
+    }
+
+    fn certified_nonce_dealers_in_window<T: NonceCertTimestamp>(
```
Bridgerz, line 1159: Weight threshold mismatch here: this walk admits any signature-verified committee cert, but in AVID mode verified_nonce_certs verifies at avid_vote_quorum (W-f) while run_as_avid_nonce_party requires total_reduced_weight for AvssVote certs. So a confirm cert with weight in [W-f, W) counts toward the floor/cutoff here but gets rejected by the party loop. One byzantine dealer landing a thin cert under its own slot wedges the batch for the rest of the epoch, which is inside the fault model AVID is supposed to tolerate. These two paths should agree on the admission threshold.

crates/hashi/src/mpc/service.rs
```diff
@@ -1039,6 +1203,176 @@ impl MpcService {
...
+                if tokio::time::Instant::now() >= wait_deadline {
+                    metrics.mpc_nonce_fetch_floor_unreached_total.inc();
+                    anyhow::bail!(
+                        "nonce certs never reached the floor for epoch {epoch} batch {batch_index}"
+                    );
+                }
+                tokio::time::sleep(NONCE_WINDOW_WAIT_POLL).await;
+                continue;
+            }
+            let read_side_past_cutoff = |cutoff_ms: u64| async move {
```
Bridgerz, line 1312: The determinism of the admitted dealer set assumes the cert table fetches and this read-side clock come from one monotone read view. Nothing anchors the chosen set on-chain and there's no cross-node check, so a node reading through load-balanced fullnodes with unequal lag can confirm the cutoff against a stale table prefix. Linked table reads are insertion-order prefixes, so any stale snapshot past the floor yields the same cutoff with tail certs silently missing. At minimum we should document and enforce one monotone RPC endpoint per node. The robust fix is anchoring the selected set on-chain.

crates/hashi/src/sui_tx_executor.rs
```diff
@@ -1782,6 +1823,9 @@ impl SuiTxExecutor {
     }
 }
 
+// TODO: the free builders still call the original package id, so they run v1
```
Bridgerz, line 1826: This TODO undersells one case: build_register_or_update_validator_tx is node-autonomous (startup registration and reconfig key updates), hardcodes the original package for all six validator calls, and its execute path also skips execute()'s must_halt refusal. So after we disable v1, registration becomes the one autonomous write still submitting doomed v1 calls instead of being routed or refused. The halt bypass predates this PR, but since everything else routes through call_target() now, lets just close both here instead of leaving the TODO.

Bridgerz (changes requested):

packages/hashi/sources/core/tob.move
zhouwfang: tob.move is already all V1 and none of them version anything, so these two would be the odd ones out. The name's also what versioned_decode dispatches on.

packages/hashi/sources/core/tob.move
zhouwfang: replied above

crates/hashi/src/mpc/service.rs
zhouwfang: Mostly done already, 62742c1e0 put the clock read on the same client as the certs. Residual needs an L7 proxy balancing per RPC.

crates/hashi/src/sui_tx_executor.rs
zhouwfang: Will route the ids. Disabling v1 doesn't halt us though, v2 is still live and supported. So it's just the hardcoded id that breaks registration. Leaving the refusal out for now, that path publishes next-epoch keys and refusing is how a node fails to join the next committee.

packages/hashi/sources/core/tob.move
Bridgerz: gotcha. well for future reference, we don't need to add V1 to all initial structs. This module is now the odd one out. Not a big deal tho so cool to leave it.

packages/hashi/sources/hashi.move
zhouwfang: That only covers one direction though. If v2 creates a stamped bucket first, a v1 node aborts, and we can't fix that from v2 since v1's bytecode is frozen. So we'd need to keep creating bare buckets until v1 is disabled, then switch.

packages/hashi/sources/core/cert_submission.move
zhouwfang: Same fix. Only a storage leak today since nothing calls destroy_all_certs from Rust, but it goes undestroyable once v1 is off.

packages/hashi/sources/core/mpc_config.move
zhouwfang: It hits any key a future upgrade adds, not just this one. Leaning toward seeding at upgrade over loosening update_config, since its no-unknown-keys invariant is pinned by a test.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: This had been on my list, and there's a note on the test. The two paths use different rules on purpose, and making them match causes a worse problem than it fixes, so I left it.

crates/hashi/src/sui_tx_executor.rs
Bridgerz: Routing sgtm — please use active_call_package_id from #1026 (just landed)  rather than a new helper

crates/hashi/src/mpc/mpc_except_signing.rs
Bridgerz: Sg

packages/hashi/sources/core/cert_submission.move
Bridgerz: might be worth to check out #1001 and #1002 (stacking shortly).

Bridgerz (approved):
LGTM

PR #955, opened 2026-08-05
[Guardian] Preserve log record serialization order

mskd12:
Follow-up to #951.

Serialize concrete V1/V2 log messages directly instead of first converting them to `serde_json::Value`. This keeps deployed JSON bytes stable regardless of whether `serde_json/preserve_order` is enabled by the surrounding dependency graph.

Deserialization continues to use `Value` for schema-version dispatch.

Local checks:
- `make fmt`
- `cargo check`
- `cargo test -p hashi-types deployed_v1_kp_share_state_verifies_and_is_retained`

### DIFF
--- a/crates/hashi-types/src/guardian/log/record.rs
+++ b/crates/hashi-types/src/guardian/log/record.rs
@@ -28,7 +28,6 @@ use crate::guardian::now_timestamp_ms;
 use serde::Deserialize;
 use serde::Serialize;
 use serde::de::Error as _;
-use serde::ser::Error as _;
 use serde_json::Value;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


crates/hashi-types/src/guardian/log/record.rs
```diff
@@ -84,18 +94,39 @@ impl Serialize for LogRecord {
...
         };
-        LogRecordWire {
-            schema_version: data.schema_version,
-            object_key: data.object_key.clone(),
-            session_id: data.session_id.clone(),
-            timestamp_ms: data.timestamp_ms,
-            message: serde_json::to_value(&data.message).map_err(S::Error::custom)?,
-            signature,
+
+        match &data.message {
```
dlukegordon, line 98: This duplicates the V1/V2 dispatch already handled by VersionedLogMessage::serialize. Could we pass &data.message directly to LogRecordWireRef and remove the match and serialize_log_record helper?

dlukegordon (approved):

crates/hashi-types/src/guardian/log/record.rs
mskd12: Done in #957 — LogRecordWireRef now stores &VersionedLogMessage directly, and the duplicate match and helper are removed.

PR #959, opened 2026-08-05
[limiter] adopt the guardian's limiter policy on rotation

0xsiddharthks:
## Summary

As of now, the local limiter in the hashi node, is only reconciling the limiter state (nonce + current limit), however changes to config are not reconciled.

With this, we area also syncing the limiter config to the local limiter config.

## Changes

- `guardian_limiter.rs`: config moves under the same lock as state as a `LimiterView`; adds `adopt_config`, and `view()` so callers needing both read them together
- `lib.rs`: the reconcile loop pulls `limiter_config` off `GetGuardianInfo` and adopts it before comparing state
- `withdrawal_request_flow.rs`, `mirror.rs`: one `view()` read in place of paired accessors
- `metrics.rs`: adds `hashi_guardian_limiter_config_changed_total`. the existing capacity/refill gauges now follow a rotation instead of reporting boot-time values forever

### DIFF
--- a/crates/hashi/src/lib.rs
+++ b/crates/hashi/src/lib.rs
@@ -1000,9 +1000,20 @@ impl Hashi {
         if !self.verify_and_pin_guardian_btc_pubkey(info.enclave_btc_pubkey) {
             return None;
         }
-        let config = info.limiter_config?;
-        let state = info.limiter_state?;
-        Some((limiter, config, state))
+        match (info.limiter_config, info.limiter_state) {
+            (Some(config), Some(state)) => Some((limiter, config, state)),
+            // The enclave installs the config at operator_init and builds the
+            // limiter from it at operator_activate, so state can never outrun
+            // config. Say so rather than stalling every reconcile in silence.
+            (None, Some(_)) => {
+                tracing::warn!(
+                    "Guardian reported limiter state without a config; skipping reconcile"
+                );
+                None
+            }
+            // Provisioned but not yet activated — ordinary mid-rotation.
+            _ => None,
+        }
     }
 
     /// Adopt the guardian's limiter policy when a re-provision changed it. The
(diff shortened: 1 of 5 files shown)


crates/hashi/src/lib.rs
```diff
@@ -985,21 +985,46 @@ impl Hashi {
...
         Arc<guardian_limiter::LocalLimiter>,
+        hashi_types::guardian::LimiterConfig,
         hashi_types::guardian::LimiterState,
     )> {
         let limiter = self.local_limiter()?;
         let info = self.fetch_guardian_info_data().await.ok()?;
         if !self.verify_and_pin_guardian_btc_pubkey(info.enclave_btc_pubkey) {
             return None;
         }
+        let config = info.limiter_config?;
```
Bridgerz: NIT: If the guardian's response has limiter_state but no limiter_config (older guardian binary — the proto field is optional), this now silently skips all limiter reconciliation, where before the state-only reconcile worked. A current guardian always sends both since readiness gates on the config being set, so this only bites version skew — but could we get a debug!/warn! on this None path so a skewed rollout shows up in logs instead of just quietly stalling reconciles?

Bridgerz (approved):

crates/hashi/src/lib.rs
0xsiddharthks: checked this one, the skew window doesn't actually exist. `limiter_state` and `limiter_config` both went into the proto in the same commit (#484), and the enclave builds the `RateLimiter` from the already installed config at `operator_activate`, so state can't show up without config.

still a new silent failure path though, so worth fixing. added a warn but only for the `state && !config` case. config without state is just a standby that hasn't activated yet, and this runs every 15s, so warning on that would spam the logs through every rotation.

matches what the proxy already does in `info.rs`, limiter is null unless both halves are there.

PR #961, opened 2026-08-05
Stop collapsing absent and truncated cert listings into an empty bucket

zhouwfang:
Followup on MystenLabs/hashi#937

MystenLabs/hashi#918

### DIFF
--- a/crates/hashi/src/onchain/mod.rs
+++ b/crates/hashi/src/onchain/mod.rs
@@ -48,6 +48,21 @@ const BROADCAST_CHANNEL_CAPACITY: usize = 100;
 /// the gRPC decode limit; the SDK still pages through every entry.
 const SCRAPE_PAGE_SIZE: u32 = 1000;
 
+const CERT_LISTING_ATTEMPTS: u32 = 3;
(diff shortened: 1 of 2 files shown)


crates/hashi/src/onchain/mod.rs
```diff
@@ -878,18 +878,35 @@ impl OnchainState {
...
-        let mut current = Some(head);
-        while let Some(dealer) = current {
-            let Some(node) = nodes.remove(&dealer) else {
-                break;
-            };
-            certificates.push((dealer, node.value));
-            current = node.next;
-        }
-        Ok(Some(certificates))
+        walk_linked_table(head, epoch_certs.certs.size, nodes)
```
dlukegordon, line 922: These reads are not pinned to one checkpoint, so a certificate appended while the dynamic fields are being paginated can make this validation fail even though the on-chain table is valid. Please retry the whole bucket read when `walk_linked_table` detects an inconsistent listing, or read the parent and all pages from one checkpoint.

zhouwfang: Re-reads the whole bucket up to 3 times on an inconsistency, RPC errors still propagate straight away.

dlukegordon (approved):

PR #964, opened 2026-08-05
[btc-monitor] upgrade kyoto to 0.6.3 and replenish its peer whitelist

0xsiddharthks:
Bumps `bip157` 0.6.0 → 0.6.3 and stops the kyoto node from dying when its peer whitelist runs dry.

Upgrade changes:

- **0.6.1** — kyoto sends `getheaders` on `inv` block announcements instead of dropping them. Core falls back to `inv` on >8-block bursts and reorgs, so we went deaf until the 30-minute stale-tip timer disconnected every peer.
- **0.6.2** — `bitcoin` no longer pinned to `=0.32.8`.
- **0.6.3** — info messages are a best-effort send, so a slow consumer can't stall the node loop that also serves `get_block` and `height_of_hash`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,9 +1077,9 @@ dependencies = [
 
 [[package]]
 name = "bip157"
-version = "0.6.0"
+version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b0baddc6c57202ec7bb76d2232aa243f8f529c4450bb873d55349f4d55ce94a2"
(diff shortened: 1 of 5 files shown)


dlukegordon (approved):
Not blocking this PR, but there are two upstream Kyoto issues worth calling out. `NeedConnections` does not wait for `add_peer`, so Kyoto can exit before processing the refill. Also, because the whitelist is LIFO, repeatedly re-adding hostnames can starve other resolved addresses and grow the whitelist. These should be fixed upstream.

PR #966, opened 2026-08-06
[btc-monitor] Wait for bitcoind to reach the start height

### LINKED ISSUE #963: [btc-monitor] A lagging bitcoind permanently anchors Kyoto at genesis
Found while reviewing #825. Separate from that PR, which only fixes the default start height.

## What happens

`checkpoint_at_height` treats bitcoind RPC error `-8` ("block height out of range") as permanent and immediately anchors at genesis:

https://github.com/MystenLabs/hashi/blob/main/crates/hashi/src/btc_monitor/monitor.rs#L283-L293

But `-8` is also what a bitcoind still in initial block download returns for a height it has not reached yet. That case is transient, and we give up on it in one shot.

`verify_bitcoind_network` (`crates/hashi/src/lib.rs:704`) only checks the chain, not the height, so nothing else catches it.

The anchor is resolved once in `Monitor::run` before the supervision loop starts (`monitor.rs:326`), so the genesis anchor sticks for the lifetime of the process — a running node never re-resolves it. And bip157 0.6.0 has no header persistence (`ChainState::Checkpoint` rebuilds the block tree from the checkpoint alone, `chain/chain.rs:53`; `data_dir` is only the peer store), so every supervision rebuild replays the chain from genesis again.

## Who hits it

An operator starting Hashi alongside a fresh bitcoind. After #825 the Signet default is `297756` and the Signet tip is ~316k, so a node that has not finished IBD is squarely in this window. Recovery is a Hashi restart once bitcoind has caught up, which is now documented in the runbook, but it should not need one.

## Options

1. Drop the `-8` early return so it takes the normal 5 x 2s retry path before falling back to genesis. Strictly better than today, no new failure modes, no config changes — but 10s does not cover a real IBD.
2. On `-8`, consult `getblockchaininfo` and keep waiting while `blocks < start_height`, logging progress. Correct semantics — the monitor cannot work without the anchor — but it makes BTC monitor startup block on bitcoind, so we need to decide what that does to readiness.
3. Preflight in `verify_bitcoind_network`: `ensure!(info.blocks >= start_height)` and fail fast with an actionable message. Simple and loud, but it refuses to boot on a deliberately future-dated start height, and it would need `default_start_height(Regtest)` changed to `0` or the regtest e2e tests break on the `800000` default.

I lean 1 as an immediate strict improvement, with 2 as the real fix if we are willing to define the readiness semantics.

0xsiddharthks:
## Summary

A node started next to a bitcoind that is still syncing anchored Kyoto at genesis for the life of the process: `getblockhash` returns `-8` for any height above the node's current tip, and we treated that as permanent. Every Kyoto rebuild then replayed the whole chain. The anchor is a floor the light client can't see below, so it can never be raised later; the right move is to wait for the block. Fixes #963.

The unused Kyoto `data_dir` cleanup that used to ride along here is now its own PR, #1090.

## Changes

- `checkpoint_at_height` retries until bitcoind has the block, logging its height every 30s. The `-8` special case and the genesis fallbacks are gone.
- Only errors that can clear are retried: `-8`, `-28`, connection-level I/O failures, HTTP 5xx. Anything else, including a reply that does not parse, fails the monitor, which exits the node.
- `default_start_height` is `0` for networks without a deployment anchor. The e2e harness needs this: regtest with no start height took the `800000` default.
- Each `MonitorClient` call carries one deadline. The caller times out on it, and the worker starts the bitcoind RPC only if corepc's fixed timeout, plus a margin, still fits before it, checked on the blocking thread. A caller that times out never has an RPC in flight; a request that can no longer fit gets an immediate error.
- The workspace declares `jsonrpc/minreq_http`, which the retry predicate names.
- Tests for the wait, a permanent RPC error, a malformed reply, the transport predicate against real errors, and the deadline check. Runbook §11 rewritten for the new log line.

### DIFF
--- a/crates/hashi/src/btc_monitor/monitor.rs
+++ b/crates/hashi/src/btc_monitor/monitor.rs
@@ -242,6 +242,8 @@ impl Monitor {
             .maximum_connection_time(Duration::MAX)
             .chain_state(kyoto::ChainState::Checkpoint(checkpoint));
 
+        // No-op as of bip157 0.6.3: the builder takes the path and `Node::new`
+        // discards it. Kept for when upstream implements persistence.
         if let Some(data_dir) = &config.data_dir {
             builder = builder.data_dir(data_dir.clone());
         }
@@ -263,19 +265,22 @@ impl Monitor {
             bitcoin::Network::Bitcoin if config.start_height > 481_823 => {
                 HashCheckpoint::segwit_activation()
             }
-            network => Self::checkpoint_at_height(bitcoind_rpc, config.start_height, network).await,
+            _ => Self::checkpoint_at_height(bitcoind_rpc, config.start_height).await,
         }
     }
 
+    /// Wait for bitcoind to have the block at `height`. A node still in initial
+    /// block download reports the same "out of range" error as a height past the
+    /// tip, so giving up would anchor Kyoto far below `height` permanently.
     async fn checkpoint_at_height(
         bitcoind_rpc: &Arc<corepc_client::client_sync::v29::Client>,
         height: u32,
-        network: bitcoin::Network,
     ) -> HashCheckpoint {
-        const MAX_ATTEMPTS: u32 = 5;
         const RETRY_DELAY: Duration = Duration::from_secs(2);
+        const LOG_EVERY: u32 = 15;
 
-        for attempt in 1..=MAX_ATTEMPTS {
+        let mut attempt: u32 = 0;
+        loop {
             match btc_rpc_call(bitcoind_rpc, move |rpc| rpc.get_block_hash(height as u64)).await {
                 Ok(raw) => match raw.into_model() {
                     Ok(model) => {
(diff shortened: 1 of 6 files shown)


dlukegordon (commented):
design/docs/node-operator-runbook.mdx:1217

This describes the old genesis fallback. Can we update it with the new waiting log and explain that Hashi continues automatically once bitcoind reaches the start height?

crates/hashi/src/btc_monitor/monitor.rs
```diff
@@ -282,29 +284,30 @@ impl Monitor {
...
-                // RPC error -8 "block height out of range": start_height is beyond
-                // the node's tip — permanent, so anchor at genesis instead of retrying.
-                Err(corepc_client::client_sync::Error::JsonRpc(jsonrpc::error::Error::Rpc(
-                    ref e,
-                ))) if e.code == -8 => {
-                    warn!(
-                        "Start height {height} is beyond bitcoind's tip; anchoring Kyoto at genesis"
-                    );
-                    return HashCheckpoint::from_genesis(network);
+                Err(e) => {
```
dlukegordon, line 304: This retries every error forever, including permanent errors like missing RPC permissions or invalid responses. Can we only retry `-8` and transient connection errors, and return other errors so the monitor fails clearly?

crates/hashi/src/btc_monitor/monitor.rs
```diff
@@ -1197,31 +1200,47 @@ impl MonitorClient {
...
+        message: impl FnOnce(oneshot::Sender<Result<T>>) -> MonitorMessage,
+        what: &str,
+    ) -> Result<T> {
         let (tx, rx) = oneshot::channel();
-        self.tx
-            .send(MonitorMessage::GetRecentFeeRate(conf_target, tx))
-            .await
-            .map_err(|e| anyhow::anyhow!(e))?;
-        rx.await.map_err(|e| anyhow::anyhow!(e))?
+        tokio::time::timeout(MONITOR_REQUEST_TIMEOUT, async {
```
dlukegordon, line 1243: Timing out here does not cancel a request that was already queued or started. This means `broadcast_transaction` can return an error and still broadcast later. Can we skip requests whose response channel is closed and use the same timeout for the underlying RPC?

0xsiddharthks:
@dlukegordon thanks! I addressed the comments . Could you please have another look.

jessiemongeon1:

crates/hashi/src/btc_monitor/monitor.rs
dlukegordon: Still one permanent-error path here: a malformed HTTP 200 response is `TransportError::Minreq(minreq::Error::SerdeJsonError)`, which matches `Some(_) => true`. `checkpoint_at_height` therefore retries invalid JSON forever. Please narrow the `Minreq` arm to actual transient connection/I/O errors and return parsing, protocol, and configuration errors.

crates/hashi/src/btc_monitor/monitor.rs
dlukegordon: This still has a race after `is_closed()`: a request queued during startup can be dispatched just before the 120-second caller deadline, pass the check, then the caller times out while the separately timed 60-second `spawn_blocking` RPC starts or continues. `send_raw_transaction` can therefore still broadcast after returning a timeout. Please propagate one authoritative deadline or request state through dispatch and the RPC side effect so the caller cannot time out while the broadcast can still run.

crates/hashi/src/btc_monitor/monitor.rs
0xsiddharthks: I see. yeah this makes sense. 
The transport arm is now an allowlist, and only `IoError` and `AddressNotFound` will retry. so other errors should fail the monitor.

Also, claude added a test for this.

crates/hashi/src/btc_monitor/monitor.rs
0xsiddharthks: gotcha. Yeah i had missed that.

Fixed this. The deadline now travels with the request.

The worker only starts the RPC if corepc's 60s (plus a margin) still fits before the caller's deadline, checked inside the blocking closure, otherwise it replies expired without touching bitcoind. 
The only thing left is minreq's own timeout, where a request already written can still get executed by bitcoind after it.

dlukegordon (approved):

PR #968, opened 2026-08-06
[guardian] add a guardian-tools image for the CI ceremony binaries

0xsiddharthks:
## Summary

sui-operations' deploy pipeline builds `hashi-guardian` and `hashi-guardian-init` from scratch in two separate jobs, on 2-vCPU runners with no cargo cache — about 45 minutes per full redeploy for a tree that includes blst, fastcrypto, sequoia-openpgp and kyoto. This packages both so mp3 can build them once, on its warm cache, and publish them to `gs://mysten-hashi-binaries` alongside the hashi CLI.

Both are `non-enclave-dev` builds: the guardian emits a mock attestation document and init accepts one, which is what lets a runner-local ceremony verify against dev stacks whose EIF sets `eif-features=non-enclave-dev`. They are packaged as `hashi-guardian-dev` and `hashi-guardian-init-dev` so that can never be mistaken for an enclave-grade build in the bucket they share with the CLI.

Deliberately its own Containerfile rather than another stage in `docker/hashi/Containerfile`. Cargo unifies features per invocation, so building these next to the hashi CLI would silently ship a CLI whose attestation verification is a no-op — separate images make that impossible.

Verified on the built artifacts: `hashi-guardian-dev` contains the mock attestation document and `hashi-guardian-init-dev` the disabled-verification warning, while `cargo tree -f '{p} :: {f}'` shows hashi-types resolving *without* `non-enclave-dev` under the CLI's own selection. Both are static musl x86-64 and `--help` runs.

## Changes

- `docker/hashi-guardian-tools/Containerfile` — deps/build/package stages mirroring `docker/hashi-guardian-k8s`, emitting both binaries under `/opt/hashi/bin`
- `docker/hashi-guardian-tools/build.sh` — local build helper

Consumed by MystenLabs/sui-operations#8628, which needs this merged and one image build first.

### DIFF
--- a/docker/hashi-guardian-tools/Containerfile
+++ b/docker/hashi-guardian-tools/Containerfile
@@ -1,0 +1,100 @@
+# syntax=docker/dockerfile:1
+
+# The guardian binaries the deploy pipeline's CI ceremony runs, built once here
+# instead of three times from a cold cargo cache on GitHub runners
+# (sui-operations scripts/hashi/{guardian-ceremony,bootstrap-guardian}.sh).
+#
+# Both carry `non-enclave-dev`: the guardian emits a mock attestation document
+# and init accepts one, which is what lets a runner-local ceremony verify
+# against dev stacks whose EIF sets eif-features=non-enclave-dev. That is the
+# ONLY thing this image is for — the `-dev` suffix on the packaged names is
+# deliberate, so these can never be mistaken for enclave-grade builds in the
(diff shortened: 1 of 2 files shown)


0xsiddharthks:
Closing — its only consumer, sui-operations#8628, is closed. That change prebuilt the guardian ceremony binaries on mp3 instead of compiling them on each runner; it added too much machinery for the time it saved, so this image has nothing left to feed.

PR #969, opened 2026-08-06
mpc: extract helpers

jonas-lj:
Extract two helpers in the `mpc` module, `time_async`, `params` and `with_manager_blocking`, and use them across `mpc_except_signing.rs` and `service.rs`.

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -407,7 +407,7 @@ impl MpcManager {
     }
 
     /// Run `f` against a write-locked manager on a blocking thread.
-    fn with_manager_blocking_mut<T: Send + 'static>(
+    pub(crate) fn with_manager_blocking_mut<T: Send + 'static>(
         mpc_manager: &Arc<RwLock<Self>>,
         f: impl FnOnce(&mut Self) -> T + Send + 'static,
     ) -> Pin<Box<dyn Future<Output = T> + Send>> {
@@ -1605,19 +1605,18 @@ impl MpcManager {
         metrics: &Metrics,
     ) -> MpcResult<()> {
         // TODO(Optimization): Skip dealer phase if certificate is already on TOB
-        let _timer = metrics
-            .mpc_dealer_crypto_duration_seconds
-            .with_label_values(&[MPC_LABEL_KEY_ROTATION])
-            .start_timer();
         let dealer_data = {
             let previous = previous.clone();
             let mut rng = StdRng::from_entropy();
-            Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
-                mgr.prepare_rotation_dealer_flow(&previous, &mut rng)
-            })
+            time_async(
+                &metrics.mpc_dealer_crypto_duration_seconds,
+                MPC_LABEL_KEY_ROTATION,
+                Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
+                    mgr.prepare_rotation_dealer_flow(&previous, &mut rng)
+                }),
+            )
             .await?
         };
-        drop(_timer);
         let mut aggregator = BlsSignatureAggregator::new_reduced(
             &dealer_data.committee,
             dealer_data.messages_hash.clone(),
@@ -1788,26 +1787,29 @@ impl MpcManager {
                 }
             }
             {
-                let _timer = metrics
-                    .mpc_message_process_duration_seconds
-                    .with_label_values(&[MPC_LABEL_KEY_ROTATION])
-                    .start_timer();
                 let previous = previous.clone();
                 let share_indices = dealer_share_indices.clone();
-                Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
-                    if share_indices.iter().any(|idx| {
-                        !mgr.dealer_outputs
-                            .contains_key(&DealerOutputsKey::Rotation(dealer, *idx))
-                            && !mgr
-                                .complaints_to_process
-                                .contains_key(&ComplaintsToProcessKey::Rotation(dealer, *idx))
-                    }) {
-                        mgr.process_certified_rotation_message(&dealer, &previous, &share_indices)?;
-                    }
-                    Ok::<_, MpcError>(())
-                })
+                time_async(
+                    &metrics.mpc_message_process_duration_seconds,
+                    MPC_LABEL_KEY_ROTATION,
+                    Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
+                        if share_indices.iter().any(|idx| {
+                            !mgr.dealer_outputs
+                                .contains_key(&DealerOutputsKey::Rotation(dealer, *idx))
+                                && !mgr
+                                    .complaints_to_process
+                                    .contains_key(&ComplaintsToProcessKey::Rotation(dealer, *idx))
+                        }) {
+                            mgr.process_certified_rotation_message(
+                                &dealer,
+                                &previous,
+                                &share_indices,
+                            )?;
+                        }
+                        Ok::<_, MpcError>(())
+                    }),
+                )
                 .await?;
-                drop(_timer);
             }
             let (signers, epoch, rotation_msgs) = {
                 let mgr = mpc_manager.read().unwrap();
@@ -1873,18 +1875,17 @@ impl MpcManager {
             );
         }
         tracing::info!("run_key_rotation_as_party: threshold met, calling complete_key_rotation",);
-        let _timer = metrics
-            .mpc_completion_duration_seconds
-            .with_label_values(&[MPC_LABEL_KEY_ROTATION])
-            .start_timer();
         let output = {
             let previous = previous.clone();
-            Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
-                mgr.complete_key_rotation(&previous, &certified_share_indices)
-            })
+            time_async(
+                &metrics.mpc_completion_duration_seconds,
+                MPC_LABEL_KEY_ROTATION,
+                Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
+                    mgr.complete_key_rotation(&previous, &certified_share_indices)
+                }),
+            )
             .await?
         };
-        drop(_timer);
         Ok(output)
     }
 
@@ -5579,20 +5580,25 @@ impl MpcManager {
                 Ok(output) => output,
                 Err(e) => {
                     tracing::info!("Reconstruction failed ({e}), falling back to new-member path");
-                    let _fetch_timer = metrics
-                        .mpc_prepare_previous_fetch_public_output_duration_seconds
-                        .with_label_values(&[MPC_LABEL_KEY_ROTATION])
-                        .start_timer();
-                    Self::fetch_and_build_public_output(mpc_manager, p2p_channel, threshold_opt)
-                        .await?
+                    time_async(
+                        &metrics.mpc_prepare_previous_fetch_public_output_duration_seconds,
+                        MPC_LABEL_KEY_ROTATION,
+                        Self::fetch_and_build_public_output(
+                            mpc_manager,
+                            p2p_channel,
+                            threshold_opt,
+                        ),
+                    )
+                    .await?
                 }
             }
         } else {
-            let _fetch_timer = metrics
-                .mpc_prepare_previous_fetch_public_output_duration_seconds
-                .with_label_values(&[MPC_LABEL_KEY_ROTATION])
-                .start_timer();
-            Self::fetch_and_build_public_output(mpc_manager, p2p_channel, threshold_opt).await?
+            time_async(
+                &metrics.mpc_prepare_previous_fetch_public_output_duration_seconds,
+                MPC_LABEL_KEY_ROTATION,
+                Self::fetch_and_build_public_output(mpc_manager, p2p_channel, threshold_opt),
+            )
+            .await?
         };
         tracing::info!(
             "prepare_previous_output: is_member_of_previous_committee={is_member_of_previous_committee}, \
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -410,6 +410,14 @@ impl MpcManager {
         spawn_blocking(move || f(&mut mgr.write().unwrap())).await
     }
 
+    /// The AVSS threshold parameters (`t`, `f`) for this manager's epoch.
+    fn params(&self) -> Parameters {
+        Parameters {
+            t: self.mpc_config.threshold,
+            f: self.mpc_config.max_faulty,
+        }
+    }
+
     // Only for devnet key recovery CLI tool
     pub fn handle_send_messages_request(
         &mut self,
@@ -2208,17 +2216,8 @@ impl MpcManager {
     ) -> avss::Message {
         let dealer_session_id = self.current_session_id().dealer_session_id(&self.address);
         let nodes = self.maybe_corrupt_nodes_for_testing(&self.mpc_config.nodes);
-        let dealer = avss::Dealer::new(
-            None,
-            nodes,
-            Parameters {
-                t: self.mpc_config.threshold,
-                f: self.mpc_config.max_faulty,
-            },
-            dealer_session_id.to_vec(),
-            rng,
-        )
-        .expect("checked threshold above");
+        let dealer = avss::Dealer::new(None, nodes, self.params(), dealer_session_id.to_vec(), rng)
+            .expect("checked threshold above");
         dealer.create_message(rng)
     }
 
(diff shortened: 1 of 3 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1214,20 +1242,15 @@ impl MpcManager {
...
-            .with_label_values(&[MPC_LABEL_DKG])
-            .start_timer();
-        let dealer_data = {
-            let mgr = Arc::clone(mpc_manager);
-            spawn_blocking(move || {
+        let dealer_data = time_async(
+            &metrics.mpc_dealer_crypto_duration_seconds,
+            MPC_LABEL_DKG,
+            Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
                 let mut rng = rand::thread_rng();
```
zhouwfang: thread_rng() is now inside the write guard, since the helper locks before calling the closure. Its lazy init can panic on a fresh blocking thread, which would poison the manager lock instead of unwinding clean. Not sure there's a clean fix given the closure has to be Send.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -352,6 +355,35 @@ impl MpcManager {
         Ok(manager)
     }
 
+    /// Run `f` against a read-locked manager on a blocking thread.
+    pub(crate) fn with_manager_blocking<T: Send + 'static>(
+        mpc_manager: &Arc<RwLock<Self>>,
+        f: impl FnOnce(&Self) -> T + Send + 'static,
+    ) -> Pin<Box<dyn Future<Output = T> + Send>> {
+        let mgr = Arc::clone(mpc_manager);
+        // This is boxed to keep the future off the caller's stack. If inlined and deeply nested,
```
zhouwfang: nit: an inlined future isn't on the caller's stack, it's a field of the enclosing future object. Maybe say it caps the enclosing future's size, since that's what actually moved.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -6598,6 +6502,22 @@ fn consume_certified_nonce_outputs<T>(
     (pre_filter, dealers, outputs)
 }
 
+/// Time an async operation on `metric[label]`, holding the timer until `f` completes.
```
zhouwfang, line 6510: Could mention the timer starts when time_async is called, not on first poll.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -352,6 +355,35 @@ impl MpcManager {
...
+        mpc_manager: &Arc<RwLock<Self>>,
+        f: impl FnOnce(&mut Self) -> T + Send + 'static,
+    ) -> Pin<Box<dyn Future<Output = T> + Send>> {
+        let mgr = Arc::clone(mpc_manager);
+        // Same as with_manager_blocking.
+        Box::pin(async move { spawn_blocking(move || f(&mut mgr.write().unwrap())).await })
+    }
+
+    /// The AVSS threshold parameters (`t`, `f`) for this manager's epoch.
+    fn params(&self) -> Parameters {
```
zhouwfang: Could this be pub(crate)? service.rs:524 and :643 hand-build the same expression and can't reach it.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1506,11 +1513,9 @@ impl MpcManager {
             .with_label_values(&[MPC_LABEL_KEY_ROTATION])
             .start_timer();
         let dealer_data = {
-            let mgr = Arc::clone(mpc_manager);
             let previous = previous.clone();
-            spawn_blocking(move || {
+            Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
```
zhouwfang: Converted the lock here but left the timer just above it manual, while DKG and nonce-generation on the same metric moved to time_async. Converting also drops the previous.clone() out of the histogram.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1506,11 +1513,9 @@ impl MpcManager {
...
                 let mut rng = rand::thread_rng();
```
zhouwfang: same as line 1249

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1810,20 +1810,15 @@ impl MpcManager {
...
-            .with_label_values(&[MPC_LABEL_NONCE_GENERATION])
-            .start_timer();
-        let dealer_data = {
-            let mgr = Arc::clone(mpc_manager);
-            spawn_blocking(move || {
+        let dealer_data = time_async(
+            &metrics.mpc_dealer_crypto_duration_seconds,
+            MPC_LABEL_NONCE_GENERATION,
+            Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
                 let mut rng = rand::thread_rng();
```
zhouwfang: same as 1249

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2883,20 +2846,15 @@ impl MpcManager {
...
-            .with_label_values(&[MPC_LABEL_NONCE_GENERATION])
-            .start_timer();
-        let mut dealer_data = {
-            let mgr = Arc::clone(mpc_manager);
-            spawn_blocking(move || {
+        let mut dealer_data = time_async(
+            &metrics.mpc_dealer_crypto_duration_seconds,
+            MPC_LABEL_NONCE_GENERATION,
+            Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
                 let mut rng = rand::thread_rng();
```
zhouwfang: same as 1249

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1703,11 +1707,9 @@ impl MpcManager {
                     .mpc_message_process_duration_seconds
                     .with_label_values(&[MPC_LABEL_KEY_ROTATION])
                     .start_timer();
-                let mgr = Arc::clone(mpc_manager);
                 let previous = previous.clone();
                 let share_indices = dealer_share_indices.clone();
-                spawn_blocking(move || {
-                    let mut mgr = mgr.write().unwrap();
+                Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
```
zhouwfang: Left the timer above this manual while DKG and nonce-generation on the same metric moved to time_async at :1398 and :1967. Converting also drops both clones out of the histogram.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1791,10 +1793,8 @@ impl MpcManager {
             .with_label_values(&[MPC_LABEL_KEY_ROTATION])
             .start_timer();
         let output = {
-            let mgr = Arc::clone(mpc_manager);
             let previous = previous.clone();
-            spawn_blocking(move || {
-                let mut mgr = mgr.write().unwrap();
+            Self::with_manager_blocking_mut(mpc_manager, move |mgr| {
```
zhouwfang: Left the timer above this manual while DKG on the same metric moved to time_async at :1493. Converting also drops the previous.clone() out of the histogram.

zhouwfang (commented):
I measured the spawned task future at service.rs:129: 13176 bytes on main, 40592 pre-boxing, 13176 here. So the boxing does exactly what you wanted. But 40 KB shouldn't overflow a 2 MB stack, so the mechanism in the commit message doesn't quite add up. Do you still have the backtrace? Minor: "contributes only a pointer" is a fat pointer and the inner future is still built on the stack and then memcpy'd into the box.

Couple of things that fall outside the diff hunks so I can't put them inline.
mpc_except_signing.rs:5392 and :5401 are single-statement timed regions, so time_async candidates too. And rpc/service.rs:49 and :112 still have the spawn_blocking plus write-lock shape the new helper replaces, but with_manager_blocking_mut is private while its read twin is pub(crate).

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -6598,6 +6502,22 @@ fn consume_certified_nonce_outputs<T>(
...
+pub(crate) fn time_async<'a, F>(
+    metric: &HistogramVec,
+    label: &str,
+    f: F,
+) -> Pin<Box<dyn Future<Output = F::Output> + Send + 'a>>
+where
+    F: Future + Send + 'a,
+{
+    let timer = metric.with_label_values(&[label]).start_timer();
+    Box::pin(async move {
```
zhouwfang, line 6522: Worth the same comment you put on with_manager_blocking. Otherwise this reads like a deletable allocation.

jonas-lj:
> I measured the spawned task future at service.rs:129: 13176 bytes on main, 40592 pre-boxing, 13176 here. So the boxing does exactly what you wanted. But 40 KB shouldn't overflow a 2 MB stack, so the mechanism in the commit message doesn't quite add up. Do you still have the backtrace? Minor: "contributes only a pointer" is a fat pointer and the inner future is still built on the stack and then memcpy'd into the box.

Yeah, you're right, the size framing is wrong. What I do know is boxing fixed the overflow in debug here, and nothing's a big object (run is 13 KB in debug). So it's likely poll-stack depth, not object size. The deep chain in debug was probably already near the 2 MB ceiling and the extra async layers tipped it over?

And no, there's no backtrace in the logs. I think we need to catch one live in the debugger, and I couldn't reproduce locally.

crates/hashi/src/mpc/mpc_except_signing.rs
jonas-lj: We now seed StdRng::from_entropy() before the lock (it's Send, unlike ThreadRng, and impl AllowedRng as required) and move it into the closure, so the entropy read can't poison the lock on panic.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1218,20 +1248,15 @@ impl MpcManager {
...
-        let _timer = metrics
-            .mpc_dealer_crypto_duration_seconds
-            .with_label_values(&[MPC_LABEL_DKG])
-            .start_timer();
-        let dealer_data = {
-            let mgr = Arc::clone(mpc_manager);
-            spawn_blocking(move || {
-                let mut rng = rand::thread_rng();
-                let mut mgr = mgr.write().unwrap();
+        let mut rng = StdRng::from_entropy();
```
zhouwfang, line 1251: It doesn't fully close it for DKG. try_sign_dkg_message goes through process_avss_message, which still does thread_rng() under the guard at :6372, and that only panics on first use per thread so the exposure here is unchanged. Same via send_messages. Nonce generation seems clean. Could thread an rng through process_avss_message if you want these two paths covered too.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -1505,21 +1514,18 @@ impl MpcManager {
...
-            .start_timer();
         let dealer_data = {
-            let mgr = Arc::clone(mpc_manager);
             let previous = previous.clone();
-            spawn_blocking(move || {
-                let mut rng = rand::thread_rng();
-                let mut mgr = mgr.write().unwrap();
-                mgr.prepare_rotation_dealer_flow(&previous, &mut rng)
-            })
+            let mut rng = StdRng::from_entropy();
```
zhouwfang, line 1518: Same for rotation, via try_sign_rotation_messages into process_avss_message at :4701.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -352,6 +357,35 @@ impl MpcManager {
...
+    pub(crate) fn with_manager_blocking_mut<T: Send + 'static>(
+        mpc_manager: &Arc<RwLock<Self>>,
+        f: impl FnOnce(&mut Self) -> T + Send + 'static,
+    ) -> Pin<Box<dyn Future<Output = T> + Send>> {
+        let mgr = Arc::clone(mpc_manager);
+        // Same as with_manager_blocking.
+        Box::pin(async move { spawn_blocking(move || f(&mut mgr.write().unwrap())).await })
+    }
+
+    /// The AVSS threshold parameters (`t`, `f`) for this manager's epoch.
```
zhouwfang: Now that it's pub(crate), "AVSS" is a bit narrow, service.rs:522 uses it for presig derivation. Might also be worth saying current mpc_config epoch, since generate_presignatures has an epoch arg in scope.

zhouwfang:
> Yeah, you're right, the size framing is wrong. What I do know is boxing fixed the overflow in debug here, and nothing's a big object (run is 13 KB in debug). So it's likely poll-stack depth, not object size. The deep chain in debug was probably already near the 2 MB ceiling and the extra async layers tipped it over?
> 
> And no, there's no backtrace in the logs. I think we need to catch one live in the debugger, and I couldn't reproduce locally.

It's not depth instead of size, the two together. And boxing doesn't reduce depth at all, so if boxing is what fixed it, size has to be doing the work. The 13 KB is post-boxing. MpcService::run is 40,496 at the pre-boxing commit and 13,080 both before the PR and at head. So the objects did triple and come back, you're just measuring after the fix.

Guess at why size matters, not measured: in debug the child future is built on the stack before moving into the parent's slot, so each frame carries the child's size. Cheap test if you want one: the async workers run on the blocking pool, and we never set thread_stack_size, so RUST_MIN_STACK controls their stacks. Run the pre-boxing commit with RUST_MIN_STACK=8388608 and see if it still overflows.

jonas-lj:
> > Yeah, you're right, the size framing is wrong. What I do know is boxing fixed the overflow in debug here, and nothing's a big object (run is 13 KB in debug). So it's likely poll-stack depth, not object size. The deep chain in debug was probably already near the 2 MB ceiling and the extra async layers tipped it over?
> > And no, there's no backtrace in the logs. I think we need to catch one live in the debugger, and I couldn't reproduce locally.
> 
> It's not depth instead of size, the two together. And boxing doesn't reduce depth at all, so if boxing is what fixed it, size has to be doing the work. The 13 KB is post-boxing. MpcService::run is 40,496 at the pre-boxing commit and 13,080 both before the PR and at head. So the objects did triple and come back, you're just measuring after the fix.
> 
> Guess at why size matters, not measured: in debug the child future is built on the stack before moving into the parent's slot, so each frame carries the child's size. Cheap test if you want one: the async workers run on the blocking pool, and we never set thread_stack_size, so RUST_MIN_STACK controls their stacks. Run the pre-boxing commit with RUST_MIN_STACK=8388608 and see if it still overflows.

You're right, it's size. I ran CI on the pre-boxing commit with 8MB stack size and no overflows. A ~27 KB delta tipped it, so debug is probably already near the 2 MB ceiling.

crates/hashi/src/mpc/mpc_except_signing.rs
jonas-lj: You're right. I traced it and it's actually broader than those two since the complaint, reconstruction and certified-message paths all go through process_avss_message under the guard too. Threading an rng through all of those touches a lot of code and since it's pre-existing and unchanged by this PR I'll do it as a follow-up.

crates/hashi/src/mpc/mpc_except_signing.rs
jonas-lj: Same as https://github.com/MystenLabs/hashi/pull/969#discussion_r3748350303

crates/hashi/src/mpc/mpc_except_signing.rs
jonas-lj: https://github.com/MystenLabs/hashi/pull/987

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -352,6 +357,35 @@ impl MpcManager {
         Ok(manager)
     }
 
+    /// Run `f` against a read-locked manager on a blocking thread.
+    pub(crate) fn with_manager_blocking<T: Send + 'static>(
+        mpc_manager: &Arc<RwLock<Self>>,
+        f: impl FnOnce(&Self) -> T + Send + 'static,
+    ) -> Pin<Box<dyn Future<Output = T> + Send>> {
+        let mgr = Arc::clone(mpc_manager);
+        // Boxed so it isn't inlined into the enclosing future; deeply nested, the inlined
```
zhouwfang, line 404: nit: Now that you've measured it, what about putting the number in?

crates/hashi/src/mpc/rpc/service.rs
```diff
@@ -46,20 +46,20 @@ impl MpcService for HttpService {
             .mpc_rpc_handler_process_duration_seconds
```
zhouwfang: This one's still the manual timer around a single awaited helper, same shape you converted in mpc_except_signing. time_async is pub(crate) so it reaches here.

zhouwfang (approved):

zhouwfang:
@jonas-lj PTAL at #1028. I found the issue while reviewing your PR. Thanks.

PR #970, opened 2026-08-06
[guardian] Gate the enclave workspace resolve in CI

0xsiddharthks:
## Summary

`guardian-enclave-lock` regenerates the lock on a GitHub runner, where resolution succeeds — so it would not have caught #960, whose failure was specific to the enclave's own cargo. This adds a CI job that resolves the pruned workspace in that toolchain instead. It fails before compiling anything, so it costs a minute rather than an EIF build.

## Changes

- Split the Containerfile's pruned tree into a `workspace` stage, with a `resolve` stage on top that only runs `cargo metadata --locked`.
- Moved the kernel copy from `base` into `build`, so resolving doesn't wait on the kernel.
- Added `make -C docker/hashi-guardian resolve-check` and a `guardian-enclave-resolve` CI job that runs it.

PCR0 is unaffected — `eif_build` still measures the same kernel, ramdisk and cmdline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,6 +135,19 @@ jobs:
 
       - run: make is-dirty
 
+  # guardian-enclave-lock resolves on this runner, where #960's failure does not
+  # reproduce. Resolve in the enclave's own toolchain instead; it fails before
+  # compiling anything, so this costs a minute rather than an EIF build.
+  guardian-enclave-resolve:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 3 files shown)


mskd12 (approved):

0xsiddharthks:
Superseded by #1036, which deletes enclave-Cargo.lock outright: the EIF now builds the {hashi-guardian, hashi-types} slice against the root Cargo.lock, and the resolve gate from this PR lives on there (same resolve-check target and guardian-enclave-resolve job, now validating the real build in the enclave toolchain, with the kernel COPYs moved after the cargo build so the gate never waits on the kernel).

PR #975, opened 2026-08-06
test(e2e): run every e2e_flow test against an upgraded chain

Bridgerz:
## Summary

Stacked on #924. Closes the e2e coverage gap where only the dedicated upgrade tests ran against an upgraded chain — and per review, the boot lives in the **builder itself**, so every builder user gets it, not just `setup_test_networks` callers.

## Changes

- `TestNetworksBuilder::build()` now publishes the deployed **v1 bytecode snapshot** (defaulted when no snapshot dir is set) and runs the full governance upgrade to the current source before returning. `without_upgrade()` opts out where v1-only is the point (`snapshot_signing_tests`) or the test drives its own upgrade (`snapshot_v1_upgrades_to_current_source`).
- `test_upgrade_v1_to_v2` → `test_upgrade_via_proposal`: rides the default boot (chain already at vN) and explicitly upgrades vN → vN+1 through the proposal flow, keeping dedicated coverage of governance-driven upgrades.
- That second upgrade surfaced two latent bugs, fixed here:
  - `prepare_upgrade_package` parameterizes the patched `PACKAGE_VERSION` and points `published-at` at the **latest** published package id (a v1 `published-at` fails the ticket check on the second upgrade);
  - `build_upgrade_execution_transaction` takes the current package id for the `Upgrade` command — referencing the original id fails `PackageIDDoesNotMatch`.
- `wait_for_package_convergence` moves into `upgrade_flow` as the shared post-upgrade barrier.

## Verification

- `test_upgrade_via_proposal` passes locally: snapshot v1 → builder auto-upgrade v2 → explicit proposal upgrade v3 → convergence → disable v1 → v1-entry rejection (66s).
- `test_bitcoin_deposit_e2e_flow` passes under the builder-default boot (56s).
- `cargo check -p hashi -p e2e-tests --tests` clean.

### DIFF
--- a/crates/e2e-tests/src/e2e_flow.rs
+++ b/crates/e2e-tests/src/e2e_flow.rs
@@ -133,9 +133,20 @@ mod tests {
             .ok_or_else(|| anyhow!("transaction effects BCS bytes were not returned"))
     }
 
+    /// Boot a network in the state a deployed chain actually runs in: the v1
+    /// bytecode snapshot is published first, then upgraded to the current
+    /// source through the full governance flow. Every test in this module
+    /// therefore exercises the post-upgrade configuration — package history
+    /// {v1, vN}, routing at the upgraded package id, types defined at the v1
+    /// address — not a fresh single-version publish no real network is in.
+    /// Fresh-publish coverage lives in the genesis-specific harnesses
+    /// (`upgrade_tests`, `snapshot_signing_tests` manage their own boots).
     async fn setup_test_networks(builder: TestNetworksBuilder) -> Result<TestNetworks> {
         info!("Setting up test networks...");
-        let networks = builder.build().await?;
+        let mut networks = builder
+            .with_v1_from_snapshot(crate::snapshot::default_snapshot_dir()?)
+            .build()
+            .await?;
 
         info!("Test networks initialized");
         info!("  - Sui RPC: {}", networks.sui_network.rpc_url);
@@ -147,6 +158,16 @@ mod tests {
             .wait_for_mpc_key(Duration::from_secs(60))
             .await?;
         info!("MPC key ready");
+
+        info!("Upgrading the chain to the current source...");
+        let new_package_id = crate::upgrade_flow::execute_full_upgrade(&mut networks).await?;
+        crate::upgrade_flow::wait_for_package_convergence(
+            &networks,
+            new_package_id,
+            Duration::from_secs(30),
+        )
+        .await?;
+        info!("Chain upgraded; tests run against the post-upgrade state");
 
         Ok(networks)
     }
--- a/crates/e2e-tests/src/e2e_flow.rs
+++ b/crates/e2e-tests/src/e2e_flow.rs
@@ -133,20 +133,13 @@ mod tests {
             .ok_or_else(|| anyhow!("transaction effects BCS bytes were not returned"))
     }
 
-    /// Boot a network in the state a deployed chain actually runs in: the v1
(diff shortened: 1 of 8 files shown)


crates/e2e-tests/src/e2e_flow.rs
```diff
@@ -133,9 +133,20 @@ mod tests {
...
+    /// address — not a fresh single-version publish no real network is in.
+    /// Fresh-publish coverage lives in the genesis-specific harnesses
+    /// (`upgrade_tests`, `snapshot_signing_tests` manage their own boots).
     async fn setup_test_networks(builder: TestNetworksBuilder) -> Result<TestNetworks> {
         info!("Setting up test networks...");
-        let networks = builder.build().await?;
+        let mut networks = builder
+            .with_v1_from_snapshot(crate::snapshot::default_snapshot_dir()?)
+            .build()
+            .await?;
```
bmwill: I think this likely needs to be done inside of the builder itself so that all uses of the builder properly do the upgrade flow

crates/e2e-tests/src/lib.rs
```diff
@@ -407,6 +430,24 @@ impl TestNetworksBuilder {
...
+            tracing::info!("upgrading the chain to the current source...");
+            let new_package_id = upgrade_flow::execute_full_upgrade(&mut test_networks).await?;
+            upgrade_flow::wait_for_package_convergence(
+                &test_networks,
+                new_package_id,
+                std::time::Duration::from_secs(30),
+            )
+            .await?;
+            tracing::info!("chain upgraded; the network runs the post-upgrade state");
+        }
```
bmwill, line 449: we technically don't need to use the proposal upgrade flow for this, we could do the package upgrade normally before finalizing the publish to save on ci time

bmwill (approved):

crates/e2e-tests/src/e2e_flow.rs
Bridgerz: Done — `build()` owns the snapshot→upgrade boot now, with `without_upgrade()` as the opt-out for the tests where v1-only is the point (snapshot signing) or the test drives its own upgrade. You were right that the setup helper wasn't enough: six direct `.build()` call sites in `e2e_flow` were bypassing it.

Bonus from the rework: the explicit-proposal test now upgrades vN→vN+1 on top of the default boot, which surfaced two latent second-upgrade bugs (stale `published-at`, and `build_upgrade_execution_transaction` passing the original package id to the `Upgrade` command → `PackageIDDoesNotMatch`) — both fixed here.

crates/e2e-tests/src/lib.rs
Bridgerz: Tried it — it can't work against the deployed v1 bytecode: the single-PTB fold trips Sui's private-entry rule (finish_publish is entry, and the cap is tainted by 0x2::authorize_upgrade), committing first fails EWrongUpgradeCap (the cap then names the v2 id), and custody-first leaves no pre-genesis path to a ticket. Proposal: keep the governance boot for now (its real premium over the fold is only the proposal/vote txs, ~5-8s/test — the Move build dominates and is common to both), and land a public fun finish_publish_with_upgrade(self, cap, receipt, …) in v2 source — public escapes the entry-taint rule, and checking the cap against the original id before committing the receipt keeps the launch authorization sound. Once the v2 snapshot is checked in, the builder flips to the folded single-PTB boot.

bmwill: lame but understandable

PR #976, opened 2026-08-06
versioning: source carries the next package version

Bridgerz:
## Summary

Stacked on #975 (→ #924 → #923). `PACKAGE_VERSION` in source declares the version the tree *ships as*, bumped at cycle start rather than patched into artifacts at build time.

## Changes

- `versioning.move`: `PACKAGE_VERSION = 2` with the policy documented (a mid-cycle tree must land as an upgrade, never a fresh publish — genesis `create()` enables `{const}` while a fresh publish is Sui sequence version 1, permanently disjoint).
- With #975's builder-owned snapshot→upgrade boot, no harness or localnet changes are needed here: the harness version patch becomes a no-op rewrite of the same value, so tests upgrade to the literal artifact a deployment would publish, and `build_upgrade_package`'s `declared == published+1` check enforces the policy end to end. Manual-mode localnet (no committee at build) stays on snapshot v1 for the operator to upgrade through governance.
- Move tests that assumed const = 1 pin to versions relative to the current const (enable/disable 3; current-version guards use 2).

## Verification

- `sui move test`: 148/148.
- `test_upgrade_via_proposal` passes at const = 2: builder auto-upgrade patches 2→2 (no-op), explicit upgrade goes 2→3 (66s).

## Notes

- `SUPPORTED_PACKAGE_VERSIONS` stays `[1]`; it grows to `[1, 2]` with the change that implements v2 semantics (#841 re-land).
- On #923 in this stack: a permanently disjoint enabled/published state (mid-cycle tree fresh-published) now warns loudly instead of an indistinguishable `NotReady`.

### DIFF
--- a/packages/hashi/sources/core/versioning.move
+++ b/packages/hashi/sources/core/versioning.move
@@ -14,7 +14,13 @@ use sui::{package::{Self, UpgradeCap, UpgradeTicket, UpgradeReceipt}, vec_set::{
 
 // ~~~~~~~ Constants ~~~~~~~
 
-const PACKAGE_VERSION: u64 = 1;
+/// The package version this tree ships as: the deployed version while the
+/// tree matches a deployment, bumped to the next version when a release
+/// cycle's changes begin landing. A mid-cycle tree must reach a chain as an
+/// upgrade (snapshot v1 + upgrade in every dev harness), never as a fresh
+/// publish — `create` enables this constant, and a fresh publish is Sui
+/// sequence version 1, so a mid-cycle fresh publish can never activate.
+const PACKAGE_VERSION: u64 = 2;
 
 // ~~~~~~~ Errors ~~~~~~~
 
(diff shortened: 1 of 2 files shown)


crates/e2e-tests/src/upgrade_flow.rs
```diff
@@ -92,19 +95,6 @@ pub fn prepare_upgrade_package(test_dir: &Path, original_package_id: Address) ->
...
-    let versioning_src = std::fs::read_to_string(&versioning_path)?;
-    let patched = versioning_src.replace(
-        "const PACKAGE_VERSION: u64 = 1;",
-        "const PACKAGE_VERSION: u64 = 2;",
-    );
-    anyhow::ensure!(
-        patched != versioning_src,
-        "PACKAGE_VERSION replacement failed — pattern not found in versioning.move"
-    );
-    std::fs::write(&versioning_path, patched)?;
```
bmwill: I think we do still want a test that explicitly updates via proposal like this does so we'll want to change this from 1 -> 2 to likely 2 -> 3 since we're updating the source to 2 and we want v2 to be publish and ready as the test networks builder finishes

bmwill (approved):

PR #977, opened 2026-08-06
test: pin the package-version bump policy in CI

Bridgerz:
## Summary

Stacked on #976 (→ #975 → #924 → #923). Closes the enforcement gap discussed on the stack: the e2e harness patches `PACKAGE_VERSION` when building upgrade artifacts, so no running test notices a missed or doubled bump — the first failure would be the real upgrade build on deploy day (`build_upgrade_package`'s `declared == published + 1` check).

## Changes

- `crates/hashi/tests/version_policy_tests.rs`: a plain test (runs in the existing `test` CI job, no workflow wiring) asserting `versioning.move` declares exactly one past `Published.toml`'s deployed testnet version — the same source of truth the compat gate and snapshot boots key off.
- Sharpens the policy stated on the constant: the bump rides the change that records a deployment (Published.toml + snapshot capture); mid-cycle Move changes never touch it. The assertion message says exactly that, so the failing PR knows what to do.

## Verification

- Test passes at the current state (declared 2, deployed 1).
- Flipping either side locally fails with the policy message.

### DIFF
--- a/crates/hashi/tests/version_policy_tests.rs
+++ b/crates/hashi/tests/version_policy_tests.rs
@@ -1,0 +1,48 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Pins the package-version bump policy in CI.
+//!
+//! The source tree always declares the *next* package version: exactly one
+//! past the deployed testnet package recorded in `Published.toml`. The e2e
+//! harness patches the constant when building upgrade artifacts, so no test
+//! that runs the code would notice a missed or doubled bump — the first
+//! thing that would is the real upgrade build on deploy day
+//! (`build_upgrade_package`'s `declared == published + 1` check). This test
+//! moves that failure to the PR that should have carried the bump:
+//!
+//! - after deploying an upgrade, bump `PACKAGE_VERSION` in the same change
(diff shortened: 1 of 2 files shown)


crates/hashi/tests/version_policy_tests.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let entries = hashi::published::published_entries(&packages_dir.join("Published.toml"))?;
+    let deployed = entries
+        .get("testnet")
+        .ok_or_else(|| anyhow::anyhow!("no `testnet` entry in Published.toml"))?
+        .version;
+
+    assert_eq!(
+        declared,
+        deployed + 1,
```
bmwill, line 41: do we want to have this be strictly equal or + 1, or always just +1?

bmwill (approved):

crates/hashi/tests/version_policy_tests.rs
Bridgerz: Do we ever want to be able to skip a version? I guess whats the use in having this?

bmwill: no we should probably avoid skipping

Bridgerz: ok so then I say as is makes sense no?

PR #985, opened 2026-08-08
Bind the rotation complaint response to the complained dealer

zhouwfang:
#95

### DIFF
--- a/crates/hashi/src/mpc/types.rs
+++ b/crates/hashi/src/mpc/types.rs
@@ -746,7 +746,7 @@ impl RotationComplainContext {
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum DealerOutputsKey {
     Dkg(Address),
-    Rotation(ShareIndex),
+    Rotation(Address, ShareIndex),
 }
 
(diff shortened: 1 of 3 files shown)


crates/hashi/src/mpc/types.rs
```diff
@@ -746,7 +746,7 @@ impl RotationComplainContext {
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum DealerOutputsKey {
     Dkg(Address),
-    Rotation(ShareIndex),
+    Rotation(Address, ShareIndex),
```
benr-ml, line 749: nit - it's reduandent since (committee, ShareIndex) -> address

benr-ml (approved):

crates/hashi/src/mpc/types.rs
zhouwfang: @benr-ml True, it's derivable. But the key isn't storing the dealer, it's enforcing it. Index-only means the check lives in a guard elsewhere and has to stay in sync with the cache read. With the dealer in the key a wrong-dealer lookup just misses.

PR #991, opened 2026-08-11
governance: committee-voted ignore/un-ignore of a member

Bridgerz:
The committee can now vote to **ignore** a registered member — treating them as no longer part of the committee — and reverse it with the same proposal type. Motivated by testnet validators that lag on upgrades and stall quorums while still counting toward every denominator.

Based on `main`, which already carries the shared version-framework base (#1026): `SUPPORTED_PACKAGE_VERSIONS = [1, 2]`, `SuiTxExecutor::active_call_package_id` + active-version `start_reconfig` routing, `fetch_initial_shared_version`, and `HashiClient::latest_package_id`. This PR is the base of the governance stack; **#993** (voluntary resignation) stacks on top.

## Semantics
- **Storage**: a `Bool` under key `"ignored"` in `MemberInfo.extra_fields` (the config-typed extension slot), written only by the new quorum-gated (6667 bps) `ignore_member` proposal type. `MemberInfo`'s layout stays frozen; a doc note marks the flag for promotion to a real field at the pre-mainnet testnet wipe.
- **Effect at the next committee formation, nothing mid-epoch**: the sole mechanism is a skip branch in `new_committee_from_validator_set`. Total weight re-sums, so certificates, proposal quorums, MPC t/f, leader rotation, and the guardian's handoff-derived thresholds all re-derive with **zero special-casing** — the member is simply absent from the members vector. The current epoch's bitmap indices / MPC party ids are never disturbed. If a reconfiguration is already in flight, the change lands one formation later.
- **Bounds documented in the module doc**: exclusion (and the reconfig enacting it, whose handoff cert still counts the ignored member in its denominator) requires non-participating weight ≤ 3333 bps — the standard BFT bound. The registry is untouched: an ignored member keeps proposal/vote authorization, and an all-ignored registry degrades to "reconfig aborts, current committee continues" — recoverable by un-ignoring.

## Rust wiring
Full proposal-kind checklist for `IgnoreMember`: `MoveType` mirror at defining version 2, decode/parse arms, `extra_fields` `Bool` decode + `get_bool`, expired-proposal GC, and CLI `proposal create ignore-member --validator <addr> [--unignore]` with effect-timing guidance and a >25%-weight warning, plus committee views showing the flag.

## Linkage handling for the v2 proposal type
Driving a v2-introduced proposal type on current fullnodes needs the linkage-safe transaction construction whose primitives live in `main` (#1026). This PR **applies** them to the governance path:
- Proposal create/execute/vote/remove-vote and the proposal GC build **fully-resolved shared inputs** (initial shared version + explicit mutability) via `fetch_initial_shared_version`, so sui ≥ 1.76 fullnodes never `INVALID_LINKAGE` on the upgrade-introduced `ignore_member` module.
- Calls that carry an upgrade-introduced type argument (`vote` / `remove_vote` / `delete_expired`) route through the **latest** package (`latest_package_id`); `get_proposal_type_arg` now resolves each type's **defining** address via `PackageVersions`, replacing the GC's hand-rolled `TypeTag` map.

## Tests
- Move: new tests in `ignore_member_tests.move` — proposal lifecycle, formation skip via a test-only voting-powers entry point, un-ignore restore, no-op/unregistered aborts, drift idempotence, and the mid-reconfig "next formation" semantics.
- Rust units: extra-fields decode, `get_bool`, proposal-type parse; Move-compat gate stays green (all Move changes are compatible additions).
- e2e `test_ignored_member_excluded_at_epoch_boundary`: snapshot-v1 publish → upgrade → DKG → governance ignore at the v2 address → epoch change → asserts the 3-member committee with re-summed weight → **deposit and withdrawal confirm with the smaller committee** (proving certs, reshare, and guardian all follow).

### DIFF
--- a/packages/hashi/sources/core/committee/committee_set.move
+++ b/packages/hashi/sources/core/committee/committee_set.move
@@ -438,6 +438,10 @@ fun remove_committee(self: &mut CommitteeSet, epoch: u64): Committee {
     self.committees.remove(epoch)
 }
 
+/// Thin wrapper extracting the epoch and voting powers from the system
+/// object. Split from `new_committee_from_voting_powers` because unit tests
+/// cannot construct a `SuiSystemState`: the pure inner function is what lets
+/// formation (including the ignore filtering) be unit-tested at all.
 fun new_committee_from_validator_set(
     self: &CommitteeSet,
     sui_system: &sui_system::sui_system::SuiSystemState,
--- a/packages/hashi/sources/core/committee/committee_set.move
+++ b/packages/hashi/sources/core/committee/committee_set.move
@@ -27,9 +27,13 @@ use sui::{
 /// before this key existed need no migration.
 ///
 /// The flag lives in `extra_fields` only because `MemberInfo`'s layout is
-/// frozen on the deployed network; when testnet is wiped before mainnet,
-/// promote it to a real `ignored: bool` field on `MemberInfo` and delete
-/// this key.
+/// frozen on the deployed network.
+///
+/// TODO(pre-mainnet-wipe): when testnet is wiped and the Move versions are
+/// squashed before mainnet, promote this to a root-level `ignored: bool`
+/// field on `MemberInfo` and delete this key (and its Rust mirror
+/// `MEMBER_IGNORED_KEY` / the `extra_fields` decode in
+/// `convert_move_member_info`).
 const MEMBER_IGNORED_KEY: vector<u8> = b"ignored";
 
 // ~~~~~~~ Errors ~~~~~~~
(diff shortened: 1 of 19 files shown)


packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -390,8 +444,25 @@ fun new_committee_from_validator_set(
...
     ctx: &TxContext,
 ): Committee {
-    let epoch = ctx.epoch();
-    let mut validator_set = sui_system.active_validator_voting_powers();
+    self.new_committee_from_voting_powers(
+        ctx.epoch(),
+        sui_system.active_validator_voting_powers(),
+        config,
+    )
+}
```
bmwill, line 456: What is the purpose of splitting this function?

bmwill (approved):

packages/hashi/sources/core/committee/committee_set.move
Bridgerz: Testability: Move unit tests can't construct a SuiSystemState, so the outer function stays a thin wrapper over the system object while new_committee_from_voting_powers is a pure function over the voting-power map that ignore_member_tests drives directly. That's what lets the ignore-filtering be unit-tested at all. Happy to add a doc comment saying so.

PR #993, opened 2026-08-11
governance: voluntary committee resignation

Bridgerz:
Stacked on #991.

A member (validator or delegated operator key) can **voluntarily resign** from the committee via `hashi validator resign`, effective at the epoch boundary, with the registration removed atomically at the transition — a clean exit; re-joining requires full re-registration.

## Semantics
- **While serving** (current committee, or a pending committee mid-reconfiguration): only a revocable flag is set (`"resigned"` Bool in `MemberInfo.extra_fields`, same mechanism and pre-mainnet-wipe TODO as the ignore flag). The member keeps every current-epoch duty. The next committee formation skips them; `finalize_resignations` — called from `end_reconfig`, deliberately **assert-free** so it can never block the transition — enumerates the *outgoing committee* (sidestepping the non-iterable members Bag) and deletes the `MemberInfo` of resigned members absent from the new committee.
- **Duty-free members** (never included, skipped, or pre-genesis) are removed immediately.
- **`abort_reconfig` sweeps pending-only resigned members** — without it, a mid-reconfig joiner who resigns and then sees an abort would leak their registration forever.
- **Best-effort last-active-member guard** (`ELastActiveMember`); the hard backstop stays `new_committee`'s non-empty assert (current committee keeps operating; healable).
- Revocable with `hashi validator withdraw-resignation` until consumed; resigned+ignored coexist with removal keyed off resigned only.

## The resignation latch (node-side, load-bearing)
A running node re-registers itself on every Sui epoch trigger and immediately after `end_reconfig` — without a latch, a resigned-and-removed validator **reverses its own exit within one trigger**. `prepare_and_register_keys` (the single choke point behind all four auto-registration paths, including key-lost rejoin) now refuses to register or upload next-epoch keys while the on-chain registration is flagged resigned, or after removal following an observed resignation — tracked by a marker file (`resigned.marker`) in the db directory so it survives the removal and restarts, and self-clearing once the chain shows the member registered and unflagged again (withdrawal, or explicit `hashi register`).

## Completing the active-version reconfig routing
The active-version helper (`active_call_package_id`) and the `start_reconfig` routing landed earlier in #1026 (now in main). The e2e caught the *rest* of the triplet still pinned to the original package: `end_reconfig` ran v1's code and silently skipped `finalize_resignations`. This PR routes the remaining two calls — `end_reconfig` and `submit_committee_handoff` — through the active version as well, so the whole reconfiguration triplet (`start_reconfig` / `submit_committee_handoff` / `end_reconfig`) executes one version's semantics. In-transaction cert constructors ride the same active package (mixed versions in one PTB are a linkage conflict).

The new `resign` / `withdraw_resignation` CLI transaction builders reuse the shared-input and package-resolution helpers from #1026/main (`latest_package_id`, `fetch_initial_shared_version` / `hashi_initial_shared_version`, `resolve_validator_address`) — the `resign` entry function is v2-only, so the call targets the latest package.

## Tests
- Move (17 new): flag+formation-skip, immediate removal (duty-free and pre-genesis), withdrawal paths, authorization matrix (operator key works; stranger aborts), last-member guard, finalize at `end_reconfig`, mid-reconfig resign surviving exactly one boundary, the abort-sweep leak test and its retain counterpart, resigned+ignored coexistence, ex-member vote rejection, and the approved-ignore-vs-deregistered-target interaction.
- Rust: mirrors (`resigned` field on `MemberInfo`; three v2 `validator` events — `ValidatorResigned`, `ValidatorResignationWithdrawn`, `ValidatorDeregistered`, all at defining version 2), plus the resignation-latch unit coverage and CLI display (`list-members` / `view-member` show a Resigned column).
- e2e `test_resigned_member_removed_at_epoch_boundary`: resign via the node's own executor → first boundary excludes + deregisters → **second boundary with the resigned node still running proves the latch** (no reappearance) → explicit re-registration re-admits the member. Reconfig-heavy neighbors (snapshot upgrade, ignore e2e) re-verified green on the triplet routing change.

### DIFF
--- a/packages/hashi/sources/core/committee/committee_set.move
+++ b/packages/hashi/sources/core/committee/committee_set.move
@@ -36,10 +36,26 @@ use sui::{
 /// `convert_move_member_info`).
 const MEMBER_IGNORED_KEY: vector<u8> = b"ignored";
 
+/// `MemberInfo.extra_fields` key holding the voluntary "resigned" flag as a
+/// `Bool` value. Set by `request_resignation`, cleared by
+/// `clear_resignation`, consumed by committee formation (skip) and by
+/// `finalize_resignations` (registry removal at the epoch transition).
+///
+/// TODO(pre-mainnet-wipe): promote to a root-level `resigned: bool` field on
+/// `MemberInfo` together with `MEMBER_IGNORED_KEY` above.
+const MEMBER_RESIGNED_KEY: vector<u8> = b"resigned";
+
 // ~~~~~~~ Errors ~~~~~~~
 
 #[error(code = 0)]
 const EMemberNotRegistered: vector<u8> = b"No member is registered under this validator address";
+#[error(code = 1)]
+const EAlreadyResigned: vector<u8> = b"Member has already requested resignation";
+#[error(code = 2)]
+const ENotResigned: vector<u8> = b"Member has no pending resignation to withdraw";
+#[error(code = 3)]
+const ELastActiveMember: vector<u8> =
+    b"Cannot resign as the last active committee member; the committee would be unable to form";
 
 // ~~~~~~~ Structs ~~~~~~~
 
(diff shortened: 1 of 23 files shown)


packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -286,6 +302,110 @@ public(package) fun is_member_ignored(self: &CommitteeSet, validator_address: ad
...
+    assert!(member.is_resigned(), ENotResigned);
+    self
+        .member_mut(validator_address)
+        .extra_fields
+        .upsert(MEMBER_RESIGNED_KEY, config_value::new_bool(false));
+}
+
+/// Remove resigned ex-members' registrations, atomically with the epoch
+/// transition that stopped including them. Called from
+/// `reconfig::end_reconfig` AFTER the new epoch is active; enumerates the
```
bmwill: I think we should rework this so that end_reconfig doesn't need to worry about this. Having this adds more risk to end_reconfig failing in some unexpected way.

Instead we should have a permissionless mechanism that allows use to remove a Member's info when:
- They are not part of the active validator set and they are not part of the Sui Validator set
- They are not part of the active validator set and they have voluntarily resigned 

We explicitly cannot remove the info for a validator who is not part of the active set and they have been explicitly ignored via governance (lest they are able to trivially rejoin without the committee permitting them)

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -286,6 +302,110 @@ public(package) fun is_member_ignored(self: &CommitteeSet, validator_address: ad
...
+        self.assert_not_last_active_member(validator_address);
+        self
+            .member_mut(validator_address)
+            .extra_fields
+            .upsert(MEMBER_RESIGNED_KEY, config_value::new_bool(true));
+        false
+    } else {
+        self.remove_member(validator_address);
+        true
+    }
```
bmwill: See comment below, lets just have this add the field but not actually do the remove_member call, that can be a separate entry point.

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -354,7 +474,10 @@ public(package) fun end_reconfig(
     (next_epoch, committee_handoff_cert)
 }
 
-public(package) fun abort_reconfig(self: &mut CommitteeSet, _ctx: &TxContext): u64 {
+public(package) fun abort_reconfig(
+    self: &mut CommitteeSet,
+    _ctx: &TxContext,
+): (u64, vector<address>) {
```
bmwill: As indicated in my comment above that change would allow us to leave the reconfig flow identical.

packages/hashi/sources/core/committee/committee_set.move
Bridgerz: Done in 4d6cf5d16 + b2c8fc3f8. end_reconfig and abort_reconfig are reverted to their previous shapes and never touch the registry. The permissionless `validator::remove_inactive_member` implements your eligibility rules exactly: outside the current and pending committee, then removable when resigned or no longer in Sui's active validator set, and never while governance-ignored (removal would delete the flag with the registration, letting the member shed the exclusion by re-registering). One extra benefit: registry cleanup no longer depends on which package version executes the epoch transition, which removes a cross-version hazard we flagged in the upgrade-window analysis.

packages/hashi/sources/core/committee/committee_set.move
Bridgerz: Done, resign is flag-only in all cases now (duty-free members included); removal only ever happens through the separate permissionless entry.

packages/hashi/sources/core/committee/committee_set.move
Bridgerz: Done, the reconfig flow is back to being identical to the base branch; the abort-time sweep is gone too.

bmwill (approved):

PR #998, opened 2026-08-11
[Guardian] [4/4] Add audit follow-up cleanups

mskd12:
Stacked on #997.

## Summary
- widen random failure-log key suffixes from 32 to 128 bits
- centralize expected-session validation for KP-signed requests and reject certificate rotation at the wrong lifecycle before signature verification
- document the heartbeat fencing assumption around enclave suspension and resume
- record a conditional state-only RPC optimization if Guardian attestation generation becomes operationally problematic
- colocate the genesis bootstrap guard with provisioner initialization and remove a stale retention-policy TODO

This draft intentionally collects small end-to-end audit follow-ups.

## Local checks
- `make fmt`
- `cargo check`


crates/hashi-guardian/src/lib.rs
```diff
@@ -13,7 +13,12 @@ pub const LIVE_SESSION_LATEST_HEARTBEAT_MAX_AGE: Duration = Duration::from_mins(
 pub const OTHER_SESSION_QUIET_PERIOD: Duration = Duration::from_mins(10);
 
 // An enclave that cannot write its next heartbeat must abort before another
-// session is allowed to treat it as quiet.
+// session is allowed to treat it as quiet. This fencing argument assumes the
+// old enclave either keeps executing until that abort or is permanently
+// terminated. Nitro Enclaves currently have no suspend/resume lifecycle (and
+// stopping the parent terminates its enclaves); a future platform that can
```
benr-ml, line 19: it could be safer if an enclave would (1) check the duration between its succesful writes and fail if too long, and (2) block other requests until the write returns (follow up pr)

benr-ml (approved):

crates/hashi-guardian/src/lib.rs
mskd12: Sounds good. Will do in a follow up!

PR #757, opened 2026-07-06
hashi: Tighten public visibility to keep signatures evolvable

Bridgerz:
## Motivation

`public fun` signatures freeze permanently at first mainnet publish; `entry` and `public(package)` do not. This reclassifies functions that don't need `public` so their signatures stay evolvable (e.g. a future chain can add a `protocol_id` parameter to validator registration or a governance action).

## Changes

**→ `entry`** (user/validator actions returning `()`; still PTB-callable by name, so off-chain is unaffected):
- `validator`: `register` + the five `update_*` setters
- `deposit`: `deposit`, `delete_expired_deposit`
- `proposal`: `vote`, `remove_vote`

**→ `public(package)`** (intra-package only, no off-chain/PTB callers):
- `tob`: `tob_key`, `protocol_type_dkg`/`_key_rotation`/`_nonce_generation`, `epoch`
- `proposal`: `votes`, `is_expired`, `quorum_reached`
- `utxo`: `amount`, `id`, `derivation_path`
- `withdrawal_queue`: `is_approved` (currently unused)

**Kept `public`** where the frozen signature is unavoidable: PTB-input constructors (`config_value::new_*`, `utxo`/`utxo_id`, `output_utxo`, `new_committee_signature`), the `Balance`-passing withdrawal entrypoints (`request_withdrawal`/`cancel_withdrawal`), the upgrade hot-potato returns (`execute`/`finalize_upgrade`), and the `propose`/`execute` governance API (returns `ID`/hot potato).

## Test plan

- [x] 141/141 Move tests pass
- [x] `cargo check --workspace` clean
- [x] prettier-move clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1001, opened 2026-08-11
move: TOB cert bucket GC entries with per-protocol floors

Bridgerz:
## Problem

`Hashi.tob` grows without bound: per committee epoch it gains one key-generation bucket plus one nonce bucket per presig batch, each a Bag field + LinkedTable + ~78 dealer nodes. Nothing has ever removed them — `destroy_all_certs` has been uncallable since #757 demoted the `ProtocolType` constructors to `public(package)` (a Move enum cannot be a PTB pure argument), and no off-chain caller ever existed. The bag is also paged in full on every node bootstrap (`scrape_tob_entries`), so growth is a startup cost too.

## Change

Two new primitive-argument, permissionless GC entries mirroring the `submit_*` split, gated (per the GC convention) only on the version check:

- **`destroy_nonce_certs(epoch, batch_index)`** — floor `current >= epoch + 2`. Nonce buckets are only ever read during their own epoch (presigs are strictly single-epoch; local nonce state is flat-pruned at the boundary), so +2 has a full epoch of margin.
- **`destroy_key_gen_certs(epoch)`** — probes both the Dkg and KeyRotation keys, so callers never need to know whether an epoch was genesis or a rotation. Two floors, both required:
  - `current >= epoch + 8`: node DBs retain dealer/rotation messages for the trailing 7 epochs specifically so break-glass key recovery (`internal-tools/key_recovery`) can pair them with the on-chain bucket. Destruction is permissionless, so this window must be enforced on-chain — driver policy alone cannot protect it.
  - `committee_set::is_before_previous_committee(epoch)`: committee epochs are Sui epochs (`ctx.epoch()`) and can gap, so the previous committee can sit at `current − k`, k ≥ 2. Its key-generation bucket seeds the next rotation (`fetch_key_generation_certificates(previous_epoch)` hard-errors if absent), so a bare age floor would let anyone brick the next reconfig. The guard requires a committee epoch strictly between the bucket and now; the walk is bounded by the current↔previous gap.

Both entries no-op on an eligible-but-absent bucket (batched GC transactions and permissionless racers cannot poison each other) but assert the floors unconditionally, so a premature call can never silently succeed. Removal probes by **stored type** (`contains_with_type<TobKey, EpochCertsV1>`), leaving any future bucket layout in place for the version that understands it — the extension point the stamped-cert work (#953) needs, avoiding the bare-bucket-stranding hazard found in its review.

`tob::destroy_all`'s `+2` assert stays as defense-in-depth; the dead `destroy_all_certs` entry is left untouched.

## Tests

18 new tests in `tob_pruning_tests.move`: floor boundaries (exact and one-below, with and without the bucket present), the gap scenario (committees {2,10} — retention passes, committee guard aborts), pending-reconfig cases (pending committee neither satisfies the guard nor is destroyable), both-buckets destruction + idempotency, foreign-typed bucket left in place, callability while paused and mid-reconfig, populated-bucket drain, and direct walk checks.

The entries ride v2 (`PACKAGE_VERSION = 2`) and gate on `versioning().assert_version_enabled()`; this PR adds **no** version-framework plumbing (that lives in #1026, now merged into main) and bumps nothing — `version_policy_tests` and `move_upgrade_compat` pass unchanged.

## Stacking

Based on `main` (the version-framework enablement + shared linkage helpers were extracted to #1026 and merged). This is the base of the TOB pruning stack; the Rust driver that feeds these entries through the leader GC loop is #1002, stacked directly on this PR.

### DIFF
--- a/packages/hashi/sources/core/cert_submission.move
+++ b/packages/hashi/sources/core/cert_submission.move
@@ -5,6 +5,31 @@ module hashi::cert_submission;
 
 use hashi::{committee::CommitteeSignature, hashi::Hashi, tob::ProtocolType};
 
+// ~~~~~~~ Constants ~~~~~~~
+
+/// Key-generation cert buckets for the last N committee epochs are retained
+/// for break-glass key recovery: the `key_recovery` internal tool pairs a
+/// node's local DB — which keeps dealer/rotation messages for the trailing 7
+/// epochs — with the on-chain bucket of the epoch being reconstructed.
+/// Enforced on-chain because destruction is permissionless.
+const KEY_GEN_CERT_RETENTION_EPOCHS: u64 = 8;
+
+/// Nonce cert buckets are only ever read during their own epoch; +2 mirrors
+/// the `tob::destroy_all` backstop.
+const NONCE_CERT_MIN_AGE_EPOCHS: u64 = 2;
+
+// ~~~~~~~ Errors ~~~~~~~
+
+#[error]
+const ETooEarlyToDestroyNonceCerts: vector<u8> =
+    b"Nonce cert buckets may only be destroyed two epochs after their epoch";
+#[error]
+const ETooEarlyToDestroyKeyGenCerts: vector<u8> =
+    b"Key-generation cert buckets are retained for break-glass key recovery";
+#[error]
+const EKeyGenCertsStillNeeded: vector<u8> =
+    b"The previous committee's key-generation certs seed the next rotation";
+
 // ~~~~~~~ Entry Functions ~~~~~~~
 
 entry fun submit_dkg_cert(
@@ -71,7 +96,71 @@ entry fun destroy_all_certs(
     };
 }
 
+/// Destroy the key-generation (DKG or rotation) cert buckets of `epoch`.
+/// Garbage collection: permissionless and deliberately NOT gated on
+/// pause/reconfig — see `destroy_all_certs`.
+///
+/// A key-generation bucket stays live longer than its certs' epoch: the NEXT
+/// rotation reads the PREVIOUS committee's bucket to seed the handoff, and
+/// committee epochs can gap, so an age floor alone cannot identify the
+/// previous committee's bucket. Both floors are asserted unconditionally
+/// (premature calls abort even when the bucket is absent); an
+/// eligible-but-absent bucket is a no-op so batched GC transactions and
+/// permissionless racers cannot poison each other.
+entry fun destroy_key_gen_certs(hashi: &mut Hashi, epoch: u64) {
+    hashi.versioning().assert_version_enabled();
+    let current_epoch = hashi.committee_set().epoch();
+    assert!(current_epoch >= epoch + KEY_GEN_CERT_RETENTION_EPOCHS, ETooEarlyToDestroyKeyGenCerts);
+    assert!(hashi.committee_set().is_before_previous_committee(epoch), EKeyGenCertsStillNeeded);
+    // One entry covers both key-generation protocols: callers never need to
+    // know whether `epoch` was a genesis (DKG) or rotation epoch.
+    destroy_bucket_if_present(
+        hashi,
+        hashi::tob::tob_key(epoch, option::none(), hashi::tob::protocol_type_dkg()),
+        current_epoch,
+    );
+    destroy_bucket_if_present(
+        hashi,
+        hashi::tob::tob_key(epoch, option::none(), hashi::tob::protocol_type_key_rotation()),
+        current_epoch,
+    );
+}
+
+/// Destroy the nonce-generation cert bucket of `(epoch, batch_index)`.
+/// Garbage collection: permissionless and deliberately NOT gated on
+/// pause/reconfig — see `destroy_all_certs`. Nonce buckets are only ever
+/// read during their own epoch, so no committee-awareness is needed. The
+/// floor is asserted unconditionally; an eligible-but-absent bucket is a
+/// no-op (see `destroy_key_gen_certs`).
+entry fun destroy_nonce_certs(hashi: &mut Hashi, epoch: u64, batch_index: u32) {
+    hashi.versioning().assert_version_enabled();
+    let current_epoch = hashi.committee_set().epoch();
+    assert!(current_epoch >= epoch + NONCE_CERT_MIN_AGE_EPOCHS, ETooEarlyToDestroyNonceCerts);
+    destroy_bucket_if_present(
+        hashi,
+        hashi::tob::tob_key(
+            epoch,
+            option::some(batch_index),
+            hashi::tob::protocol_type_nonce_generation(),
+        ),
+        current_epoch,
+    );
+}
+
 // ~~~~~~~ Private Functions ~~~~~~~
+
+/// Remove and destroy the bucket at `key` if one is stored as an
+/// `EpochCertsV1`; otherwise do nothing. Probing by STORED type (not by the
+/// requested protocol) means a bucket written under a future value layout is
+/// left in place for the module version that understands it, instead of
+/// aborting the whole transaction or being destroyed as the wrong type.
+fun destroy_bucket_if_present(hashi: &mut Hashi, key: hashi::tob::TobKey, current_epoch: u64) {
+    let tob = hashi.tob_mut();
+    if (tob.contains_with_type<hashi::tob::TobKey, hashi::tob::EpochCertsV1>(key)) {
+        let epoch_certs: hashi::tob::EpochCertsV1 = tob.remove(key);
+        hashi::tob::destroy_all(epoch_certs, current_epoch);
+    }
+}
 
 fun submit_cert_internal(
     hashi: &mut Hashi,
--- a/packages/hashi/sources/core/cert_submission.move
+++ b/packages/hashi/sources/core/cert_submission.move
@@ -8,11 +8,17 @@ use hashi::{committee::CommitteeSignature, hashi::Hashi, tob::ProtocolType};
 // ~~~~~~~ Constants ~~~~~~~
 
 /// Key-generation cert buckets stay on-chain until their Sui epoch number is
-/// at least 8 below the current one. This matches the node DB's numeric
-/// retention window: dealer/rotation messages at `current - 7` and newer are
-/// retained for break-glass key recovery. This is an epoch-number distance,
-/// not a count of finalized committee generations (committee epochs can gap).
-/// Enforced on-chain because destruction is permissionless.
+/// at least 8 below the current one. This strictly exceeds the node DB's
+/// retention window (`RETENTION_EXTRA_EPOCHS = 7` in `db.rs`, anchored to the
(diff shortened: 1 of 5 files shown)


jessiemongeon1:
## Style Guide Audit

packages/hashi/sources/core/cert_submission.move
```diff
@@ -10,6 +10,31 @@ module hashi::cert_submission;
 
 use hashi::{committee::CommitteeSignature, hashi::Hashi, tob::ProtocolType};
 
+// ~~~~~~~ Constants ~~~~~~~
+
+/// Key-generation cert buckets for the last N committee epochs are retained
```
zhouwfang: This reads as generations ("the last N committee epochs") but :111 measures Sui-epoch distance, and under a gap those diverge badly.

packages/hashi/sources/core/cert_submission.move
```diff
@@ -69,8 +94,72 @@ entry fun destroy_all_certs(
     hashi::tob::destroy_all(epoch_certs, current_epoch);
 }
 
+/// Destroy the key-generation (DKG or rotation) cert buckets of `epoch`.
+/// Garbage collection: permissionless and deliberately NOT gated on
+/// pause/reconfig — see `destroy_all_certs`.
```
zhouwfang, line 109: Why not delete destroy_all_certs? It only enforces tob::destroy_all's +2, so it bypasses both key-gen floors.

packages/hashi/sources/core/cert_submission.move
```diff
@@ -69,8 +94,72 @@ entry fun destroy_all_certs(
...
+/// committee epochs can gap, so an age floor alone cannot identify the
+/// previous committee's bucket. Both floors are asserted unconditionally
+/// (premature calls abort even when the bucket is absent); an
+/// eligible-but-absent bucket is a no-op so batched GC transactions and
+/// permissionless racers cannot poison each other.
+entry fun destroy_key_gen_certs(hashi: &mut Hashi, epoch: u64) {
+    hashi.versioning().assert_version_enabled();
+    let current_epoch = hashi.committee_set().epoch();
+    assert!(current_epoch >= epoch + KEY_GEN_CERT_RETENTION_EPOCHS, ETooEarlyToDestroyKeyGenCerts);
+    assert!(hashi.committee_set().is_before_previous_committee(epoch), EKeyGenCertsStillNeeded);
```
zhouwfang, line 122: This fires whenever no committee sits in (epoch, current), including when there's no previous committee at all. Nothing is misprotected, the message just points at the wrong reason.

packages/hashi/sources/core/cert_submission.move
```diff
@@ -69,8 +94,72 @@ entry fun destroy_all_certs(
...
 // ~~~~~~~ Private Functions ~~~~~~~
 
+/// Remove and destroy the bucket at `key` if one is stored as an
+/// `EpochCertsV1`; otherwise do nothing. Probing by STORED type (not by the
+/// requested protocol) means a bucket written under a future value layout is
+/// left in place for the module version that understands it, instead of
+/// aborting the whole transaction or being destroyed as the wrong type.
+fun destroy_bucket_if_present(hashi: &mut Hashi, key: hashi::tob::TobKey, current_epoch: u64) {
+    let tob = hashi.tob_mut();
+    if (tob.contains_with_type<hashi::tob::TobKey, hashi::tob::EpochCertsV1>(key)) {
```
zhouwfang: The gap is the case where it declines: once v2 writes stamped nonce buckets, a bare-probing destroy skips every one of them, and nothing surfaces it since the entry succeeds and the sweep logs success. Worth making that case loud, abort here or a signal in the driver.

Bridgerz: Fixed in cb3834bc5. Nonce GC now explicitly handles both bare EpochCertsV1 and stamped StampedEpochCertsV1 buckets, and aborts with EUnsupportedCertBucketLayout when a present bucket has an unknown layout instead of reporting false success. Added stamped_nonce_bucket_is_destroyed and destroy_rejects_unknown_bucket_layout.

packages/hashi/sources/core/cert_submission.move
Bridgerz: Fixed in cb3834bc5. Upgrade compatibility requires retaining the legacy ABI, so destroy_all_certs now delegates to destroy_nonce_certs and destroy_key_gen_certs and cannot bypass the +8 key-generation floor. Added legacy_destroy_all_obeys_keygen_retention_floor.

packages/hashi/sources/core/cert_submission.move
Bridgerz: Fixed in cb3834bc5. The error is now EKeyGenCertsStillNeeded and says to require a later finalized committee before pruning, covering both no previous committee and no committee in the open epoch interval. The previous, pending, and gap tests cover those cases.

packages/hashi/sources/core/cert_submission.move
Bridgerz: Fixed in cb3834bc5. The comment now states that retention is numeric Sui-epoch distance, not finalized committee generations, and calls out that committee epochs can gap. The gap regression remains covered by keygen_gap_previous_committee_protected.

packages/hashi/sources/core/cert_submission.move
```diff
@@ -49,29 +78,121 @@ entry fun submit_nonce_cert(
...
+/// pause/reconfig — see `destroy_all_certs`.
+///
+/// A key-generation bucket stays live longer than its certs' epoch: the NEXT
+/// rotation reads the PREVIOUS committee's bucket to seed the handoff, and
+/// committee epochs can gap, so an age floor alone cannot identify the
+/// previous committee's bucket. Both floors are asserted unconditionally
+/// (premature calls abort even when the bucket is absent); an
+/// eligible-but-absent bucket is a no-op so batched GC transactions and
+/// permissionless racers cannot poison each other.
+entry fun destroy_key_gen_certs(hashi: &mut Hashi, epoch: u64) {
```
zhouwfang, line 118: Worth flagging that this is the first time these buckets can be destroyed at all. The thing I'd want a decision on is that it removes the on-chain half of break-glass recovery, anyone can call it, and it stays callable while paused. The retention floor assumes the node already dropped its own copy, but that pruning is best effort. If a node still has it, someone else can make that epoch unrecoverable, and there is no way to put the bucket back.

packages/hashi/sources/core/cert_submission.move
zhouwfang: I don't think compatibility actually forces this. Private entry functions aren't linkage checked, and we've already relied on that when submit_nonce_cert picked up a &Clock param. Fine to keep it either way, but the comment states it as a rule and someone will cite it.

packages/hashi/sources/hashi.move
```diff
@@ -45,7 +45,7 @@ public struct Hashi has key {
     versioning: Versioning,
     treasury: Treasury,
     proposals: Proposals,
-    /// TOB certificates by (epoch, batch_index) -> EpochCertsV1
+    /// TOB certificates by (epoch, batch_index, protocol_type) -> EpochCertsV1
```
zhouwfang: The value half of this is still stale, since the bag holds both layouts now. Same comment in the Rust mirror has the same problem.

packages/hashi/sources/core/cert_submission.move
zhouwfang: I think this still reads wrong. It only fires once epoch is at or after the previous committee, so by then there is always a later finalized committee. What you actually need is for it to be older than the previous committee.

packages/hashi/sources/core/cert_submission.move
Bridgerz: @bmwill WDYT?

Bridgerz: "No way to put the bucket back" is only true of live state. Every cert field enters as caller-supplied transaction input, so the full contents persist in Sui transaction/checkpoint history and are in theory re-derivable from an archival node. Destruction deletes a cache, not the only copy.

packages/hashi/sources/core/cert_submission.move
Bridgerz: Extended further in 1dc9ddcd2 — resolving, reopen if it still reads wrong.

packages/hashi/sources/core/cert_submission.move
Bridgerz: Fair point — linkage does not force retention (the `submit_nonce_cert` `&Clock` change is the precedent). Reworded in 1dc9ddcd2: framed as defense in depth, kept so any historical caller path routes through the real floors.

packages/hashi/sources/core/cert_submission.move
Bridgerz: Agreed — the message pointed at the wrong condition. Reworded in 1dc9ddcd2: buckets must be strictly older than the previous committee, whose bucket seeds the next rotation.

packages/hashi/sources/core/cert_submission.move
Bridgerz: Resolving — stamped handling + `EUnsupportedCertBucketLayout` landed with tests.

packages/hashi/sources/core/cert_submission.move
Bridgerz: The rationale is now recorded in code: 1dc9ddcd2 (`KEY_GEN_CERT_RETENTION_EPOCHS` doc — certs-vs-payloads + the db.rs coupling) and b50c1aa6e (`RETENTION_EXTRA_EPOCHS` mirror note).

packages/hashi/sources/hashi.move
Bridgerz: Fixed in 1dc9ddcd2 (Move) and b50c1aa6e (Rust mirror) — both docs now name the bare and stamped layouts.

packages/hashi/sources/core/cert_submission.move
zhouwfang: You're right for the key generation certs, but it is not right for every cert field. Stamped nonce certs carry a Clock timestamp instead, and this PR destroys those buckets too. I'd also avoid "cache", because nothing can put the bucket back, and the node reads it live during rotation and hard errors when it's gone.

Bridgerz: Both corrections taken. The stamped timestamp is a Clock read, not a tx input (though it equals the enclosing checkpoint's timestamp, so it is still derivable from history), and "cache" was the wrong word for something with no refill path. But the argument I should have led with is about readers, not reconstruction: nonce buckets are only ever read during their own epoch (every waiter is superseded when the chain's epoch advances), so a +2-destroyed nonce bucket has no reader left to hard-error. And the rotation read you're describing targets the previous committee's keygen bucket, which `is_before_previous_committee` protects unconditionally at any age, so these entries cannot destroy it. Reconstructibility was only ever the last-resort footnote for keygen certs, where it stands.

zhouwfang (approved):

PR #1002, opened 2026-08-11
rust: leader GC job pruning dead TOB cert buckets

Bridgerz:
Stacked on #1001 (base: `main` → #1001 → this), whose Move entries this drives. Closes the loop on #1001: the on-chain destroy entries existed but nothing called them.

The upgrade-safe transaction plumbing this job depends on — `SuiTxExecutor::active_call_package_id`, `cli::client::fetch_initial_shared_version`, and `SUPPORTED_PACKAGE_VERSIONS = [1, 2]` — now lives in `main` (extracted to #1026, merged), so it is no longer part of this PR; this PR only consumes it.

## Change

A fourth singleton GC job in the leader's per-checkpoint loop, alongside deposits / proposals / UTXO cleanup:

- **`OnchainState::list_tob_keys`** — pages the TOB bag decoding dynamic-field **names only**, so bucket value layouts this binary cannot decode (e.g. a future `StampedEpochCertsV1`) never fail the sweep; undecodable names are skipped with a warning (with `field_id` in the read mask so the warning is actionable). Type safety of the actual removal lives in the Move entries' stored-type probe.
- **`find_tob_buckets_to_prune`** — pure selection predicate mirroring the Move floors (which stay authoritative; drift here only produces aborting transactions, never unsafe deletes). Nonce at `current ≥ epoch+2`; keygen at `current ≥ epoch+8` AND a committee epoch strictly between bucket and now (gap-safe, pending-committee-safe). Unknown key shapes are never selected; saturating arithmetic throughout since field names are attacker-shapeable. Epochs drain oldest-first, nonce batches descending by index so cap truncation always leaves a contiguous `0..m` prefix.
- **Scheduling** — epoch-gated: the gate only closes on a sweep that fully drained (selected under the 200-target cap), so backlogs re-sweep every checkpoint until empty; gated on `active_package_version() >= 2` (the entries are v2-introduced — on a v1-active chain the job stays dormant by construction); unbounded retry with backoff like the UTXO job. Mirror staleness is strictly conservative: a lagging epoch shrinks both floors, a missing committee only disables keygen selection.
- **`execute_destroy_tob_certs`** — one move call per bucket (scalar pures only), routed to the active package with a pre-resolved Hashi shared input (the INVALID_LINKAGE class documented on `delete_expired_proposals`: sui ≥ 1.76 fullnodes fail simulation when inferring an unresolved shared input's mutability requires inspecting an upgraded package's module). Chunking is **adaptive**: 10 calls per transaction, halving any failing chunk down to singletons. Bucket sizes are committee-sized and unknowable from a name-only listing, and destroying one deletes ~dealers+2 object ids against Sui's 2048 deleted-ids-per-tx cap — a fixed chunk would deterministically wedge the sweep behind an unbounded identical retry once committees pass ~100 members (found in adversarial review; mainnet-realistic).

## Tests

- 12 predicate unit tests: floors (exact boundaries both sides), the gap scenario, pending committees, ordering, contiguous-prefix truncation, Dkg+KeyRotation dedup, malformed shapes, absurd-epoch non-panic.
- e2e (extends the resignation flow, which already crosses two epoch boundaries): asserts a below-floor destroy is rejected at `initial+1`, then that the leader **actually prunes** the initial epoch's nonce buckets on-chain after the second boundary while the key-generation bucket survives the retention floor. Verified locally (88s).
- Full `hashi` suite 643/643 (gpg-gated test excluded), clippy clean, `version_policy_tests` untouched.

### DIFF
--- a/crates/hashi/src/leader/garbage_collection.rs
+++ b/crates/hashi/src/leader/garbage_collection.rs
@@ -4,6 +4,8 @@
 //! Garbage collection for expired on-chain data.
 
 use super::LeaderService;
+use crate::onchain::TobKey;
+use crate::onchain::TobPruneTarget;
 use crate::onchain::types::DepositRequest;
 use crate::onchain::types::Proposal;
 use crate::onchain::types::ProposalType;
@@ -12,6 +14,7 @@ use crate::onchain::types::UtxoRecord;
 use crate::sui_tx_executor::ArchiveVictim;
 use crate::sui_tx_executor::SuiTxExecutor;
 use std::collections::BTreeMap;
+use std::collections::BTreeSet;
 use std::sync::Arc;
 use std::time::Duration;
 use sui_sdk_types::Address;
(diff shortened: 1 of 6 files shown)


crates/hashi/src/leader/garbage_collection.rs
```diff
@@ -377,6 +438,95 @@ impl LeaderService {
...
+            debug!("No TOB cert buckets eligible for pruning");
+            return Ok(TobPruneOutcome {
+                swept_epoch,
+                drained: true,
+            });
+        }
+
+        // Whether more work may remain past the cap, decided before the
+        // destroy so a partially-failed batch also re-runs via the retry path.
+        let drained = targets.len() < MAX_TOB_PRUNES_PER_GC;
```
zhouwfang, line 682: The empty-target path already reports drained, so this is a second, weaker way to reach the same conclusion, inferred from one listing coming in under the cap.

crates/hashi/src/leader/garbage_collection.rs
```diff
@@ -428,6 +578,94 @@ fn find_spent_utxos_pending_cleanup(utxo_records: &BTreeMap<UtxoId, UtxoRecord>)
...
+/// Pure-data core, extracted for unit testing like
+/// [`find_spent_utxos_pending_cleanup`].
+fn find_tob_buckets_to_prune(
+    keys: &[TobKey],
+    committee_epochs: &BTreeSet<u64>,
+    current_epoch: u64,
+) -> Vec<TobPruneTarget> {
+    use hashi_types::move_types::ProtocolType;
+
+    // Saturating arithmetic throughout: key epochs come from on-chain field
```
zhouwfang: Only committee members or their operators can insert into the bag, and only for the current or pending epoch, so "anyone can shape" overstates it.

crates/hashi/src/sui_tx_executor.rs
```diff
@@ -1835,6 +1835,121 @@ impl SuiTxExecutor {
...
+
+        let mut queue: std::collections::VecDeque<&[crate::onchain::TobPruneTarget]> =
+            targets.chunks(Self::TOB_DESTROYS_PER_TX).collect();
+        while let Some(chunk) = queue.pop_front() {
+            match self
+                .destroy_tob_chunk(chunk, call_package_id, hashi_initial_shared_version)
+                .await
+            {
+                Ok(()) => {}
+                Err(e) if chunk.len() > 1 => {
```
zhouwfang: Splitting is the response to every error here, not just size, so a transient RPC failure costs a few halving retries before it gives up.

bmwill (approved):

PR #1003, opened 2026-08-11
withdraw: defer bag-to-bag moves to a batched archival GC

Bridgerz:
Move base of the withdrawal deferred-archival stack. Based on `main`; the Rust caller (batch-cap raise + leader GC job) lands stacked on top in #1023. v2-gated — every new archival entry asserts `assert_version_enabled()`, so the deferred flow only runs once the v2 upgrade is enabled.

## Problem

`commit_withdrawal_tx` and `confirm_withdrawal` are the withdrawal flow's two worst transactions by live object reads (Sui object-runtime cache loads):

- `commit_withdrawal_tx`: `12 + 3N + U` runtime objects, hitting the 922 engineered budget at the 298-request cap. The `3N` term is entirely the `requests → processed` ObjectBag move (old Field + child + new Field), one per request.
- `confirm_withdrawal`: `43 + 2N + I`, second worst.

## Change

Make the `status` field authoritative and write it in place; cert-free GC entries perform the deferred bag-to-bag moves in batches after final confirmation — the same pattern `mark_spent` / `cleanup_spent_utxos` already uses, and like them deliberately **not** pause/reconfig-gated (they move no funds and must stay callable during an emergency pause).

- `commit_requests` keeps the request in `requests` (drains BTC, sets Processing + txn link via `borrow_mut`) → commit drops to `12 + 2N + U`. #1023 turns that headroom into a 298 → 447 batch-cap raise.
- `confirm_withdrawal` records `confirmed_timestamp_ms` in place, emits `WithdrawalConfirmed`, and marks input UTXOs spent → `~42 + I`, decoupled from request count.

## Archival entries

All idempotent, all safe for permissionless callers — every write is derivable from already-certified monotone state, so an adversarial caller can at worst archive earlier than the operator would (a semantic no-op). A new `WithdrawalArchived` event announces the terminal status flip for indexers tracking the request objects.

- `archive_confirmed_withdrawals(ids)`: whole-txn atomic archival — the normal path for txns that fit one Sui transaction.
- `archive_withdrawal_requests(txn_id, request_ids)` + `finish_archive_withdrawal_txns(ids)`: chunked path for txns whose request cost exceeds one Sui transaction (needed once the cap rises past ~300). The `withdrawal_txn_id` cross-check runs in **both** location branches because the chunk entry takes caller-supplied ids; the finisher moves the txn only after a contains-only walk (one probe/request) confirms every request has left the hot bag, and silently skips incomplete/already-finished ids so batched calls survive races with in-flight chunk transactions.

## Explicit guards (previously implicit in bag membership)

Deferred archival keeps a committed request in `requests` instead of moving it out, so protections that used to come free from the request having left the bag are now explicit asserts (each unit-tested):

- `approve_withdrawal` refuses non-active requests — without this, an approval-cert replay would reset a committed request to Approved and re-arm a second commit against already-drained BTC.
- `mark_txn_confirmed` refuses replayed confirmation certs — the `confirmed_timestamp_ms`-already-set check is now the only barrier against double-emitting events and re-running the UTXO spend marking.
- `cancel_withdrawal` asserts the request is still active before destroying it — destroying a committed request (still referenced by a live `WithdrawalTransaction`) would permanently abort that txn's archival.

## Cross-version state

Requests committed pre-upgrade already live in `processed`. `update_requests_signed`, `is_request_processing`, and the archival walks are dual-location: they check `requests` first (status-first) and fall back to `processed`.

## Tests

19 new Move tests, full package suite green — in-place commit/confirm, confirmation-replay guard, dual-location `update_requests_signed`, archival idempotency + v1-leftover, chunked partial-then-finish, foreign-request rejection, unconfirmed-txn abort, GC-while-paused, and mixed batches. No changes to existing struct/enum layouts (only a new `WithdrawalArchived` event and new error constants).

Stacked: Rust caller in #1023.

### DIFF
--- a/packages/hashi/sources/btc/withdrawal_queue.move
+++ b/packages/hashi/sources/btc/withdrawal_queue.move
@@ -633,17 +633,77 @@ public(package) fun archive_withdrawal_txn(
         assert!(txn.confirmed_timestamp_ms.is_some(), EWithdrawalNotConfirmed);
         txn.request_ids
     };
-    request_ids.do!(|id| {
-        if (self.requests.contains(id)) {
-            let mut request: WithdrawalRequest = self.requests.remove(id);
-            assert!(request.withdrawal_txn_id == option::some(withdrawal_id), ERequestTxnMismatch);
-            request.status = WithdrawalStatus::Confirmed;
-            self.processed.add(id, request);
-        } else {
-            let request: &mut WithdrawalRequest = self.processed.borrow_mut(id);
-            request.status = WithdrawalStatus::Confirmed;
-        }
+    request_ids.do!(|id| self.archive_request(withdrawal_id, id));
+    let txn: WithdrawalTransaction = self.withdrawal_txns.remove(withdrawal_id);
+    sui::event::emit(WithdrawalArchived {
+        withdrawal_txn_id: withdrawal_id,
+        request_ids,
     });
+    self.confirmed_txns.add(withdrawal_id, txn);
+}
+
+/// Archive one request of a confirmed withdrawal wherever it lives. The
+/// `withdrawal_txn_id` cross-check runs in BOTH branches: the chunked entry
+/// takes caller-supplied ids, and without the check in the fallback branch a
+/// caller could flip an unrelated archived request's status.
+fun archive_request(
+    self: &mut WithdrawalRequestQueue,
+    withdrawal_id: address,
+    request_id: address,
+) {
+    if (self.requests.contains(request_id)) {
+        let mut request: WithdrawalRequest = self.requests.remove(request_id);
+        assert!(request.withdrawal_txn_id == option::some(withdrawal_id), ERequestTxnMismatch);
+        request.status = WithdrawalStatus::Confirmed;
+        self.processed.add(request_id, request);
+    } else {
+        let request: &mut WithdrawalRequest = self.processed.borrow_mut(request_id);
+        assert!(request.withdrawal_txn_id == option::some(withdrawal_id), ERequestTxnMismatch);
+        request.status = WithdrawalStatus::Confirmed;
+    }
+}
+
+/// Chunked archival: archive only the listed requests of a confirmed
+/// withdrawal, leaving the txn in the hot bag. Lets a txn whose request
+/// count exceeds one Sui transaction's runtime-object budget archive across
+/// several transactions; `finish_archive_withdrawal_txn` moves the txn once
+/// every request is archived. No-ops if the txn is already fully archived
+/// (re-run or raced GC); idempotent per request via the `processed` branch.
+public(package) fun archive_withdrawal_requests(
+    self: &mut WithdrawalRequestQueue,
+    withdrawal_id: address,
+    request_ids: &vector<address>,
+) {
+    if (!self.withdrawal_txns.contains(withdrawal_id)) return;
+    {
+        let txn: &WithdrawalTransaction = self.withdrawal_txns.borrow(withdrawal_id);
+        assert!(txn.confirmed_timestamp_ms.is_some(), EWithdrawalNotConfirmed);
+    };
+    request_ids.do_ref!(|id| self.archive_request(withdrawal_id, *id));
+}
+
+/// Finish a chunked archival: move the txn to `confirmed_txns` once none of
+/// its requests remain in `requests`. Silently no-ops while any request is
+/// still unarchived (or if the txn is already archived), so batched GC calls
+/// survive races with in-flight chunk transactions; the caller re-arms and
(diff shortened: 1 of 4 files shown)


packages/hashi/sources/btc/withdrawal_queue.move
```diff
@@ -318,7 +352,7 @@ public(package) fun commit_requests(
     let mut total_btc = sui::balance::zero<BTC>();
 
     withdrawal_txn.request_ids.do_ref!(|id| {
-        let mut request: WithdrawalRequest = self.requests.remove(*id);
+        let request: &mut WithdrawalRequest = self.requests.borrow_mut(*id);
```
dlukegordon, line 355: Keeping committed requests in `requests` makes them unsafe while v1 remains enabled. A caller can invoke the old v1 `cancel_withdrawal`, which only checks `processed`, and destroy a v2-committed request after the cooldown. That strands the batch’s burned hBTC and locked UTXOs. The old approval entry can similarly reset it to `Approved`. Please make v1 disabling atomic with v2 activation and add a test that directly calls the v1 entries against a v2-committed request.

Bridgerz: You are right about the cross-version safety invariant. For this testnet rollout we are intentionally leaving v1 enabled: updated binaries route normal withdrawal and cancellation calls through the resolved active package, but old binaries and direct v1 callers remain able to hit the unsafe legacy path. We are accepting that testnet-only risk to preserve rollback and compatibility testing. Operationally, we will create the proposal for the exact package digest, roll the v1+v2 binary, require validators representing at least two-thirds of committee weight to restart and verify healthy v1 operation before voting, and only then manually execute and publish; votes cannot auto-execute. Publication enables v2 and updated nodes switch automatically. Before mainnet we will squash package versions and storage assumptions so this dual-layout and v1-callable state is not carried forward. I am leaving this unresolved for reviewer follow-up because the code intentionally does not establish the invariant you requested.

Bridgerz: landed #1040! great catch

Bridgerz: Both points addressed. (1) #1040 (merged) adds upgrade_v2 with exclusive=true: publishing and retiring every previous version happen in one transaction, so any future upgrade touching this flow disables v1 atomically. The bootstrap v1->v2 transition itself cannot use it (deployed v1 predates upgrade_v2), so that one uses the documented two-transaction sequence with withdrawals quiesced and verified empty on-chain first; the full risk analysis and checklist are in move-upgrades.mdx. (2) Direct test added in cebbb4629 (on the rust PR of this stack, where the upgrade e2e harness lives): it commits a withdrawal under v2, disables v1, then calls the deployed v1 package's cancel_withdrawal and approve_request directly against the committed request. The assertions pin the abort to EVersionDisabled in the v1 package rather than accepting any failure, and a sanity tail proves the request was untouched and completes normally under v2.

packages/hashi/sources/btc/withdrawal_queue.move
```diff
@@ -553,10 +602,117 @@ public(package) fun finalize_withdrawal_txn(
...
+    if (!self.withdrawal_txns.contains(withdrawal_id)) return;
+    let request_ids = {
+        let txn: &WithdrawalTransaction = self.withdrawal_txns.borrow(withdrawal_id);
+        assert!(txn.confirmed_timestamp_ms.is_some(), EWithdrawalNotConfirmed);
+        txn.request_ids
+    };
+    let mut i = 0;
+    let n = request_ids.length();
+    while (i < n) {
+        if (self.requests.contains(request_ids[i])) return;
```
dlukegordon: Pre-upgrade committed requests already live in `processed` while still `Processing` or `Signed`, so this check lets anyone call the finisher immediately after confirmation and move the transaction to `confirmed_txns` without setting those requests to `Confirmed`. Later archival calls then no-op because the transaction is no longer in `withdrawal_txns`. Please track chunk completion independently of bag location, or verify every request is `Confirmed` before moving the transaction, and add a v1-style regression test.

Bridgerz: Confirmed and fixed in 54c6df830. You are right that bag location cannot signal completion here: processed doubles as both the archive destination and the pre-upgrade commit home, so the walk now additionally requires any processed-resident request to be Confirmed before the txn moves (one status read per processed resident, so the probe stays cheap). Regression test added with a v1-style committed request: the adversarial early finish right after confirmation now no-ops, and archive-then-finish completes; the test was verified to fail against the unfixed finisher before the fix landed.

bmwill (approved):

PR #1004, opened 2026-08-11
leader: archive confirmed withdrawals through a batched, version-gated GC

Bridgerz:
Stacked on #1003 (Move base). Rust side of the deferred withdrawal archival, **including the capacity claim: the batch cap rises 298 → 447 behind the active-version gate.**

**Version gating** (two switches, both keyed on the resolved active version so the flip is atomic with the upgrade checkpoint):
1. The archival GC spawn (`active >= 2` — the entries don't exist in v1 bytecode, and pre-flip there's nothing to archive since v1 confirm archives inline).
2. The batch cap and runtime-object cost models: `max_withdrawal_requests(active)` returns 447 once v2 bytecode executes (in-place commit = 2 objects/request, confirm = 0/request) and the legacy 298 with v1 coefficients (3 and 2) otherwise. The leader sizes batches and validators check commitment shapes with the same function, so mixed generations always agree on what's certifiable — a pre-upgrade chain keeps today's envelope exactly. Compile-time asserts pin both cap derivations: `(922−12−16)/2 = 447` and `(922−12−16)/3 = 298`.

Everything else is unconditional tolerance of both package generations:
- **Mirror**: status-derived `Effect::WithdrawalTxnConfirmed` drives the duration histograms at confirm time; the retire path dedups so v1 inline archival keeps its accounting and GC latency stays out of the histograms.
- **Leader**: broadcast/confirm selection excludes confirmed txns; the approve-retry probe drops requests that are no longer status-Requested; `ConfirmedOnSui` arms the archival scan.
- **GC**: `cleanup_spent_utxos` template (singleton, infinite-backoff retry, checkpoint-freshness waits, boot-armed). Victims carry request ids; the executor **plans per victim**: txns fitting one transaction greedy-pack through the atomic entry, oversized ones (possible once the 447 cap is in effect) split into `archive_withdrawal_requests` chunks + a batched finish, each call priced against the 922 budget.
- **Metrics/CLI**: txn-level `confirmed` class = archival-backlog alert signal; CLI renders confirmed-awaiting-archival accurately.

**Verified**: 630/630 hashi unit tests (incl. archival planning, both-version cost models, version-aware shape checks); `test_bitcoin_withdrawal_e2e_flow` green in 75s (v2 path end-to-end incl. archival GC); `test_withdrawal_committed_under_v1_completes_after_upgrade` green in 66s (fully-signed withdrawal parked under v1 bytecode, governance upgrade, v2 confirm + archival over cross-version state).

Follow-up before the upgrade ships: sui-replay measurement of the real commit/archive coefficients (2/request commit and 3/request archive are principled but unmeasured; the caps only get *more* headroom if they're lower).

### DIFF
--- a/crates/e2e-tests/src/e2e_flow.rs
+++ b/crates/e2e-tests/src/e2e_flow.rs
@@ -37,6 +37,7 @@ mod tests {
     use crate::test_helpers::txid_to_address;
     use crate::test_helpers::wait_for_deposit_confirmation;
     use crate::test_helpers::wait_for_spent_utxo_cleanup;
+    use crate::test_helpers::wait_for_withdrawal_archival;
 
     const MAX_TX_SIZE_BYTES: usize = 131_072;
     const MAX_SERIALIZED_TX_EFFECTS_SIZE_BYTES: usize = 524_288;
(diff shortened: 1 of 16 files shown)


Bridgerz:
GitHub's mergeability cache wedged on this PR after the base advance + head rebase landed simultaneously (git proves the stack merges clean: the base is an ancestor of the head). Recreated with identical branches — superseded by the replacement PR linked below.

PR #1006, opened 2026-08-12
[db] Retire nonce and AVID keyspaces by generation so the disk is freed

0xsiddharthks:
## Summary

Pruning the nonce and AVID keyspaces deleted the right rows but never freed the disk — a delete tombstone is a few dozen bytes against a ~560KB value, so leveled compaction never had reason to rewrite the bottom level. That is what filled every testnet volume and stalled key rotation. These keyspaces are now split across two generations by epoch parity, so retiring an epoch drops a whole generation instead of tombstoning rows.

## Changes

- Add `EpochGenerations` and use it for `nonce_messages` and the three AVID keyspaces.
- Retire an epoch by deleting and recreating the generation that cannot hold it, rather than deleting rows. `Keyspace::clear` would not do — fjall only unlinks a table once a compaction marks it deleted, so a clear orphans the files until the next recovery.
- Drop the per-row prune for these keyspaces, which also takes a full-keyspace scan out of every reconfig.
- Reads fall back to the other generation, so rows written before the split stay visible while their epoch is live.
- Generation 0 keeps the pre-split keyspace name, so existing data is retired by the normal epoch change and there is no migration.

## Notes

The drop needs no check: every writer stores under the node's own current MPC epoch, which `handle_reconfig` sets to the target before any of this runs, so the newest row the dropped generation can hold is `cutoff - 1`.

Retention holds at most two epochs, ~3.5GB at current testnet rates. Existing data is reclaimed within two epoch changes, since the first one may drop the empty generation.

`dealer_messages`, `rotation_messages` and the key keyspaces are unchanged, and none of these keyspaces are in `BACKUP_KEYSPACES`.

A node that has already filled its volume needs a resize and a restart first — fjall stays poisoned until the process restarts.

## Test plan

- `test_retiring_an_epoch_returns_the_bytes_to_the_filesystem` reads the keyspace directories rather than fjall's accounting, which is what caught the `clear` behaviour above.
- Tests for the two-epoch bound, the pre-split fallback, deletes across generations, and reopening after a retire.
- `test_prune_messages_below_basic` updated to the generation contract.

### DIFF
--- a/crates/hashi/src/db.rs
+++ b/crates/hashi/src/db.rs
@@ -69,39 +69,171 @@ pub struct Database {
     //
     // key: (big endian u64 epoch) + (big endian u32 batch_index) + (32-byte validator address)
     // value: BCS-serialized batch_avss::Message
-    nonce_messages: Keyspace,
+    nonce_messages: EpochGenerations,
 
     // Column Family used to store per-round AVID receiver state for restart/resume.
     //
     // key: (big endian u64 epoch) + (big endian u32 batch_index) + (32-byte validator address)
     // value: BCS-serialized AvidRoundState
-    avid_round_states: Keyspace,
+    avid_round_states: EpochGenerations,
 
     // Column Family used to store this node's own AVID dealer builder per batch.
     //
     // key: (big endian u64 epoch) + (big endian u32 batch_index)
     // value: BCS-serialized AvssMessageBuilder
-    avid_dealer_builders: Keyspace,
+    avid_dealer_builders: EpochGenerations,
 
     // Column Family used to store this node's held AVID vote and outbound echoes per round.
     //
     // key: (big endian u64 epoch) + (big endian u32 batch_index) + (32-byte dealer address)
     // value: BCS-serialized HeldAvidEchoes
-    avid_held_echoes: Keyspace,
+    avid_held_echoes: EpochGenerations,
 }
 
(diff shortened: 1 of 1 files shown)


crates/hashi/src/db.rs
```diff
@@ -69,40 +69,150 @@ pub struct Database {
...
+    /// at most two epochs on disk.
+    ///
+    /// Deleted and recreated rather than cleared, because fjall only unlinks a
+    /// table once a compaction marks it deleted — clearing would orphan the
+    /// files until the next recovery.
+    fn retire_epochs_below(&self, db: &fjall::Database, cutoff_epoch: u64) -> Result<()> {
+        if cutoff_epoch == 0 {
+            return Ok(());
+        }
+        let index = Self::index(cutoff_epoch) ^ 1;
```
dlukegordon, line 207: Epoch parity is not safe here because reconfig can skip epochs. For example, going from epoch 10 to 12 keeps both in generation 0 and repeatedly deletes the empty generation 1. Same-parity jumps would let this keyspace grow forever. Can we track and rotate the active generation on every reconfig instead of deriving it from epoch parity, and add a test for 10 -> 12 -> 14?

0xsiddharthks: 👀 makes sense. Let me add this

PR #1008, opened 2026-08-12
[db] Report not-ready when fjall has poisoned the database

0xsiddharthks:
Stacked on #1007.

## Summary

fjall poisons the database on a failed write and never recovers in-process — every later write returns `Poisoned` and only reopening clears it. The node carried on anyway, answering RPCs and passing its readiness probe while persisting nothing, which is how a full volume on testnet turned into a stalled key rotation with a fleet that still looked healthy.

## Changes

- Latch a flag on `Database` when a write comes back `Poisoned`, and expose `is_poisoned`.
- Fail `/ready` while latched, so the node stops taking traffic it cannot persist and a rollout will not proceed past it.
- Export `hashi_db_poisoned`.

## Notes

Only writes are checked, because fjall only consults the flag on the write path — a read cannot surface poisoning.

The flag never clears by design: the process has to be restarted, and if the cause was `ENOSPC` an operator has to resize first.

## Test plan

`test_poisoning_latches_and_nothing_else_does` — a healthy write and an unrelated IO error must not latch, `Poisoned` must, and a later success must not clear it.

### DIFF
--- a/crates/hashi/src/db.rs
+++ b/crates/hashi/src/db.rs
@@ -137,6 +137,21 @@ impl EpochGenerations {
         (epoch % 2) as usize
     }
 
+    /// The name this keyspace had before it was split, which generation 0 keeps.
+    fn name(&self) -> &'static str {
+        self.names[0]
+    }
+
+    /// Bytes fjall accounts to this keyspace across both generations.
+    fn disk_space(&self) -> u64 {
+        self.generations
+            .read()
+            .unwrap()
+            .iter()
+            .map(Keyspace::disk_space)
+            .sum()
+    }
+
     fn generation(&self, index: usize) -> Keyspace {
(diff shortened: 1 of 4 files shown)


0xsiddharthks:
Closing: a readiness gate is the wrong response to a poisoned database. Only DKG, rotation and nonce-batch generation write to fjall — signing, the leader and the RPC handlers never touch it — so a poisoned node keeps signing withdrawals from its in-memory presig pool and only loses refill and rotation. The realistic cause, a full disk, hits the whole fleet at once, so failing `/ready` would have pulled every pod from its endpoints and turned a stalled rotation into a bridge outage.

The `hashi_db_poisoned` gauge moved into #1007, probing fjall's own flag instead of latching write errors, with a rule in the validator alert pack.

PR #1010, opened 2026-08-13
[db] Compact the database after reconfig so pruned rows free the disk

0xsiddharthks:
## Summary

Pruning deletes the right rows but never frees the disk — a tombstone is a few dozen bytes next to a ~560KB nonce message, so leveled compaction never has reason to rewrite the bottom level. This runs a major compaction after each epoch change, off the reconfig path, which is the simpler alternative to #1006 that Luke and Brandon suggested.

## Changes

- Add `Database::major_compact`: per keyspace, flush then compact, reporting duration and size; a failed flush fails that keyspace instead of compacting blind.
- After compacting, write a marker row to a new `maintenance` keyspace and rotate it, so the retired table files are always unlinked before returning. If that rotation fails the run is reported failed — logged and counted under `hashi_db_major_compaction_failures_total{keyspace="unlink"}` — since nothing was actually released.
- Call it after the prune in `handle_reconfig`, awaited on a blocking thread, and enqueue the backup only after it finishes — the backup's database-wide snapshot would otherwise pin every value the compaction is meant to drop.
- Add `hashi_db_major_compaction_duration_seconds` and `hashi_db_major_compaction_failures_total`.

## Notes

Two ordering details that fjall's API doesn't hint at. Both showed up once the test measured real bytes in the keyspace directory instead of `disk_space()`, which only counts tables in the current version:

- **Flush before compacting.** A compaction merges tables, not memtables, so the prune's tombstones have to reach a table first — otherwise they aren't in the merge and the values they shadow survive it. Skipping the flush left 5.5MB of a 10.3MB keyspace behind instead of 3.7KB.
- **`major_compact` doesn't unlink anything.** It retires the tables it merged away, but fjall only unlinks them while rotating a memtable, and rotating an empty memtable is a no-op. The marker row in the `maintenance` keyspace gives the rotation something to flush, and one rotation sweeps every keyspace, so the disk is actually given back before `major_compact` returns.

Worth watching on one node before mainnet, which is what the duration metric is for. Major compaction takes fjall's per-keyspace major-compaction lock and background compactions take it shared, so a long run lets L0 pile up on that keyspace and can throttle writes. It also rewrites the whole keyspace and needs room for the output before the old tables go, so it is weakest exactly when the disk is nearly full. Running it right after the prune and before `prepare_signing` refills the presig pool keeps both windows small — nearly everything is tombstoned at that point, so the merge writes almost nothing.

## Test plan

`test_major_compaction_frees_what_pruning_only_tombstoned` writes 4096 nonce rows, prunes them, asserts the bytes are still on disk (the bug), then compacts and asserts they come back. It also pins the flush ordering — drop the flush and the assertion fails.

Full suite passes, clippy clean with `-D warnings`.

Canaried on devnet `hashi-server-2` and `-3` (partition roll, nodes 0-1 untouched as controls), across two live epoch boundaries. At the epoch-99 reconfig both nodes compacted all 11 keyspaces in ~310ms, awaited on the reconfig path, with the backup enqueued strictly after (visible in the logs 5ms behind the compaction summary):

- The node carrying 15h of accumulated garbage: `rotation_messages` 88.8MB → 21.3MB, `nonce_messages` 41.8MB → 0, and the DB volume dropped 125M → 41M **immediately** — the maintenance-keyspace rotation unlinked the retired tables with no follow-up write traffic.
- The node compacted the epoch before: a clean ~325ms no-op (21.3MB → 21.3MB), i.e. the steady-state cost of running this every epoch.

Both stayed ready throughout with zero restarts, `hashi_db_major_compaction_duration_seconds{keyspace}` populated, and no failure counters incremented.

### DIFF
--- a/crates/hashi/src/db.rs
+++ b/crates/hashi/src/db.rs
@@ -122,6 +122,15 @@ pub struct CompactedKeyspace {
     pub error: Option<fjall::Error>,
 }
 
+/// Outcome of a whole-database major compaction.
+pub struct MajorCompaction {
+    pub keyspaces: Vec<CompactedKeyspace>,
+    /// Error from the final rotation that unlinks the retired tables. When
+    /// set, the compacted bytes stay on disk until a later rotation, so the
+    /// run as a whole failed even if every keyspace compacted cleanly.
+    pub unlink_error: Option<fjall::Error>,
+}
+
 /// Keyspaces included in snapshot backups. Add new backup/restore keyspaces here.
 #[derive(Clone, Copy)]
 enum BackupKeyspace {
@@ -250,7 +259,7 @@ impl Database {
     ///
     /// Blocking and slow, and it locks out background compaction for each
     /// keyspace while it runs — call it from a blocking thread.
-    pub fn major_compact(&self) -> Vec<CompactedKeyspace> {
+    pub fn major_compact(&self) -> MajorCompaction {
         let compacted: Vec<_> = self
             .all_keyspaces()
             .into_iter()
@@ -281,14 +290,15 @@ impl Database {
             .duration_since(std::time::UNIX_EPOCH)
             .unwrap_or_default()
             .as_secs();
-        if let Err(e) = self
+        let unlink_error = self
             .maintenance
             .insert(MAINTENANCE_COMPACTION_KEY, now.to_be_bytes())
             .and_then(|_| self.maintenance.rotate_memtable_and_wait().map(|_| ()))
-        {
-            tracing::warn!("unlinking compacted tables failed, retrying next epoch: {e}");
+            .err();
+        MajorCompaction {
+            keyspaces: compacted,
+            unlink_error,
         }
-        compacted
     }
 
     /// Store an encryption private key, keyed by its public key, plus the
@@ -1960,8 +1970,14 @@ pub(crate) mod tests {
             "pruning alone does not free the disk — this is the bug being fixed",
         );
 
-        let compacted = db.major_compact();
-        let nonce = compacted
+        let compaction = db.major_compact();
+        assert!(
+            compaction.unlink_error.is_none(),
+            "unlink rotation failed: {:?}",
+            compaction.unlink_error
+        );
+        let nonce = compaction
+            .keyspaces
             .iter()
             .find(|keyspace| keyspace.name == NONCE_MESSAGES_CF_NAME)
             .expect("nonce_messages must be compacted");
--- a/crates/hashi/src/db.rs
+++ b/crates/hashi/src/db.rs
@@ -214,6 +214,7 @@ impl Database {
         self.db.snapshot()
     }
 
+    /// Every keyspace, for whole-database maintenance. Add new keyspaces here.
     fn all_keyspaces(&self) -> [(&'static str, &Keyspace); 10] {
         [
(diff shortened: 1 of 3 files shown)


crates/hashi/src/mpc/service.rs
```diff
@@ -1378,6 +1380,7 @@ impl MpcService {
         {
             error!("Failed to prune old MPC messages below epoch {target_epoch}: {e}");
         }
+        self.spawn_major_compaction(target_epoch);
```
dlukegordon: Backup can hold a fjall snapshot that stops compaction from dropping the old values. Please move backup until after compaction, and run compaction with `spawn_blocking(...).await` before `prepare_signing`. This also removes the need for the detached task, mutex, and skipped-run handling.

crates/hashi/src/db.rs
```diff
@@ -205,6 +214,65 @@ impl Database {
...
+    /// tombstone is a few dozen bytes against a nonce message of hundreds of
+    /// kilobytes, so leveled compaction never rewrites the bottom level.
+    ///
+    /// Blocking and slow, and it locks out background compaction for each
+    /// keyspace while it runs — keep it off the reconfig path.
+    pub fn major_compact(&self) -> Vec<CompactedKeyspace> {
+        // Compaction merges tables, not memtables, so the prune's tombstones
+        // have to reach a table first or the values they shadow survive it.
+        for (name, keyspace) in self.all_keyspaces() {
+            if let Err(e) = keyspace.rotate_memtable_and_wait() {
```
dlukegordon: If this flush fails, compaction can still return success without seeing the prune tombstones. Please mark this keyspace as failed and skip its compaction so the failure is logged and counted correctly.

crates/hashi/src/db.rs
```diff
@@ -205,6 +214,65 @@ impl Database {
...
+                CompactedKeyspace {
+                    name,
+                    before,
+                    after: keyspace.disk_space(),
+                    elapsed: started.elapsed(),
+                    error,
+                }
+            })
+            .collect();
+        // A compaction retires the tables it merged away but does not unlink
```
dlukegordon: This rotation does nothing when the memtable is empty, so the old table files may stay on disk even though compaction is reported as successful. Please trigger a guaranteed non-empty rotation, possibly through a small maintenance keyspace, wait for it to finish, and only then report completion.

0xsiddharthks:
@dlukegordon Thanks! Implemented the changes.

crates/hashi/src/db.rs
```diff
@@ -205,6 +226,71 @@ impl Database {
...
+            .collect();
+        // Compaction retires the tables it merged away, but only a memtable
+        // rotation unlinks them, and rotating an empty memtable is a no-op.
+        // Write a marker row so the rotation always has something to flush;
+        // one rotation sweeps every keyspace.
+        let now = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs();
+        if let Err(e) = self
```
dlukegordon: This rotation is what actually releases the disk, but its failure is only logged and isn't reflected in `hashi_db_major_compaction_failures_total`; the caller still reports the run as finished. Please return this error to `run_major_compaction` and count/report the whole compaction run as failed.

dlukegordon (approved):
The original review comments are addressed. One remaining observability issue is noted inline.

PR #1013, opened 2026-08-13
Fix end_reconfig signature in reconfiguration docs

leonschh:
### What

Update the end_reconfig signature in `design/docs/reconfiguration.mdx` to match the implementation in [`packages/hashi/sources/core/reconfig.move`](https://github.com/MystenLabs/hashi/blob/main/packages/hashi/sources/core/reconfig.move#L81-L86).

### Why

The documentation previously listed `signature` and `signers_bitmap` as separate `vector<u8>` parameters, but the actual entry function accepts a single `mpc_cert: CommitteeSignature`.

The documentation now reflects the actual API.


jessiemongeon1:
## Style Guide Audit

PR #1023, opened 2026-08-14
leader: archive confirmed withdrawals through a batched, version-gated GC

Bridgerz:
Stacked on #1003 (Move base). Rust side of the deferred withdrawal archival — the leader-driven archival GC, the active-version routing that makes v2's in-place-commit semantics reachable, and the capacity bump (batch cap 298 → 447) it unlocks.

The shared version-framework plumbing — `SUPPORTED_PACKAGE_VERSIONS = [1, 2]`, the executor's generic active-version resolution (`active_call_package_id`), `fetch_initial_shared_version`, `HashiClient::latest_package_id` — now lives in **main** via the extracted base #1026, so this PR no longer carries it. What remains here is withdrawal-flow-specific and reads that already-in-place support set.

## Change — version routing (all keyed on the resolved active version, atomic with the upgrade checkpoint)

1. **Active-package resolvers.** `active_version_package()` resolves the active version against the already-`[1, 2]` support set and returns its published package id; `withdrawal_call_package()` wraps it with an original-id fallback pre-resolution — at active=1 it returns the original id, preserving today's behavior byte-for-byte. Every withdrawal-flow entry call (finalize, confirm, cancel, and the committee-signature args) routes through this one resolver, so a batch's whole lifecycle executes one consistent bytecode generation: v2's in-place commit leaves requests where v1's finalize/confirm wouldn't find them, and the v1 cancel bytecode's processed-bag gate misses v2 in-place-committed requests and would destroy a request a live txn references. All calls within a PTB use the same resolution (Sui links one version of a package per transaction; mixed targets fail with `INVALID_LINKAGE`). A resolution torn across the upgrade flip fails simulation and self-heals via the callers' retry machinery. The CLI cancel path resolves the same way through a governance-only reader.

2. **Capacity.** `max_withdrawal_requests(active)`: 447 under v2 (in-place commit = 2 runtime objects/request, confirm = 0/request), legacy `LEGACY_MAX_WITHDRAWAL_REQUESTS = 298` with v1 coefficients otherwise — used by BOTH leader sizing and validator shape checks, so mixed generations always agree. Compile-asserted: `922 budget − 12 fixed − 2·447 = 16` reserved funding inputs (v1: `− 3·298 = 16`).

3. **Archival GC driver.** The GC spawn gates on active ≥ 2; victims (`ArchiveVictim`) carry request ids and `execute_archive_confirmed_withdrawals` plans per victim — atomic packing via `archive_confirmed_withdrawals` for txns that fit one GC transaction, `archive_withdrawal_requests` chunks followed by `finish_archive_withdrawal_txns` for oversized ones. The entries exist only in the v2 package; both the gate and the call target resolve through the active version.

## Change — both-worlds (unconditional)

Status-derived `Effect::WithdrawalTxnConfirmed` + retire dedup (duration histograms exact in both worlds), confirmed-txn exclusion in leader selection, a still-approvable retry probe, a metrics `confirmed` class (the archival-backlog alert signal), and status-accurate CLI (a request stays in the mirrored map for its whole live lifecycle, so status renders "Signing (X/N)" / confirmed rather than a flat "Committed").

## Tests

The v2 path genuinely executes end-to-end — an earlier revision's e2e greens were vacuous (with the binary effectively pinned to v1 the GC never spawned and v1 archived inline; the CI drain-mode failure exposed it):
- `test_drain_mode_max_request_batch`: 447-request batch through v2 commit at the modeled 922-object budget, signed, confirmed, chunk-archived — 102s
- `test_bitcoin_withdrawal_e2e_flow` (98s) and `test_batch_withdrawal` (99s) with the archival GC live
- `test_withdrawal_committed_under_v1_completes_after_upgrade` (70s): cross-version state through the real routed upgrade flip
- 633/633 unit tests (both-version cost models, version-aware shape checks, archival planning)

## Follow-up

sui-replay measurement of the real commit/archive coefficients before the upgrade ships — the drain e2e empirically pins commit ≤ 1000 objects at N=447, but the model margin still needs to be measured precisely.

### DIFF
--- a/crates/hashi/src/cli/commands/withdraw.rs
+++ b/crates/hashi/src/cli/commands/withdraw.rs
@@ -18,6 +18,7 @@ use crate::cli::config::CliConfig;
 use crate::cli::print_info;
 use crate::cli::print_success;
 use crate::cli::types::display;
+use crate::onchain::types::WithdrawalRequest;
 use crate::onchain::types::WithdrawalStatus;
 use crate::onchain::types::WithdrawalTransaction;
 
@@ -87,6 +88,12 @@ async fn request(
         .context("Withdrawal address does not match the configured Bitcoin network")?;
     let destination_bytes = witness_program_from_address(&btc_addr)?;
 
+    // `request_withdrawal` gates on `assert_version_enabled`, so the call
+    // must target the active version's package: a v1-targeted creation
+    // aborts once v1 is disabled. The resolved state also routes the batch
+    // executor below.
+    let (onchain, call_package) = super::resolve_active_call_package(config, hashi_ids).await?;
+
     let mut client = crate::sui_rpc_client::new_sui_rpc_client(&config.sui_rpc_url)?;
 
     // A single request supports all tx modes (execute / dry-run /
@@ -97,7 +104,7 @@ async fn request(
 
         let builder = crate::sui_tx_executor::build_create_withdrawal_request(
             hashi_ids,
-            hashi_ids.package_id,
+            call_package,
             amount,
             destination_bytes,
         );
@@ -136,9 +143,12 @@ async fn request(
          and --dry-run apply to a single withdrawal request"
     );
 
-    // Execute mode guarantees a signer (rejected above otherwise).
+    // Execute mode guarantees a signer (rejected above otherwise). The
+    // attached reader state makes the executor route its calls through the
+    // active version's package.
     let signer = signer.expect("execute mode requires a signer");
-    let mut executor = crate::sui_tx_executor::SuiTxExecutor::new(client, signer, hashi_ids);
+    let mut executor = crate::sui_tx_executor::SuiTxExecutor::new(client, signer, hashi_ids)
+        .with_onchain_state(&onchain);
 
     // ~3 PTB commands per request (split + call) vs the 1024 command cap.
     const CHUNK_SIZE: usize = 250;
@@ -202,23 +212,11 @@ async fn cancel(config: &CliConfig, tx_opts: &TxOptions, request_id: &str) -> Re
 
     // Resolve the active version's package so the cancel runs the bytecode
     // generation whose committed-request gate matches the chain's state
-    // (v1's bag-membership gate misses v2 in-place-committed requests).
-    let call_package = crate::onchain::OnchainState::new_reader(
-        &config.sui_rpc_url,
-        hashi_ids,
-        None,
-        crate::onchain::ScrapeScope::GovernanceOnly,
-    )
-    .await
-    .ok()
-    .and_then(|state| {
-        let guard = state.state();
-        let version = guard
-            .version_support(crate::constants::SUPPORTED_PACKAGE_VERSIONS)
-            .active_version()?;
-        guard.package_versions().get(version)
-    })
-    .unwrap_or(hashi_ids.package_id);
+    // (v1's bag-membership gate misses v2 in-place-committed requests). A
+    // resolution failure must abort the command, not fall back: a v1-routed
+    // cancel aimed at a v2-committed request would destroy a request its
+    // live withdrawal txn still references.
+    let (_onchain, call_package) = super::resolve_active_call_package(config, hashi_ids).await?;
     let builder =
         crate::sui_tx_executor::build_cancel_withdrawal(hashi_ids, call_package, &req_addr, sender);
 
@@ -464,6 +462,11 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
 
     let requests = client.fetch_withdrawal_requests()?;
     let pending = client.fetch_withdrawal_txns()?;
+    // The v2 flow commits requests in place, so the mirrored map holds both
+    // the actionable queue and requests already committed into a withdrawal
+    // txn (those linger until the archival GC sweeps them and are counted
+    // through their txn's request_count).
+    let (queued, committed_in_place) = partition_queued(&requests);
     // Confirmed txns linger in the pending map until the archival GC sweeps
     // them; classify them first so they are never counted as actionable
     // "signed" (they are also fully signed).
@@ -476,7 +479,7 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
 
     match output_format {
         OutputFormat::Json => {
-            let queued: Vec<_> = requests
+            let queued_rows: Vec<_> = queued
                 .iter()
                 .map(|wr| {
                     serde_json::json!({
@@ -494,9 +497,10 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
             println!(
                 "{}",
                 serde_json::to_string_pretty(&serde_json::json!({
-                    "queued": queued,
+                    "queued": queued_rows,
                     "withdrawal_txns": withdrawal_txns,
-                    "queued_count": requests.len(),
+                    "queued_count": queued.len(),
+                    "committed_in_place_count": committed_in_place.len(),
                     "committed_count": committed_count,
                     "signed_count": signed_count,
                     "confirmed_count": confirmed_count,
@@ -510,7 +514,7 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
             if requests.is_empty() && pending.is_empty() {
                 print_info("No withdrawal requests found.");
             } else {
-                if !requests.is_empty() {
+                if !queued.is_empty() {
                     println!("  {}", "Queued:".bold().underline());
                     println!(
                         "  {:<20} {:<14} {:<10} {:<20} {}",
@@ -520,7 +524,7 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
                         "Caller".bold(),
                         "Requested".bold()
                     );
-                    for wr in &requests {
+                    for wr in &queued {
                         println!(
                             "  {:<20} {:<14} {:<10} {:<20} {}",
                             display::format_address_full(&wr.id),
@@ -533,7 +537,7 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
                 }
 
                 if !pending.is_empty() {
-                    if !requests.is_empty() {
+                    if !queued.is_empty() {
                         println!();
                     }
                     println!("  {}", "Pending Broadcast:".bold().underline());
@@ -556,8 +560,10 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
                 }
 
                 println!(
-                    "\n  {} queued, {} committed, {} signed, {} confirmed awaiting archival",
-                    requests.len(),
+                    "\n  {} queued, {} committed in place; txns: {} committed, {} signed, \
+                     {} confirmed awaiting archival",
+                    queued.len(),
+                    committed_in_place.len(),
                     committed_count,
                     signed_count,
                     confirmed_count
@@ -569,6 +575,17 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
     }
 
     Ok(())
+}
+
+/// Split the mirrored request map into the actionable queue
+/// (Requested/Approved) and the requests the v2 flow committed in place
+/// (Processing/Signed, plus the defensively handled Confirmed). The latter
+/// must not report as queued backlog: their BTC is already drained into a
+/// withdrawal txn, which the txn view counts.
+fn partition_queued(
+    requests: &[WithdrawalRequest],
+) -> (Vec<&WithdrawalRequest>, Vec<&WithdrawalRequest>) {
+    requests.iter().partition(|wr| wr.status.is_active())
 }
 
 /// The JSON row for one in-flight withdrawal transaction.
@@ -658,5 +675,51 @@ mod tests {
         txn.confirmed_timestamp_ms = Some(2);
         let row = withdrawal_txn_row(&txn);
         assert_eq!(row["status"], "confirmed");
+    }
+
+    fn request_with_status(status: WithdrawalStatus) -> WithdrawalRequest {
+        WithdrawalRequest {
+            id: sui_sdk_types::Address::new([1; 32]),
+            sender: sui_sdk_types::Address::new([2; 32]),
+            btc_amount: 1,
+            bitcoin_address: vec![0; 20],
+            created_timestamp_ms: 0,
+            status,
+            approval_cert: None,
+            approved_timestamp_ms: None,
+            withdrawal_txn_id: None,
+            sui_tx_digest: sui_sdk_types::Digest::new([0; 32]),
+            btc: 1,
+        }
+    }
+
+    /// Requests the v2 flow committed in place linger in the mirrored map
+    /// until the archival GC sweeps them; the queued view must exclude them
+    /// (they are already counted through their withdrawal txn).
+    #[test]
+    fn queued_view_excludes_committed_in_place_requests() {
+        let requests: Vec<_> = [
+            WithdrawalStatus::Requested,
+            WithdrawalStatus::Approved,
+            WithdrawalStatus::Processing,
+            WithdrawalStatus::Signed,
+            WithdrawalStatus::Confirmed,
+        ]
+        .into_iter()
+        .map(request_with_status)
+        .collect();
+
+        let (queued, committed_in_place) = partition_queued(&requests);
+        let statuses = |group: &[&WithdrawalRequest]| {
+            group
+                .iter()
+                .map(|wr| wr.status.as_str())
+                .collect::<Vec<_>>()
+        };
+        assert_eq!(statuses(&queued), ["requested", "approved"]);
+        assert_eq!(
+            statuses(&committed_in_place),
+            ["processing", "signed", "confirmed"]
+        );
     }
 }
--- a/crates/hashi/src/cli/commands/withdraw.rs
+++ b/crates/hashi/src/cli/commands/withdraw.rs
@@ -89,10 +89,11 @@ async fn request(
     let destination_bytes = witness_program_from_address(&btc_addr)?;
 
     // `request_withdrawal` gates on `assert_version_enabled`, so the call
-    // must target the active version's package: a v1-targeted creation
-    // aborts once v1 is disabled. The resolved state also routes the batch
-    // executor below.
-    let (onchain, call_package) = super::resolve_active_call_package(config, hashi_ids).await?;
+    // must target the withdrawal-effective version's package: v1 while the
+    // bootstrap window keeps v1 enabled (matching the flow's dormancy), the
+    // active version after, where a v1-targeted creation would abort. The
+    // resolved state also routes the batch executor below.
+    let (onchain, call_package) = super::resolve_withdrawal_call_package(config, hashi_ids).await?;
 
     let mut client = crate::sui_rpc_client::new_sui_rpc_client(&config.sui_rpc_url)?;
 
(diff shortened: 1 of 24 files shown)


crates/hashi/src/cli/commands/withdraw.rs
```diff
@@ -199,12 +200,27 @@ async fn cancel(config: &CliConfig, tx_opts: &TxOptions, request_id: &str) -> Re
...
+    .await
+    .ok()
+    .and_then(|state| {
+        let guard = state.state();
+        let version = guard
+            .version_support(crate::constants::SUPPORTED_PACKAGE_VERSIONS)
+            .active_version()?;
+        guard.package_versions().get(version)
+    })
+    .unwrap_or(hashi_ids.package_id);
```
dlukegordon: If this governance read fails while v1 and v2 are both enabled, we silently target v1 even though v2 is active. The v1 gate still accepts the call in that state, and `cancel_withdrawal` can remove a request referenced by a live v2 transaction. Please propagate reader/resolution failures and use the original package only when a successful snapshot resolves active version 1.

crates/hashi/src/metrics.rs
```diff
@@ -1374,77 +1374,88 @@ impl Metrics {
...
+            let mut requested = Vec::new();
+            let mut approved = Vec::new();
+            let mut committed = Vec::new();
+            let mut confirmed_pending_archive = Vec::new();
+            for r in hashi.bitcoin().withdrawal_queue.requests().values() {
+                match r.status {
+                    WithdrawalStatus::Requested => requested.push(r),
+                    WithdrawalStatus::Approved => approved.push(r),
+                    WithdrawalStatus::Processing | WithdrawalStatus::Signed => committed.push(r),
+                    WithdrawalStatus::Confirmed => confirmed_pending_archive.push(r),
```
dlukegordon, line 1429: A v2-confirmed request remains `Signed` until archival moves it out of this map, so this `Confirmed` branch cannot represent pending archival. The new gauge stays zero and confirmed requests remain counted as committed. Please classify a hot request as pending archival when its `withdrawal_txn_id` references a confirmed mirrored transaction.

crates/hashi/src/cli/commands/withdraw.rs
```diff
@@ -403,6 +499,7 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
                     "queued_count": requests.len(),
```
dlukegordon: Under v2 this collection also contains Processing and Signed requests, but every entry is emitted and counted as queued here. That duplicates committed withdrawals in both `queued` and `withdrawal_txns` and makes `queued_count` inaccurate until archival. Please restrict the queued view to Requested/Approved requests or expose separately classified lifecycle groups.

crates/hashi/src/leader/garbage_collection.rs
```diff
@@ -246,6 +276,101 @@ impl LeaderService {
...
+        }
+
+        let Some(active) = self.inner.onchain_state().active_package_version() else {
+            return;
+        };
+        if active < WITHDRAWAL_ARCHIVE_MIN_PACKAGE_VERSION {
+            return;
+        }
+
+        if !self.withdrawal_archive_scan_needed {
```
dlukegordon, line 310: This scan is re-armed only from the local confirmation task result. Other nodes observe `WithdrawalTxnConfirmed` in the mirror but leave `withdrawal_archive_scan_needed` false after a barren scan. That can delay archival for a full leader rotation, or indefinitely when a valid confirmation is submitted outside this local result path. Please make visible confirmed transactions arm the scan, rather than relying exclusively on local submission bookkeeping.

dlukegordon (commented):
Review findings for the deferred withdrawal archival changes.

crates/hashi/src/withdrawals.rs
```diff
@@ -107,17 +107,44 @@ pub(crate) fn withdrawal_limiter_consumption_amount(txn: &WithdrawalTransaction)
...
+pub(crate) const DEFERRED_ARCHIVAL_MIN_PACKAGE_VERSION: u64 = 2;
+
+/// The request cap for the active package version. The larger v2 cap is
+/// only safe once the chain executes v2 bytecode (in-place commit) AND the
+/// fleet runs binaries that accept it — the same rollout discipline the
+/// upgrade train already requires: binaries first, then the upgrade tx.
+/// Gating on the resolved active version makes the flip atomic with the
+/// upgrade checkpoint on every node.
+pub(crate) fn max_withdrawal_requests(active_package_version: Option<u64>) -> usize {
+    if active_package_version.is_some_and(|v| v >= DEFERRED_ARCHIVAL_MIN_PACKAGE_VERSION) {
```
dlukegordon: The legacy v1→v2 upgrade leaves both versions enabled, so selecting v2 here starts creating in-place committed requests while anyone can still directly call the enabled v1 package. As the new upgrade test documents, v1 cancellation/approval does not recognize that layout and can refund or reset a request referenced by a live transaction, duplicating withdrawal value. Please do not activate v2 withdrawal routing, capacity, or archival semantics until v1 has been disabled; the CLI fail-closed fix alone does not prevent direct v1 calls.

crates/hashi/src/config.rs
```diff
@@ -152,8 +152,10 @@ pub struct Config {
     /// transaction. The batch commits immediately once this many requests are
     /// ready, without waiting for `withdrawal_batching_delay_ms` to elapse.
     ///
-    /// Defaults to 298 (the algorithm's hard upper bound, set by the Sui
-    /// commit transaction's runtime-object budget). Note that large batches
+    /// Defaults to 447 (the algorithm's hard upper bound, set by the Sui
+    /// commit transaction's runtime-object budget under the deferred-archival
+    /// package; a chain still executing v1 bytecode is additionally capped
+    /// at 298 at batch-build time). Note that large batches
```
dlukegordon, line 158: The scheduler still uses this configured 447 maximum as its immediate-fire threshold, while the v1 cap is applied only later during commitment construction. With v1 active, 298–446 approved requests therefore wait for 447 requests or the batching timeout instead of firing at the executable v1 capacity. Please use `withdrawal_max_batch_size().min(max_withdrawal_requests(active_package_version))` for the readiness threshold and add a v1 test at 298 requests.

dlukegordon (commented):
Two additional findings not covered by the existing review comments.

crates/hashi/src/withdrawals.rs
```diff
@@ -152,9 +179,64 @@ const _: () = assert!(
...
+/// Conservative upper bounds pending sui-replay measurement; the executor's
+/// greedy packer sizes each GC transaction from these.
+pub(crate) const WITHDRAWAL_ARCHIVE_FIXED_RUNTIME_OBJECTS: usize = 12;
+pub(crate) const WITHDRAWAL_ARCHIVE_RUNTIME_OBJECTS_PER_TXN: usize = 3;
+pub(crate) const WITHDRAWAL_ARCHIVE_RUNTIME_OBJECTS_PER_REQUEST: usize = 3;
+pub(crate) const WITHDRAWAL_ARCHIVE_RUNTIME_OBJECT_BUDGET: usize = WITHDRAWAL_RUNTIME_OBJECT_BUDGET;
+
+/// Cost of `finish_archive_withdrawal_txn`'s completeness walk: one
+/// `contains` probe per request, on top of the txn borrow.
+pub(crate) const WITHDRAWAL_ARCHIVE_FINISH_RUNTIME_OBJECTS_PER_REQUEST: usize = 1;
```
dlukegordon: `finish_archive_withdrawal_txn` does more than one probe per request: after `processed.contains(id)` it borrows the request to read its status (`withdrawal_queue.move:716-718`). That touches the dynamic-field wrapper and child request, so this should cost 2 runtime objects per archived request. At 1, the planner packs two 447-request victims into one `Finish` call (900 modeled objects, roughly 1,800 actual), and every retry rebuilds the same failing call. Please use 2/request or finish one oversized victim per transaction, and add a two-max-victim planning test.

crates/hashi/src/cli/commands/withdraw.rs
```diff
@@ -199,12 +200,27 @@ async fn cancel(config: &CliConfig, tx_opts: &TxOptions, request_id: &str) -> Re
             "No sender available: pass --sender (the refund recipient) or configure a keypair",
         )?;
 
-    let builder = crate::sui_tx_executor::build_cancel_withdrawal(
+    // Resolve the active version's package so the cancel runs the bytecode
+    // generation whose committed-request gate matches the chain's state
+    // (v1's bag-membership gate misses v2 in-place-committed requests).
+    let call_package = crate::onchain::OnchainState::new_reader(
```
dlukegordon: This resolver only fixes cancellation. Single-request creation still passes `hashi_ids.package_id` at line 100, and batch creation builds `SuiTxExecutor::new` without on-chain state at line 141, so it also falls back to the original package. Once v1 is disabled—the required safe cutover—both request paths abort with `EVersionDisabled`. Please resolve the active package fail-closed and use it for both creation paths too.

dlukegordon (commented):
Two additional findings from the current head review.

crates/hashi/src/cli/commands/withdraw.rs
Bridgerz: Fixed in b061b5ad2: resolution errors now propagate and there is no fallback to the original package; all four CLI submission paths share one fail-closed resolver.

crates/hashi/src/metrics.rs
Bridgerz: Fixed in b061b5ad2: hot Processing/Signed requests whose transaction is confirmed in the mirror now classify as confirmed_pending_archive; the Confirmed arm stays as a labeled-defensive fallback.

crates/hashi/src/cli/commands/withdraw.rs
Bridgerz: Fixed in b061b5ad2: the list view partitions on active status, so committed-in-place requests report separately from queued and the counts reconcile.

crates/hashi/src/leader/garbage_collection.rs
Bridgerz: Fixed in b061b5ad2: when disarmed, the check probes the mirror for confirmed unarchived txns and arms; the utxo_cleanup flag had the identical hole and got the same fix.

crates/hashi/src/withdrawals.rs
Bridgerz: Implemented in 32c1a0ab5: withdrawal semantics resolve an effective version that stays 1 while version 1 remains enabled, threaded together through routing, caps, coefficients, the validator shape check, the scheduler threshold, and the archival GC gate. Inert under exclusive upgrades and tagged TODO(pre-mainnet-wipe). The default e2e boot now disables v1 and asserts the gate opened; the direct-v1-entries test reorders disablement before the commit accordingly.

crates/hashi/src/config.rs
Bridgerz: Fixed in b061b5ad2 with your min() and a unit test pinning the version-resolved threshold; 32c1a0ab5 then routes it through the effective version so dormancy and readiness agree.

crates/hashi/src/withdrawals.rs
Bridgerz: You were right and it was worse: the negative requests-bag probe is also cached, so a single 447 walk broke the cap alone. 2b684904b drops that probe (a committed request lives in exactly one bag, so processed membership plus Confirmed is equivalent) landing at two objects per request, and b061b5ad2 raises the planner coefficient to 2 with a two-max-victim planning test.

crates/hashi/src/cli/commands/withdraw.rs
Bridgerz: Fixed in b061b5ad2: single and batch creation route through the same fail-closed resolver (the batch executor now carries on-chain state and self-routes); hashi deposit had the identical bug and is fixed in the same commit. Note the tradeoff: creation now requires a successful governance read where it previously needed none.

crates/hashi/src/withdrawals.rs
dlukegordon: The off-chain effective-version routing fixes the normal node path, but the unsafe state is still reachable on-chain while {1,2} are enabled. `commit_withdrawal_tx` only checks that the executing package version is enabled, and `WithdrawalCommitmentMessage` does not bind a package/version. A valid commitment certificate can therefore be submitted directly through v2 during dormancy, creating the in-place layout while v1 cancel/approve remain callable. Please enforce the dormancy gate in the Move withdrawal entries (and add a direct-v2-commit regression test); this thread is not resolved yet.

Bridgerz: Dug into this more. Agree the node-side gate alone isn't an on-chain guarantee, but I don't want to enforce dormancy in Move: we're squashing and republishing the whole package right before mainnet, so it would be scaffolding we immediately delete. The replay vector needs a v2 commit to land while v1 is still enabled, and commit/approve are pause-gated in both packages while governance execution is not. So the rollout is: pause, execute the upgrade, execute DisableVersion(1), unpause, back to back. While paused no commit can execute through either package, so there's no state for v1's cancel to destroy and no window where a commitment cert can be replayed through v2. Once v1 is disabled, replaying a cert through v2 is just the normal commit path. The queue doesn't need to drain either: committed in-flight withdrawals ride through (finalize/confirm/archival all handle requests in the v1 locations, and we have an e2e covering commit-and-sign under v1 then confirm through v2). Node dormancy stays as the belt for the rollout; the pause is the on-chain suspenders.

design/docs/move-upgrades.mdx
```diff
@@ -120,27 +120,41 @@ an exclusive publication causes old binaries to stop writing.
...
-
-1. Avoid active withdrawals during the rollout window.
-2. Create and fully vote both the v2 upgrade proposal and a
-   `DisableVersion(1)` proposal while v1 is active. Proposal creation permits
-   targeting the current version; only execution rejects disabling it.
+race window for this one migration. Run the migration under the emergency
+pause, which closes the withdrawal-replay surface described below; the
+withdrawal queue does not need to drain first.
+
+1. Create and fully vote the pause, v2 upgrade, `DisableVersion(1)`, and
```
dlukegordon: Fully approving the unpause before this window makes the pause externally revocable. `emergency_pause::execute` is public, and proposal execution only checks quorum, so any account can execute the unpause and a direct v2 commitment in one PTB after v2 is published but before v1 is disabled. Please create and vote the unpause only after `DisableVersion(1)` has finalized and been verified.

crates/hashi/src/cli/commands/withdraw.rs
```diff
@@ -423,31 +528,28 @@ async fn list(config: &CliConfig, output_format: OutputFormat) -> Result<()> {
...
                 }
 
                 if !pending.is_empty() {
-                    if !requests.is_empty() {
+                    if !queued.is_empty() {
                         println!();
                     }
                     println!("  {}", "Pending Broadcast:".bold().underline());
                     for pw in &pending {
                         let txid: bitcoin::Txid = pw.txid.into();
```
dlukegordon, line 549: Confirmed transactions remain in this collection until archival, so this prints `Confirmed (archival pending)` under `Pending Broadcast`. That is contradictory and can make completed withdrawals look actionable. Please rename the section to something neutral such as `Withdrawal Transactions`, or split confirmed rows into an awaiting-archival section.

dlukegordon (changes requested):
Most of the previous findings are addressed. I’m requesting changes for one remaining issue: the pause-based migration control is sound in principle, but the runbook pre-approves a permissionlessly executable unpause, allowing the unsafe mixed-version window to be reopened. Once the runbook creates and votes the unpause only after v1 is disabled and verified, I’m comfortable with the procedural mitigation for this explicitly testnet-only migration.

design/docs/move-upgrades.mdx
Bridgerz: Good catch. Quorum-complete proposals are executable by anyone, so a pre-voted unpause would make the pause revocable inside the window. Reordered the doc: pause, upgrade, and disable are pre-voted; the unpause is created and voted only after DisableVersion(1) is verified on chain. The bridge stays paused through the unpause vote, which we accept for this one migration.

crates/hashi/src/cli/commands/withdraw.rs
Bridgerz: Renamed the section to Withdrawal Transactions. Each row already shows its precise status, and the summary line separates confirmed-awaiting-archival.

design/docs/move-upgrades.mdx
Bridgerz: Follow-up on this one. After walking through ceremony logistics we're going to relax this on testnet and pre-vote the unpause with the other proposals. You're right that execution is permissionless at quorum, so the exposure is real in principle, but on testnet nothing of value is at stake, the window between upgrade and disable is seconds wide, and in practice every proposal we've run has been executed by us. We'll re-verify the pause is still active immediately before executing the upgrade and have the disable transaction pre-built so the window stays minimal. The doc now documents the strict ordering (unpause created and voted only after DisableVersion(1) is verified) as the required procedure whenever real value is at stake, with the testnet relaxation and its rationale called out explicitly. Mainnet starts from the exclusive upgrade path, so this two-transaction window should never exist there at all.

dlukegordon: The CLI issue is addressed. The rollout blocker is not: a89788b50f56 fixed it, but the latest commit explicitly restores the pre-voted unpause for testnet. Rechecking the pause before publication and pre-building DisableVersion(1) only shorten the unsafe window; they do not prevent anyone from atomically executing the approved unpause and a direct v2 commitment before disable-v1 lands. Please make the strict ordering in steps 1–7 the actual testnet procedure and remove the pre-vote exception at lines 146–149.

Bridgerz: For testnet we have had to explicitly execute proposals ourselves everytime. this won't be a problem again after this first upgrade (thanks to exclusive upgrades). @bmwill are you good with this plan? essentially the risk is someone executes the proposals out of order and causes a potential for someone to cancel their withdrawal, and brick their utxo (very low likelyhood and this is a one off thing and its testnet... lets ship it)

crates/hashi/src/withdrawals.rs
dlukegordon: That scope changes my assessment. The unsafe mixed-version state remains technically reachable, but given that this is an explicitly accepted testnet-only risk with no real value, the strict procedure is documented for any value-bearing migration, and mainnet uses the atomic exclusive path, I’m okay with avoiding temporary Move scaffolding here. I’m treating this as accepted risk rather than a code fix.

design/docs/move-upgrades.mdx
dlukegordon: Given the explicit testnet-only scope, no real value at risk, the short controlled window, and the atomic exclusive path for mainnet, I’m okay treating this as consciously accepted operational risk rather than requiring the strict ceremony here.

dlukegordon (approved):
Approved with the mixed-version rollout exposure explicitly accepted as a one-off, testnet-only operational risk. Mainnet uses the atomic exclusive upgrade path.

jessiemongeon1:
## Style Guide Audit

### `design/docs/move-upgrades.mdx` (3 violation(s))

_3 violation(s) (0 regex, 3 claude)_

- **Line 159** — passive_voice
  - Current: `both versions are enabled a live certificate can be replayed through v2's`
  - Fix: `both versions are enabled an attacker can replay a live certificate through v2's`
- **Line 164** — passive_voice
  - Current: `referencing withdrawal transaction forever. Every corrupting interleaving`
  - Fix: `referencing withdrawal transaction forever. Every corrupting interleaving requires a v2 commit to land while v1 remains enabled.`
- **Line 165** — passive_voice
  - Current: `requires such a v2 commit to land while v1 is still enabled.`
  - Fix: `requires such a v2 commit to land while v1 remains enabled.`

---

PR #1028, opened 2026-08-18
Move the MPC RPC handlers off the worker threads

zhouwfang:
#16

### DIFF
--- a/crates/hashi/src/mpc/rpc/service.rs
+++ b/crates/hashi/src/mpc/rpc/service.rs
@@ -47,21 +47,24 @@ impl MpcService for HttpService {
             .with_label_values(&[label])
             .start_timer();
         let response = spawn_blocking(move || -> Result<_, Status> {
-            let mut mgr = mpc_manager.write().unwrap();
-            validate_epoch(mgr.mpc_config.epoch, external_request.epoch)?;
-            mgr.handle_send_messages_request(sender, &internal_request)
-                .map_err(|e| {
-                    if matches!(&e, MpcError::NotReady(_)) {
-                        tracing::info!("send_messages from {sender:?}: {e}");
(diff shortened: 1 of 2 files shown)


crates/hashi/src/mpc/rpc/service.rs
```diff
@@ -73,30 +76,32 @@ impl MpcService for HttpService {
...
-            mgr.handle_retrieve_messages_request(requester, &internal_request)
-                .map_err(|e| {
-                    match &e {
-                        MpcError::NotFound(_) => {
-                            tracing::debug!("retrieve_messages: {e}");
-                        }
-                        _ => {
-                            tracing::warn!("retrieve_messages failed: {e}");
+        let mpc_manager = self.mpc_manager()?;
+        let response = spawn_blocking(move || -> Result<_, Status> {
```
dlukegordon, line 80: This is specifically fixing async-runtime starvation, but the existing manager tests bypass the RPC service and cannot catch a regression here. Please add a service-level test with a blocked storage read on a single-thread Tokio runtime and verify another task still gets polled.

dlukegordon (commented):
Non-blocking follow-up: consider adding an MPC RPC concurrency limit in the future. The handlers now use Tokio's shared blocking pool, so an explicit limit would provide a clearer resource envelope and protect protocol-critical blocking work from accidental request storms.

zhouwfang:
> Non-blocking follow-up: consider adding an MPC RPC concurrency limit in the future. The handlers now use Tokio's shared blocking pool, so an explicit limit would provide a clearer resource envelope and protect protocol-critical blocking work from accidental request storms.

Replied offline: "Yes, and also need to add a server side deadline"

dlukegordon (approved):

PR #1032, opened 2026-08-18
Stop holding the manager lock across MPC RPC storage reads

zhouwfang:
MystenLabs/hashi#918

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -448,6 +448,17 @@ impl MpcManager {
         requester: Address,
         request: &RetrieveMessagesRequest,
     ) -> MpcResult<RetrieveMessagesResponse> {
+        match self.try_retrieve_in_memory(requester, request)? {
+            Some(response) => Ok(response),
+            None => retrieve_from_store(&*self.public_messages_store, request),
+        }
+    }
+
+    pub fn try_retrieve_in_memory(
+        &self,
+        requester: Address,
+        request: &RetrieveMessagesRequest,
+    ) -> MpcResult<Option<RetrieveMessagesResponse>> {
         if request.epoch == self.mpc_config.epoch
             && let Some(messages) = self.get_dealer_messages(
                 request.protocol_type,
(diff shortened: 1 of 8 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -6637,6 +6603,130 @@ fn consume_certified_nonce_outputs<T>(
...
+    requester: Address,
+    common: Lookup<batch_avss_avid::AvssCommonMessage>,
+    held: Lookup<HeldAvidEchoes>,
+}
+
+pub(crate) fn finish_avid_retrieval(
+    store: &dyn PublicMessagesStore,
+    pending: AvidPending,
+) -> MpcResult<RetrieveMessagesResponse> {
+    let common = match pending.common {
```
dlukegordon, line 6636: Can we read `held` before `common` here? If both are initially missing, a writer can persist common and held between these reads, causing us to return a vote without its common message. The consumer rejects that bundle. Since common is persisted before held, reading held first preserves a consistent response. Please add a race test that pauses between these fallback reads.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -6637,6 +6603,130 @@ fn consume_certified_nonce_outputs<T>(
...
+                        ..
+                    }) => echo.clone(),
+                    _ => unreachable!("held echoes are echo messages"),
+                })
+            });
+            (Some(vote.clone()), echo)
+        }
+        None => (None, None),
+    };
+    if common.is_none() && avid_vote.is_none() && echo.is_none() {
```
dlukegordon, line 6674: The reordered reads fix the publication race, but this can still serve a vote without common after a partial write. `cache_and_persist_avid_round_state` caches common before persisting it, so if that write fails, a later dispersal can use the cached state and successfully persist held state. After restart, only held exists and this returns a vote-only bundle that the consumer rejects. Can we reject or suppress held whenever common is absent, and add a partial-write/restart test?

dlukegordon (approved):
Approving with one remaining correctness follow-up.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: PTAL at #1034 though on the write side rather than the read side.

PR #1035, opened 2026-08-19
Return an error when a dealer misses its certificate quorum

zhouwfang:
MystenLabs/hashi#918


dlukegordon (approved):
Non-blocking: the key rotation, vanilla nonce generation, and AVID pessimistic Vote shortfall branches are also changed to return `NotEnoughApprovals`, but the updated tests do not exercise them. Please consider adding coverage for those paths and matching the error explicitly so `needed` and `got` are verified.

PR #1037, opened 2026-08-20
Stop rotation reconstruction walking past its threshold

zhouwfang:
#95

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -5392,6 +5392,9 @@ impl MpcManager {
         let mut local_outputs: HashMap<ShareIndex, avss::AvssOutput> = HashMap::new();
         let mut certified_share_indices = Vec::new();
         for cert in certificates {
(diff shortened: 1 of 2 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -5392,6 +5392,9 @@ impl MpcManager {
         let mut local_outputs: HashMap<ShareIndex, avss::AvssOutput> = HashMap::new();
         let mut certified_share_indices = Vec::new();
         for cert in certificates {
+            if certified_share_indices.len() >= context.input_threshold as usize {
```
dlukegordon, line 5395: The threshold check also needs to happen inside the `rotation_msgs` loop. A certificate can contain multiple weighted shares, so we may reach the threshold partway through a dealer batch and still fail on a later unnecessary share. Please stop the inner loop once `certified_share_indices.len()` reaches `input_threshold`, and add a test where an extra share in the threshold-crossing dealer is unusable.

dlukegordon (approved):

PR #1039, opened 2026-08-20
Derive `t` from `f`

zhouwfang:
MystenLabs/hashi#368

### DIFF
--- a/crates/hashi-types/src/move_types/mod.rs
+++ b/crates/hashi-types/src/move_types/mod.rs
@@ -366,15 +366,11 @@ pub enum ConfigValue {
 
 /// MPC parameter keys, in the canonical order Move's `mpc_config::pin` writes
 /// them. Load-bearing for [`Config::from_mpc_params`].
-const KEY_MPC_THRESHOLD_IN_BASIS_POINTS: &str = "mpc_threshold_in_basis_points";
 const KEY_MPC_WEIGHT_REDUCTION_ALLOWED_DELTA: &str = "mpc_weight_reduction_allowed_delta";
 const KEY_MPC_MAX_FAULTY_IN_BASIS_POINTS: &str = "mpc_max_faulty_in_basis_points";
 const KEY_MPC_NONCE_GENERATION_PROTOCOL: &str = "mpc_nonce_generation_protocol";
 const KEY_MPC_NONCE_ACCUMULATION_WINDOW_MS: &str = "mpc_nonce_accumulation_window_ms";
 
-/// Default MPC threshold in basis points. Mirrors `DEFAULT_THRESHOLD_IN_BASIS_POINTS`
-/// in `mpc_config.move`.
-pub const DEFAULT_MPC_THRESHOLD_IN_BASIS_POINTS: u16 = 3334;
 /// Mirrors `DEFAULT_WEIGHT_REDUCTION_ALLOWED_DELTA` in `mpc_config.move`.
 pub const DEFAULT_MPC_WEIGHT_REDUCTION_ALLOWED_DELTA: u16 = 800;
 /// Mirrors `DEFAULT_MAX_FAULTY_IN_BASIS_POINTS` in `mpc_config.move`.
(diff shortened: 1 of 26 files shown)


jessiemongeon1:

crates/hashi-types/src/move_types/mod.rs
```diff
@@ -436,17 +434,12 @@ impl Config {
...
         nonce_generation_protocol: u16,
         nonce_accumulation_window_ms: u64,
     ) -> Self {
         Self(vec![
-            (
-                KEY_MPC_THRESHOLD_IN_BASIS_POINTS.to_string(),
-                ConfigValue::U64(threshold_in_basis_points as u64),
-            ),
             (
                 KEY_MPC_WEIGHT_REDUCTION_ALLOWED_DELTA.to_string(),
```
benr-ml, line 444: don't we need to make this one also configurable?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -6749,31 +6752,97 @@ fn build_reduced_nodes(
...
+        .map(|value| -> MpcResult<u16> {
+            match value {
+                hashi_types::move_types::ConfigValue::U64(bps) => {
+                    u16::try_from(*bps).map_err(|_| {
+                        MpcError::CryptoError(format!(
+                            "pinned mpc_threshold_in_basis_points {bps} exceeds u16::MAX"
+                        ))
+                    })
+                }
+                other => Err(MpcError::CryptoError(format!(
```
benr-ml: nit - it's not a crypto error (also the case above)

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -6749,31 +6752,97 @@ fn build_reduced_nodes(
...
+                    as u16,
+                weight_reduction_allowed_delta_in_basis_points,
+            ),
+            None => {
+                let max_faulty = (total_weight as u32 * max_faulty_in_basis_points as u32
+                    / MAX_BASIS_POINTS)
+                    .max(1);
+                let threshold = (total_weight as u32).saturating_sub(2 * max_faulty);
+                if threshold <= max_faulty {
+                    return Err(MpcError::CryptoError(format!(
```
benr-ml: when we limit f by 1/3 as you did above, this should never happen, no? if so, why return error and not panic?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -6749,31 +6752,97 @@ fn build_reduced_nodes(
...
+                other => Err(MpcError::CryptoError(format!(
+                    "pinned mpc_threshold_in_basis_points is not a u64: {other:?}"
+                ))),
+            }
+        })
+        .transpose()?;
+    let (threshold, max_faulty, weight_reduction_allowed_delta) =
+        match legacy_threshold_in_basis_points {
+            Some(threshold_in_basis_points) => (
+                (total_weight as u32 * threshold_in_basis_points as u32).div_ceil(MAX_BASIS_POINTS)
```
benr-ml, line 6776: just confirming that here we didnt change the current calculation at all, right?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -6749,31 +6752,97 @@ fn build_reduced_nodes(
...
+    let (nodes, reduced_threshold, reduced_max_faulty) = Nodes::prop_reduce(
         nodes_vec,
         threshold,
         max_faulty,
         weight_reduction_allowed_delta,
         lower_bound,
     )
-    .map_err(|e| MpcError::CryptoError(e.to_string()))
+    .map_err(|e| MpcError::CryptoError(e.to_string()))?;
+    if u32::from(reduced_threshold) + u32::from(reduced_max_faulty)
```
benr-ml: seems redundant - why do we need this check?

crates/hashi-types/src/move_types/mod.rs
zhouwfang: Allowed delta is already configurable.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Yes, it shouldn't happen. The only way in is the .max(1) floor when W <= 3, which is test-only.

I kept it as an error rather than a panic because the input is a pinned on-chain committee, so it is remote data rather than a local invariant. Every node reads the same bytes, so a panic would fail the whole fleet at once. The check just below turns prop_reduce's assert into an error for the same reason.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Yes. It has to stay verbatim since prop_reduce derives its divisor from t, f and delta together, so any change would reshape the share layout of already-pinned committees.

benr-ml (approved):

PR #1040, opened 2026-08-20
move: add atomic package upgrade policy

Bridgerz:
## Motivation

Publishing a Sui package upgrade does not make the previous package uncallable. Hashi package versions share the same mutable `Hashi` object, so an old entry function can violate an invariant introduced by a newer package.

#1003  exposed a concrete case: a request committed through v2 can remain in a location that v1 considers cancellable. If v1 remains enabled, its cancellation path can destroy the committed request and potentially strand burned hBTC and locked UTXOs.

Adding `exclusive` directly to the deployed `upgrade::Upgrade` struct would change the BCS layout of existing proposal objects. This PR therefore introduces a temporary `upgrade_v2` proposal type for testnet and leaves the legacy type unchanged.

## What changed

- Add `upgrade_v2::Upgrade { digest, exclusive }`.
- Bind the approved policy to publication with a no-abilities `UpgradeAuthorization` returned alongside the Sui `UpgradeTicket`.
- Finalize publication and the version policy in the same PTB:
  - `exclusive = true` replaces the enabled set with only the newly committed version.
  - `exclusive = false` deliberately preserves older enabled versions.
- Require callers to provide `--exclusive true|false` when creating an `upgrade-v2` proposal.
- Resolve proposal type arguments from the package version that introduced the type, so voting, vote removal, decoding, and garbage collection work for both legacy and v2 proposal payloads.
- Add Move coverage for both policies and an e2e test that bootstraps v1 -> v2 through the legacy path, publishes v3 through `UpgradeV2(exclusive = true)`, verifies only v3 remains enabled, and verifies the v1+v2 binary halts autonomous writes.
- Add a Move-upgrade guide requiring every future upgrade to classify itself as exclusive or non-exclusive, plus operator-runbook and governance documentation.

## Rollout contract

For a future exclusive upgrade:

1. Create the proposal for the exact Move package digest.
2. Release a binary that understands both the current and proposed packages.
3. Restart validators individually and verify healthy operation on the current package before voting.
4. Confirm upgraded validators represent at least two thirds of committee **weight**, not merely node count.
5. Reach vote quorum, then manually execute/publish. Votes do not auto-execute an upgrade.
6. Publication atomically enables only the new version. Old binaries observe that they support no enabled package and halt writes.

A non-exclusive upgrade is allowed only when review and cross-version tests establish that old and new entry functions can safely mutate the same shared state. Binary compatibility alone is not sufficient justification.

## Current testnet v1 -> v2 plan

The deployed v1 package cannot use `UpgradeV2`, so the first transition necessarily has a small two-transaction gap. Testnet has no real funds, and we accept that risk for this migration:

1. Keep withdrawals quiet during the rollout.
2. Release the v1+v2-capable binary and upgrade validators representing at least two thirds of committee weight.
3. Create and fully vote the exact-digest legacy upgrade proposal and `DisableVersion(1)` proposal while v1 is active.
4. Execute/publish/finalize v2 through the legacy upgrade proposal and wait for success and the new v2 package ID.
5. Immediately execute the already-approved disable-v1 proposal through v2.
6. Verify v1 calls fail with `EVersionDisabled` and normal routing uses v2.

The two executions must not be submitted concurrently because their shared-object ordering is not guaranteed. Emergency pause is not a complete substitute because the legacy v1 withdrawal-cancellation path does not honor it.

## Before mainnet

Delete the original legacy `upgrade` module, rename `upgrade_v2` to `upgrade`, and squash the testnet-only package-version and storage assumptions before the first mainnet publish. The new module includes a source-level TODO recording this requirement.

## Verification

Passed locally:

- `make fmt-move`
- `make check-fmt`
- `make buf-lint`
- `make proto` (no generated drift)
- `RUSTFLAGS=-Dwarnings make clippy`
- `sui move build --path packages/hashi -e testnet`
- `sui move test --path packages/hashi -e testnet` (157 passed)
- `cargo test --locked -p hashi --test move_upgrade_compat -- --nocapture` (9 passed)
- `RUSTFLAGS=-Dwarnings cargo test --workspace --all-features --doc`
- `RUSTFLAGS=-Dwarnings make doc`
- `cd design && npm ci && npm run build`
- `make -C docker/hashi-guardian enclave-lock`
- `RUSTFLAGS=-Dwarnings cargo nextest run -p e2e-tests --test upgrade_tests test_upgrade_v2_exclusive_via_proposal --profile ci`
- `test_drain_mode_max_request_batch` in the aggregate nextest run
- `make is-dirty`

The full local `cargo nextest run --workspace --all-features --profile ci` did not finish green. Under aggregate load, `test_nonce_accumulation_window_open` observed stale enabled-version state (it passed when rerun in isolation), and `test_large_withdrawal_signature_chunking` timed out. The latter progressed further in isolation but later hit `ENotReconfiguring` in committee handoff and timed out waiting for `WithdrawalSigned`; I believe these are local contention/existing e2e failures, but did not reproduce them on `main` to verify that conclusion.

Local tools differ from pinned CI: Sui 1.76.1 locally vs 1.73.2 in Move CI, and Node 22 locally vs Node 20 in docs CI. Pinned GitHub CI should be treated as authoritative for those environments.

### DIFF
--- a/crates/e2e-tests/src/upgrade_flow.rs
+++ b/crates/e2e-tests/src/upgrade_flow.rs
@@ -14,6 +14,7 @@ use hashi::cli::client::build_vote_transaction;
 use hashi::cli::upgrade::build_execute_proposal_transaction;
 use hashi::cli::upgrade::build_upgrade_execution_transaction;
 use hashi::cli::upgrade::build_upgrade_package;
+use hashi::cli::upgrade::build_upgrade_v2_execution_transaction;
 use hashi::cli::upgrade::extract_new_package_id_from_response;
 use hashi::cli::upgrade::extract_proposal_id_from_response;
 use hashi::config::HashiIds;
@@ -290,10 +291,36 @@ fn hex_encode(bytes: &[u8]) -> String {
     bytes.iter().map(|b| format!("{b:02x}")).collect()
 }
 
-/// Run the full upgrade lifecycle: prepare → build → propose → vote → execute+publish+finalize.
+#[derive(Clone, Copy)]
+enum UpgradeProposal {
+    Legacy,
+    V2 { exclusive: bool },
+}
+
+/// Run the full legacy upgrade lifecycle.
 ///
+/// This is the only path available for the deployed v1 → v2 transition.
 /// Returns the new package ID on success.
 pub async fn execute_full_upgrade(networks: &mut TestNetworks) -> Result<Address> {
+    execute_full_upgrade_with_proposal(networks, UpgradeProposal::Legacy).await
+}
+
+/// Run the full `upgrade_v2` lifecycle with an explicit version policy.
+///
+/// The chain must already have package v2, which introduced the proposal
+/// payload type. Returns the new package ID on success.
+pub async fn execute_full_upgrade_v2(
+    networks: &mut TestNetworks,
+    exclusive: bool,
+) -> Result<Address> {
+    execute_full_upgrade_with_proposal(networks, UpgradeProposal::V2 { exclusive }).await
+}
+
+/// Run prepare → build → propose → vote → execute+publish+finalize.
+async fn execute_full_upgrade_with_proposal(
+    networks: &mut TestNetworks,
+    proposal: UpgradeProposal,
+) -> Result<Address> {
     // Running nodes only: a pending member (started mid-test by the
     // key-rotation tests) has no Hashi instance to read state from or vote
     // with — and is not a registered committee member yet anyway.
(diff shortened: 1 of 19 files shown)


jessiemongeon1:

bmwill (commented):
So does this mean we have to upgrade to a version with this and then do another upgrade with the other changes we want to make? i guess alternatively we can add logic to avoid calling the new code path while v1 is enabled still?

bmwill (approved):

PR #1044, opened 2026-08-25
Epoch-key the rotation complaint map

zhouwfang:
#95

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -1797,9 +1797,13 @@ impl MpcManager {
                     if share_indices.iter().any(|idx| {
                         !mgr.dealer_outputs
                             .contains_key(&DealerOutputsKey::Rotation(dealer, *idx))
-                            && !mgr
-                                .complaints_to_process
-                                .contains_key(&ComplaintsToProcessKey::Rotation(dealer, *idx))
+                            && !mgr.complaints_to_process.contains_key(
+                                &ComplaintsToProcessKey::Rotation {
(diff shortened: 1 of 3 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -4826,9 +4838,11 @@ impl MpcManager {
...
             .iter()
             .filter_map(|(key, complaint)| match key {
-                ComplaintsToProcessKey::Rotation(d, share_index) if d == dealer => {
-                    Some((*share_index, complaint.clone()))
-                }
+                ComplaintsToProcessKey::Rotation {
+                    epoch: key_epoch,
+                    dealer: d,
+                    share_index,
+                } if *key_epoch == epoch && d == dealer => Some((*share_index, complaint.clone())),
```
dlukegordon, line 4845: This can now select a previous-epoch complaint, but the receiver below still uses the current encryption key. Use `encryption_key_for_epoch(epoch)` here and test with different current and previous keys; otherwise previous-epoch rotation recovery fails after a real key change.

dlukegordon (commented):
Review finding

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: I don't think it was actually broken though. That receiver only verifies and recovers, neither reads the key. Skipped the key test for the same reason. Left handle_complain_request alone. Same change there would break nodes that weren't in the previous committee.

dlukegordon (approved):

PR #1045, opened 2026-08-25
Drop partial sigs from peers reporting a different signing nonce

zhouwfang:
#179


crates/hashi/src/mpc/signing.rs
```diff
@@ -1016,11 +1070,22 @@ impl SigningManager {
...
+                    let mut nonce_mismatches = 0u64;
                     for (signing_id, sigs) in response.partial_sigs {
                         if let Some(&i) = index.get(&signing_id)
                             && pending[i].peers_remaining.remove(&peer)
                         {
                             let st = &mut pending[i];
+                            let reported = response
+                                .signing_nonces
+                                .get(&signing_id)
+                                .filter(|n| n.len() == POINT_SIZE_IN_BYTES);
```
dlukegordon: If a response includes partials but omits this nonce or sends the wrong length, `reported` becomes `None` and we still merge the partials below. A peer signing under another nonce can therefore bypass this check and send us back into failed aggregation/`TooManyInvalidSignatures`. Please accept partials only when an exactly sized nonce is present and matches ours; missing or malformed values can be logged separately as unverifiable.

zhouwfang: Wrong-length values are rejected now, and so is a peer that reports nonces for some ids but skips others, since the responder always fills both maps together and a partial map can't happen honestly. What I left permissive is a peer sending no nonces at all, because that's just an un-upgraded node, and dropping those shares would put us under threshold on every rolling deploy rather than only this one.

dlukegordon (approved):

PR #1048, opened 2026-08-27
Select oldest confirmed utxo for withdrawals

dlukegordon:
Addresses [certora audit issue I-04](https://linear.app/mysten-labs/issue/IOP-630/mpc-only-recovery-path-becomes-available-while-the-guardian-is-still). The goal is to select old utxos to ensure none of them age past 60 days during normal operation.

This PR starts with a very simple naive solution, just adding the oldest utxo to each withdrawal tx as an input. If we want a more sophisticated solution we can add it as a subsequent PR.

### DIFF
--- a/crates/hashi/src/utxo_pool/mod.rs
+++ b/crates/hashi/src/utxo_pool/mod.rs
@@ -258,6 +258,13 @@ pub struct CoinSelectionParams {
     /// Step 3.
     pub max_inputs: usize,
 
+    /// UTXO that must be included in selection.
+    ///
+    /// An absent or ineligible required input causes selection to fail. When
+    /// eligible, it is selected before the otherwise largest-first funding
+    /// inputs and counts toward [`Self::max_inputs`].
+    pub required_input: Option<UtxoId>,
+
     /// Maximum number of withdrawal requests that may be batched into a single
     /// transaction.
     pub max_withdrawal_requests: usize,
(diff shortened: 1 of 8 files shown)


crates/hashi/src/utxo_pool/mod.rs
```diff
@@ -258,6 +258,13 @@ pub struct CoinSelectionParams {
     /// Step 3.
     pub max_inputs: usize,
 
+    /// UTXO that must be included in selection.
+    ///
+    /// An absent or ineligible required input causes selection to fail. When
+    /// eligible, it is selected before the otherwise largest-first funding
+    /// inputs and counts toward [`Self::max_inputs`].
+    pub required_input: Option<UtxoId>,
```
bmwill: I think this actually presents a worse issue, we can end up not being able to actually satisfy the withdrawal. I think a better setup would be do the following:

- selection coins from the larget-first in order to fund the txn
- with the space left over in our input budget, sort the elegible utxos by remaining time (if they are <30d left) otherwise select the smallest like we do today

crates/hashi/src/utxo_pool/mod.rs
```diff
@@ -283,33 +283,23 @@ pub struct CoinSelectionParams {
...
+    /// The fee rate threshold at which the algorithm performs no pool
+    /// consolidation. `max_fee_per_request` is derived from this rate applied
+    /// to the worst-case transaction shape.
     pub high_fee_rate_threshold: FeeRate,
 
-    /// Maximum number of extra consolidation inputs per request when the
-    /// fee rate is at or above `long_term_fee_rate`. At high fee rates,
-    /// consolidation is capped at `input_budget × N` extra inputs; below
-    /// the long-term rate the cap is lifted (bounded only by `max_inputs`).
-    /// Matches the on-chain `input_budget` config parameter.
```
bmwill, line 305: i don't think we need to edit any of these comments

bmwill (approved):
lgtm with the nit to fix the edited comments

PR #1052, opened 2026-08-27
Fix nondeterministic sui test port collisions

dlukegordon:
Serializes Sui genesis and startup across e2e test processes, while keeping the test bodies parallel. Sui startup port conflicts now fail fast, terminate the partial Sui process, regenerate genesis, and retry.

Caps e2e concurrency at four tests independently of runner CPU count, and makes the resignation lifecycle test globally exclusive on every runner size.

### DIFF
--- a/crates/e2e-tests/src/sui_network.rs
+++ b/crates/e2e-tests/src/sui_network.rs
@@ -1,11 +1,14 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use anyhow::Context;
 use anyhow::Result;
 use anyhow::anyhow;
 use anyhow::bail;
 use hashi::config::get_available_port;
 use std::collections::BTreeMap;
+use std::fs::File;
+use std::fs::OpenOptions;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Child;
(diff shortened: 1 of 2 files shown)


0xsiddharthks (approved):

0xsiddharthks (approved):

crates/e2e-tests/src/sui_network.rs
```diff
@@ -166,20 +268,60 @@ impl SuiNetworkBuilder {
...
-            user_keys,
-            admin_ports,
-        } = load_keys(&dir)?;
 
-        let rpc_port = self.rpc_port.unwrap_or_else(get_available_port);
-        let process = self.start_network(&dir, rpc_port)?;
-
-        let rpc_url = format!("http://127.0.0.1:{rpc_port}");
+        let (process, rpc_url, client, validator_keys, user_keys, admin_ports) = {
+            // Sui genesis probes ports and releases them before `sui start` binds. Serialize that
```
zhouwfang, line 273: The lock only covers Sui, but bitcoin_node and hashi_network still pick ports unserialized from the same ephemeral pool. The retry absorbs those, so not a bug, but this comment reads like the gap is closed.

zhouwfang (commented):
FYI, the CI tests have already taken 25 minutes.

bmwill (changes requested):
serializing this does not really work as the OS already doesn't give out a port it previously did until after a timeout period and so we likely shouldn't merge this given the vast increase in runtime

dlukegordon:
> serializing this does not really work as the OS already doesn't give out a port it previously did until after a timeout period and so we likely shouldn't merge this given the vast increase in runtime

Yeah definitely not merging this as is, working on a better fix

Issue #882, opened 2026-07-24
security(governance): `UpdateConfig` replaces the fixed guardian BTC key used for future 2-of-2 deposit address derivation


alexbelij:
**Component:** `hashi::update_config` (`execute`) + `hashi::config` (`is_valid_config_update`, `set_guardian_btc_public_key`)
**Type:** Security — governance bypass of a documented, code-enforced immutability invariant
**Security impact:** A passing `UpdateConfig` proposal (6667 bps of committee weight) can replace the on-chain guardian BTC key even though the dedicated setter and published governance documentation define that key as immutable. The replacement key is then read by the Rust client and passed to `derive_deposit_address` for future deposit-address derivation. No unprivileged path is claimed.
**Repo:** `MystenLabs/hashi`, `main` at commit `bbce0c9ae48d2da0616d7c072e53a5289b6db82f`

## Summary

`update_config::execute` writes every proposed `(key, value)` entry straight into the global `Config` via `config::upsert`, gated only by `config::is_valid_config_update` (key exists + same value type) and `mpc_config::is_valid_value` (range-checks MPC keys only):

```move
// packages/hashi/sources/core/proposal/types/update_config.move, lines 56-67
public fun execute(hashi: &mut Hashi, proposal_id: ID, clock: &Clock) {
    let UpdateConfig { entries } = proposal::execute(hashi, proposal_id, clock);
    let (keys, values) = entries.into_keys_values();
    keys.zip_do!(values, |key, value| {
        assert!(
            hashi.config().is_valid_config_update(&key, &value)
                && mpc_config::is_valid_value(&key, &value),
            EInvalidConfigEntry,
        );
        hashi.config_mut().upsert(*key.as_bytes(), value);   // writes directly, no per-key setter
    });
}
```

```move
// packages/hashi/sources/core/config/config.move, lines 99-102
public(package) fun is_valid_config_update(self: &Config, key: &String, value: &Value): bool {
    if (!self.config.contains(key)) return false;
    self.config.get(key).same_variant(value)
}
```

`guardian_btc_public_key` is a core (non-MPC) config key, so `mpc_config::is_valid_value` falls through to its `else => true` branch (lines 44-46) and performs no validation for it. Neither gate consults the dedicated setter that actually enforces the key's immutability:

```move
// packages/hashi/sources/core/config/config.move, lines 130-139
public(package) fun set_guardian_btc_public_key(self: &mut Config, btc_public_key: vector<u8>) {
[...]

PR #1062, opened 2026-08-28
move: refuse the pinned keys on the UpdateConfig path

Bridgerz:
## Summary

`UpdateConfig` proposals can no longer rewrite the two keys the package pins for a deployment's lifetime. Fixes [IOP-543](https://linear.app/mysten-labs/issue/IOP-543/securitygovernance-updateconfig-replaces-the-fixed-guardian-btc-key) (GitHub #882) and the guardian-key impact of [IOP-592](https://linear.app/mysten-labs/issue/IOP-592/update-config-accepts-any-value-for-non-mpc-governance-keys-add-a-per) (Certora, 2026-08-04); see also [SEC-547](https://linear.app/mysten-labs/issue/SEC-547/hashi-move-governance-stale-proposal-execution-and-a-write-once-config) F65 and [SEC-554](https://linear.app/mysten-labs/issue/SEC-554/hashi-governance-parameter-changes-apply-retroactively-to-in-flight).

`update_config::execute` writes every entry through `config::upsert` and never calls the dedicated setters, so a passing proposal could:

- rotate `guardian_btc_public_key`, which `set_guardian_btc_public_key` declares write-once because every 2-of-2 deposit address is derived against it;
- change `bitcoin_chain_id`, which every node pins at startup.

Both keys are now refused with a new `EProtectedConfigKey` abort, checked after the existing key-exists / same-type / MPC-range assert. The guardian key is refused even for a same-value rewrite: governance has no business writing it at all. `guardian_url` and every other key stay governable exactly as before.

## What this PR deliberately does not do

Certora also listed out-of-range values for the remaining keys (zero pause threshold, `u64::MAX` cooldown, and so on). Those are not bounded onchain by this change. An `UpdateConfig` proposal needs a supermajority of committee weight, and reviewing the proposed value is part of voting; a quorum willing to pass a harmful value can pass a harmful package upgrade just as easily, so onchain bounds would not protect against it. Documented in `governance-actions.mdx`.

## Upgrade safety

Body-only plus two new `public(package)` predicates and one error constant; no struct or public signature changes. `move_upgrade_compat` and `version_policy_tests` pass against the deployed testnet v1 snapshot.

## Tests

- 6 new Move tests in `packages/hashi/tests/update_config_protected_keys_tests.move`: guardian key rotation and same-value rewrite refused, chain id refused, a protected entry aborts a mixed proposal, guardian URL and other keys still governable
- `sui move test -e testnet`: 234 passed
- e2e `test_create_update_config_proposal` on a snapshot-v1 chain upgraded to this branch: passed (44 s)

## Follow-up

The pending epoch/instant config split ([IOP-455](https://linear.app/mysten-labs/issue/IOP-455/make-epoch-pinned-committee-config-extensible)) should route `UpdateEpochConfig` through the same governable-key check.

### DIFF
--- a/design/docs/config.mdx
+++ b/design/docs/config.mdx
@@ -211,6 +211,17 @@ The 32-byte Bitcoin chain identifier as defined by
 (the genesis block hash). Set at genesis and not updatable through the
 `UpdateConfig` proposal.
 
+### `guardian_btc_public_key`
+
+| | |
+|---|---|
+| **Type** | `bytes` |
+
+The guardian's x-only BTC public key (32 bytes), pinned at genesis. Every
+2-of-2 deposit address is derived against it, so it is write-once and not
+updatable through the `UpdateConfig` proposal. The guardian URL
+(`guardian_url`) remains governable.
+
 ## Derived values
 
 Several values are computed from the configurable parameters above rather than
--- a/design/docs/governance-actions.mdx
+++ b/design/docs/governance-actions.mdx
@@ -72,9 +72,17 @@ active version cannot be disabled, to avoid bricking the protocol.
 
 ## `UpdateConfig`
 
-Updates a protocol configuration parameter by key. Supports any config
-key-value pair (for example, deposit fee, rate limits). See
+Updates one or more protocol configuration parameters by key. See
 [Configuration](config.mdx) for the full list of keys and defaults.
+
+Every entry is validated when the proposal executes, and the whole proposal
+aborts if any entry fails: the key must already exist, the value must have the
+same type as the current entry, MPC keys must pass their range check
+(`EInvalidConfigEntry`), and the key must be governable. The guardian BTC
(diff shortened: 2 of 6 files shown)


0xsiddharthks (approved):

jessiemongeon1:
## Style Guide Audit

Audited **2** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

**2** violation(s) found. All must be fixed before merge.

### `design/docs/governance-actions.mdx` (2 violation(s))

_2 violation(s) (0 regex, 2 claude)_

- **Line 78** — passive-voice
  - Current: `Every entry is validated when the proposal executes`
  - Fix: `The system validates every entry when the proposal executes`
- **Line 84** — passive-voice
  - Current: `Values themselves are otherwise not bounded onchain`
  - Fix: `The onchain logic does not otherwise bound values`

---

PR #1068, opened 2026-08-31
cli: execute-upgrade runs the execute + publish + finalize PTB

Bridgerz:
## Summary

Adds the operator command that runs an approved upgrade proposal to completion. Closes [IOP-664](https://linear.app/mysten-labs/issue/IOP-664/cli-execute-upgrade-runs-the-execute-publish-finalize-ptb-for-an); needed for the first testnet package upgrade.

Until now `hashi proposal execute` refused upgrade proposals ("use the full upgrade flow"), `vote --execute` skipped them, and the only code that drove the transaction they need was the e2e harness, which requires a `TestNetworks` and cannot point at a live network. `sui client upgrade` cannot do it either: the `UpgradeCap` sits in Hashi custody and the `UpgradeTicket` comes out of `upgrade::execute` inside the same PTB.

```
hashi proposal execute-upgrade <proposal-id> --package-path packages/hashi [--sui-binary <path>] [--sui-client-config <client.yaml>]
```

- Reads the proposal and dispatches on its type: the legacy `upgrade` module (what the deployed v1 has) or `upgrade_v2`; any other type is refused with a pointer to `proposal execute`.
- Builds the package with `build_upgrade_package`, so the same pre-flight as `create upgrade` applies: `PACKAGE_VERSION` must be exactly +1 of the highest published version.
- One PTB: `<module>::execute` (returns the `UpgradeTicket`), `Upgrade(modules, deps, ticket)`, `<module>::finalize_upgrade` (commits the receipt and enables the new version). Built by the existing `build_upgrade_execution_transaction` / `build_upgrade_v2_execution_transaction` against the chain's latest package id.
- Honors the shared `TxOptions`: `--dry-run`, `--serialize-unsigned-transaction` for a multisig sender, `-y`.
- Prints the built digest before the confirmation prompt (the chain accepts the publish only if it matches the proposal's, so build from the same commit with the same `sui`), then the new package id and the enabled-version set read back from the chain. After a legacy upgrade it reminds the operator that the previous version stays enabled until `DisableVersion` runs through the new package.

The two refusal messages in `execute` and `vote --execute` now name the command; the runbook and `move-upgrades.mdx` do too, where they used to defer to "the release tooling".

## Tests

- `crates/hashi/tests/cli_execute_upgrade.rs`: the type dispatch (both upgrade payloads map to their module, every other payload is refused).
- Localnet acceptance run on a manual v1 localnet with this binary: register and launch four validators, `hashi launch`, DKG, `create upgrade --package-path`, three votes, `execute-upgrade`, then `create disable-version 1` executed through the new package via `vote --execute`, ending with `enabled_versions == [2]`. Both negative paths checked on the way (`execute` refuses the upgrade proposal, `execute-upgrade` refuses the disable proposal). Passed on this branch (built package digest `0x61b6c0e9…431ca1`, v2 published, `[1, 2]` then `[2]`); the run found and depends on the governance-only scrape fix this PR is stacked on.
- `cargo clippy -p hashi --all-features --all-targets` and CI's rustfmt config clean.

Stacked on the governance-only scrape fix (IOP-665): without it, every governance-only `HashiClient` from today's main panics at startup, this command included.

No Move or node change; the command runs from an operator machine.

### DIFF
--- a/crates/hashi/src/cli/commands/proposal.rs
+++ b/crates/hashi/src/cli/commands/proposal.rs
@@ -20,6 +20,11 @@ use crate::cli::print_info;
 use crate::cli::print_warning;
 use crate::cli::types::Proposal;
 use crate::cli::types::display;
+use crate::cli::upgrade::build_upgrade_execution_transaction;
+use crate::cli::upgrade::build_upgrade_package;
+use crate::cli::upgrade::build_upgrade_v2_execution_transaction;
+use crate::cli::upgrade::extract_new_package_id_from_response;
+use crate::onchain::types::ProposalType;
 
 /// Print metadata if present
 fn print_metadata(metadata: &[(String, String)]) {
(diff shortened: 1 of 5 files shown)


0xsiddharthks (approved):

jessiemongeon1:

PR #1073, opened 2026-08-31
Require all key provisioners to confirm ceremonies

dlukegordon:
Requires every key provisioner to verify and confirm its ceremony share before the guardian marks setup or rotation complete. The operator waits for all confirmations before publishing the BTC public key.


0xsiddharthks (approved):

0xsiddharthks (approved):

crates/hashi-types/src/guardian/lifecycle.rs
```diff
@@ -56,6 +56,7 @@ impl From<WithdrawStage> for EnclaveLifecycle {
 pub enum CeremonyStage {
     Uninitialized,
     OperatorInitialized,
+    AwaitingKeyProvisioners,
```
mskd12: `AwaitingKeyProvisionerConfirmations`?

crates/hashi-types/src/guardian/mod.rs
```diff
@@ -299,6 +299,68 @@ pub struct SetupNewKeyResponse {
...
+
+/// Progress returned after accepting one ceremony confirmation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CeremonyConfirmationResponse {
+    pub have: u32,
+    pub need: u32,
+    pub completed: bool,
+}
+
+impl CeremonyConfirmationRequest {
```
mskd12: nit: i prefer having struct defs in one contiguous section of the file & the impl's come later.

crates/hashi-guardian/src/ceremony_mode/confirm.rs
```diff
@@ -0,0 +1,224 @@
...
+use hashi_types::guardian::SessionBoundRequest;
+use std::sync::Arc;
+use tracing::info;
+
+pub async fn confirm_ceremony(
+    enclave: Arc<Enclave>,
+    signed: KpSigned<CeremonyConfirmationRequest>,
+) -> GuardianResult<CeremonyConfirmationResponse> {
+    let lifecycle = enclave.lifecycle();
+    if lifecycle != CeremonyStage::AwaitingKeyProvisioners.into()
```
mskd12: nit: return early when `lifecycle == CeremonyStage::Completed`?

crates/hashi-guardian/src/ceremony_mode/confirm.rs
```diff
@@ -0,0 +1,224 @@
...
+    info!(
+        share_id = share_id.get(),
+        signer_fingerprint,
+        have = status.have,
+        need = status.need,
+        "Accepted key provisioner ceremony confirmation."
+    );
+
+    if status.completed && lifecycle == CeremonyStage::AwaitingKeyProvisioners.into() {
+        enclave
```
mskd12, line 48: As discussed, we need to add an S3 log here to mark ceremony completion & all readers need to check for this marker when reading. (it's okay to perhaps do it in a separate PR tho)

mskd12 (approved):

crates/hashi-types/src/guardian/lifecycle.rs
dlukegordon: Renamed this to `AwaitingKeyProvisionerConfirmations` and updated the protobuf enum and all callers.

crates/hashi-types/src/guardian/mod.rs
dlukegordon: Moved the confirmation request/response impls into the helper impl section so the struct definitions remain contiguous.

crates/hashi-guardian/src/ceremony_mode/confirm.rs
dlukegordon: I left this as-is because returning early here would skip signature, session, ceremony-digest, and roster validation, so any request would succeed once the ceremony is complete. The `already_confirmed` return below keeps retries idempotent while still authenticating and validating the request.

crates/hashi-guardian/src/ceremony_mode/confirm.rs
dlukegordon: Tracked this in [IOP-695](https://linear.app/mysten-labs/issue/IOP-695/guardian-persist-and-enforce-ceremony-completion-markers) for a follow-up PR. The plan is to write a signed, long-lived completion marker after the final KP confirmation; completed-state readers will select by that marker, while the operator/KP ceremony flow will use an explicit session-bound pending-state read so confirmations do not deadlock.

PR #1075, opened 2026-09-01
squash reset: ship v1 from a fresh cycle and drop the deployed-package snapshots

Bridgerz:
## Summary

The base PR for the pre-mainnet cycle. Closes [IOP-656](https://linear.app/mysten-labs/issue/IOP-656/post-cut-main-fresh-cycle-reset-package-version-1-version-gates). The `testnet` maintenance branch (cut at b3786bfad) carries the deployed v1/v2 chain; main now targets the pre-mainnet release, a fresh publish with every versioned addition squashed back into v1. Breaking Move changes merge here from now on.

Two commits:

### 1. squash-day reset: the tree ships as v1 again

Everything that treated "v1 = the legacy testnet package" stops doing so:

- Move `versioning.move`: `PACKAGE_VERSION = 1`. `BARE_CERT_PACKAGE_VERSION` deleted: a new nonce bucket is always stamped; an existing bare bucket keeps taking bare writes.
- Rust: `SUPPORTED_PACKAGE_VERSIONS = [1]`, every `MoveType` `PACKAGE_VERSION` override removed, `ProposalType::package_version` routes every known type to v1, and the four feature gates (`STAMPED_NONCE_CERTS`, `DEFERRED_ARCHIVAL`, `TOB_PRUNE`, `WITHDRAWAL_ARCHIVE`) to 1. Without the gate change a fresh v1 chain dies in DKG (`ARITY_MISMATCH` on `submit_nonce_cert`).
- Follow-ups stay marked with TODOs in the source: the `withdrawal_effective_version` hold rule, the bare nonce-bucket write path, the `committee_set` `extra_fields` promotions, and deleting the legacy `upgrade` proposal type.

### 2. remove the deployed-package snapshots; the harness boots fresh

- `crates/hashi/tests/move_upgrade_snapshots/` deleted, along with the two self-tests that hard-load it and the `snapshot_signing_tests` suite (its premise was the deployed bytecode). `Published.toml` no longer records a testnet deployment: that record lives on the maintenance branch.
- The compat gate and the version-policy test go fresh-cycle: with no recorded deployment the gate has nothing to check and says so, and the policy test pins `PACKAGE_VERSION == 1`. Recording the pre-mainnet publish (Published.toml entry + new snapshot) re-arms both automatically, including the every-deployment-has-a-valid-snapshot test.
- e2e harness: the default boot is a plain fresh v1 publish of the current source (the state the pre-mainnet deployment will be in). The upgrade lifecycle stays covered as an opt-in: `with_upgraded_chain()` / `keep_v1_enabled()` publish v1 and upgrade to a const-patched v2 through the full governance flow, exactly as before minus the snapshot. The snapshot-boot upgrade tests are reworked to that scheme; their version-gate assertions are unchanged.

### 3. the exclusive-capable proposal becomes the only upgrade path

The legacy `upgrade` module is deleted and `upgrade_v2` is renamed to `upgrade` (module, Move type, Rust `ProposalType`, CLI), completing the TODO(mainnet) it carried. Every upgrade now classifies itself: `proposal create upgrade --exclusive <true|false>` (the `--exclusive true` + `--digest` brick-guard stays), and `execute-upgrade` drives the single `upgrade::execute` + `Upgrade` + `upgrade::finalize_upgrade` PTB. `versioning::commit_upgrade` carries the version policy (exclusive replaces the enabled set atomically). The `PackageUpgraded` event lands back in module `upgrade`, matching the node watcher's existing decode. e2e's `execute_full_upgrade(networks, exclusive)` is the one lifecycle helper. CI: the Move workflow's compat filter and comments no longer reference the deleted snapshot tree. Docs follow (proposal tables, governance actions, move-upgrades).

### 4. kill the v1 dormancy hold and the legacy withdrawal cost model

Per review: `withdrawal_effective_version` (the hold-at-v1-while-enabled rule, whose own TODO said it dies at the squash) is deleted, along with `LEGACY_MAX_WITHDRAWAL_REQUESTS`, the v1 commit/confirm coefficients, and the version parameter threaded through the batch-cap and input-budget model. The flow routes at the active package everywhere (`withdrawal_call_package`, `resolve_withdrawal_call_package` and the leader's dormancy observer are gone). Fresh-cycle test truths follow: the default-boot asserts pin `active == Some(1)`, the bare-only-version guard is gone, and the two migration-window e2e tests (`test_withdrawal_committed_under_v1_completes_after_upgrade`, `test_v1_entries_abort_against_v2_committed_request`) are deleted: their premise is the deployed-testnet migration, which lives on the `testnet` maintenance branch, and their post-disable phases require a binary that supports v2, which this fresh-cycle binary correctly does not (`SUPPORTED_PACKAGE_VERSIONS = [1]`).

Also fixed here after it failed both tries on this branch: the TOB scrape raced the leader's bucket GC. The RPC's defaulting accessor returns an empty value for an entry destroyed between page assembly and read, which decoded as `failed to deserialize EpochCertsV1: unexpected end of input`. A destroyed bucket needs no seeding, so the scan now skips a vanished entry with a warning (bare and stamped buckets share one BCS layout, so present values decode unchanged).

## Tests

- `sui move test -e testnet`: 234 passed
- `cargo build` and `cargo clippy` over hashi, hashi-types and e2e-tests (lib + tests): clean
- Gate tests: all 8 of `move_upgrade_compat` + `version_policy_tests` pass in fresh-cycle mode (the run surfaced that `published_entries` rejected an entry-less `Published.toml`; it now treats that as the legal fresh-cycle state, serde-default plus doc)
- After the single-path and dormancy-kill rework: hashi lib 764/764, the gates, and the e2e set (withdrawal flow, drain mode, nonce window, exclusive upgrade, non-exclusive upgrade + disable, governance upgrade, backup/restore, rotation rejoin, resignation) all pass locally
- e2e on the squashed tree: `fresh_v1_upgrades_via_governance` (the reworked governance-upgrade lifecycle, fresh v1 to const-patched v2), `test_upgrade_via_proposal` (`keep_v1_enabled` boot, 110s) and `test_resigned_member_excluded_then_removed_permissionlessly` (fresh-default boot) all pass

## Follow-ups

IOP-660 (the record PR after the pre-mainnet publish) restores Published.toml, captures the fresh snapshot, bumps `PACKAGE_VERSION` to 2 and deletes the legacy v1 code paths.


0xsiddharthks (approved):

crates/hashi/src/withdrawals.rs
```diff
@@ -2784,10 +2782,10 @@ mod tests {
             safe_withdrawal_commit_max_inputs(
                 LEGACY_MAX_WITHDRAWAL_REQUESTS,
                 CoinSelectionParams::DEFAULT_MAX_INPUTS,
-                V1,
+                UNRESOLVED,
```
bmwill: Can we just kill off the old v1 handling instead of having it like this? Or do you want to do that in a follow up?

jessiemongeon1:

bmwill (approved):

PR #1077, opened 2026-09-01
watcher: track TOB entries

bmwill:


crates/hashi/src/onchain/mod.rs
```diff
@@ -577,7 +562,77 @@ impl OnchainState {
...
+        let state = self.state();
+        let Some(bucket) = state.hashi.tob.buckets.get(&key) else {
+            return Ok(None);
+        };
+        let certs: Vec<(Address, move_types::StampedDealerSubmissionV1)> = bucket
+            .certs_in_order()
+            .into_iter()
+            .map(|(dealer, submission)| (dealer, submission.clone()))
+            .collect();
+        if certs.len() != bucket.nodes.len() {
```
zhouwfang: Both operands come from bucket.nodes, so this catches "I hold nodes I can't reach" but not "I can't reach nodes I don't hold". A missing tail passes, and head with no nodes returns Ok(empty) for a bucket that has certs. The old walk compared against the LinkedTable's size, which is decoded at both write sites and just isn't stored here.

crates/hashi/src/onchain/mod.rs
```diff
@@ -577,7 +562,77 @@ impl OnchainState {
...
+        }
+        Ok(Some((bucket.layout, certs)))
+    }
+
+    /// True once the object mirror has applied every Hashi transaction
+    /// through a checkpoint whose timestamp is past `cutoff_ms`: any TOB
+    /// submission stamped at or before the cutoff is either already in
+    /// the mirror or will never land. `false` while the clock stream or
+    /// the mirror still trails the cutoff, and with no mirror running.
+    pub fn mirror_past_timestamp_ms(&self, cutoff_ms: u64) -> bool {
```
zhouwfang: This re-samples the latest checkpoint each call, so it asks whether the mirror has caught the tip now rather than whether it's past the cutoff. A mirror steadily one checkpoint behind never returns true. Latching the first (height, ts) past the cutoff and calling wait_until_checkpoint would match what the rest of the codebase does.

crates/hashi/src/communication/sui_tob.rs
```diff
@@ -397,23 +331,46 @@ impl OrderedBroadcastChannel<CertificateV1> for SuiTobSessionChannel {
...
-        let outcome = raced_outcome(&classify_published_cert(
-            &settled,
-            &ours.dealer_address,
-            ours.messages_hash,
-        ));
+        // Nothing inserted means a certificate for this dealer already
+        // holds the slot on-chain, but the mirror may not have applied
+        // the transaction that beat ours yet — poll briefly for it
+        // rather than misreading the lag as divergence.
+        let settled_deadline = tokio::time::Instant::now() + TX_CONFIRMATION_TIMEOUT;
```
zhouwfang: settled_deadline is 30s, the same as CALL_TIMEOUT, and it's set after the submit, so the break arm can't be reached before the outer timeout cancels the future. main had a static assert forbidding exactly this and it's deleted here. A stale dedup read then resubmits and retries up to ten times, each a paid transaction.

crates/hashi/src/onchain/mod.rs
```diff
@@ -1604,57 +1446,170 @@ async fn scrape_hashi(
...
-            // layout, so a present value decodes through the bare mirror.
+            // needs no seeding; skip it — the replay covers its deletion.
             let Some(value) = field.value_opt().filter(|bcs| !bcs.value().is_empty()) else {
                 tracing::warn!(
                     field_id = field.field_id(),
                     "tob entry vanished mid-scrape; skipping"
                 );
                 continue;
             };
+            let key: move_types::TobKey = field
```
zhouwfang, line 1504: These decode failures are bootstrap-fatal, so an unknown bucket type stops the node starting and stops the mirror re-bootstrapping. list_tob_keys skips and warns on the same condition at line 1042, for this same bag. The trigger is a package upgrade adding a cert layout, which is when you least want every node failing to boot.

crates/hashi/src/communication/sui_tob.rs
```diff
@@ -22,24 +22,28 @@ use crate::mpc::types::CertificateV1;
...
+/// stream into the mirror.
 const POLL_INTERVAL: Duration = Duration::from_millis(500);
 const TX_CONFIRMATION_TIMEOUT: Duration = Duration::from_secs(30);
-pub(crate) const FETCH_STALL_TIMEOUT: Duration = Duration::from_secs(60);
-const DEDUP_READ_TIMEOUT: Duration = Duration::from_secs(10);
+/// Bound for the post-race settled read: how long publish waits for the
+/// mirror to show the certificate that beat ours before classifying the
+/// race from whatever it holds.
+const SETTLED_READ_TIMEOUT: Duration = Duration::from_secs(10);
 const _: () = assert!(
```
zhouwfang, line 44: nit: The constant and the assert are right, but the message claims more than it proves. TX_CONFIRMATION_TIMEOUT is also 30s, so a slow submit can still eat the budget and skip the deadline arm

crates/hashi/src/onchain/types.rs
```diff
@@ -78,6 +78,108 @@ impl Hashi {
...
+    /// complete: the walk must cover exactly the `LinkedTable`'s
+    /// on-chain size. Comparing against held nodes instead would pass a
+    /// walk that reaches every node the mirror holds while the chain
+    /// holds more (a missing tail, or a head with no nodes applied
+    /// yet).
+    pub fn complete_certs_in_order(
+        &self,
+    ) -> Result<Vec<(Address, &move_types::StampedDealerSubmissionV1)>, IncompleteTobWalk> {
+        let certs = self.certs_in_order();
+        if certs.len() as u64 != self.size {
```
zhouwfang: This rejects walk > size, which the old walk deliberately accepted. It's reachable because the scrape reads size from the bag before listing nodes, so a submission landing in between gives size N with N+1 nodes.

crates/hashi/src/onchain/mod.rs
bmwill: yes thats expected, if we don't understand the layout the node needs to stop as its likely running older software. IIRC we already have a check elsewhere for this in the service code

zhouwfang: Fair, under a binary-first rollout that's the right call. One small thing: this fails before the version gate can flag BinaryUnsupported, so on a fresh boot the operator sees a BCS error instead of the halt signal they already watch. Checking the versions before decoding the bag would keep this as a pure backstop. Not blocking.

zhouwfang (approved):

PR #1080, opened 2026-09-01
Harden committee handoff reconfiguration

dlukegordon:
- F40: Nodes cached committee handoff signatures by epoch. After an aborted reconfiguration, a retry could reuse a signature for the discarded committee and stall. This PR removes the cache so nodes always sign the current transition.
- F37: Guardian accepted a valid handoff before the new committee was active. This could leave Guardian on the wrong committee and halt withdrawals. This PR requires a second certificate signed by the new committee after it becomes active.
- Multi-epoch catch-up still works: Guardian verifies the stored handoff chain and one activation certificate from the current committee.
- Guardian and the proxy do not read Sui. The Move intent registry reserves `CommitteeActivation` at `0x0006`; activation verification remains off-chain.
- Deployment: This requires a coordinated testnet rollout: upgrade the Hashi Move package, redeploy the Guardian and proxy, and release the matching Hashi build to all testnet validators.
- Linear: https://linear.app/mysten-labs/issue/SEC-531/hashi-guardiannode-committee-handoff-kill-chain-unactivated-handoff


crates/hashi/src/mpc/service.rs
zhouwfang, line 1973: Move aborts fail in the dry run, so they land in the generic retry arm. If the handoff cert is already on chain, every node retries this PTB forever. Anyone with the cert can put it there, so one compromised member can cause the abort on its own. On the already-submitted case I'd retry without the handoff call.

dlukegordon: Handled in the current version. The handoff and `end_reconfig` are now separate transactions: `submit_end_reconfig` first calls `submit_committee_handoff_if_needed`, which classifies the `set_pending_committee_handoff_cert` Move abort as `CommitteeHandoffAlreadySubmitted` and returns success. It then submits `execute_end_reconfig` without the handoff call. So a cert submitted by another node no longer leaves every node retrying the same combined PTB.

crates/hashi/src/leader/guardian.rs
```diff
@@ -302,20 +293,57 @@ impl LeaderService {
...
+                "stored guardian handoff did not advance: {from_epoch}->{to_epoch}"
+            );
+            transitions.push(signed);
             guardian_epoch = to_epoch;
         }
 
-        if transitions.is_empty() {
-            return Ok(());
-        }
+        anyhow::ensure!(
```
zhouwfang, line 311: If the guardian is ahead of this node's view, this now errors every checkpoint with a message about a chain that was never built. Main warned once and returned. That's the state the disclosure leaves behind, so the explicit branch is worth keeping.

crates/hashi/src/mpc/service.rs
zhouwfang: That arm only fires for an abort at execution time. The executor simulates first, so the Move abort comes back as a generic retry. Pre-existing and outside this PR now, so fine as a follow-up.

crates/hashi-guardian/src/withdraw_mode/committee_update.rs
```diff
@@ -3,300 +3,486 @@
...
+        .ok_or_else(|| InvalidInputs("committee transition chain is empty".to_string()))?;
+    if activation.message().new_committee != final_transition.message().new_committee {
+        return Err(InvalidInputs(
+            "activation payload does not match final transition payload".to_string(),
+        ));
     }
 
-    let new_committee: HashiCommittee = signed
+    let final_epoch = activation.message().new_committee.epoch;
+    let final_committee: HashiCommittee = activation
```
zhouwfang: This verifies the caller's activation committee before checking the installed epoch, and the proxy forwards this call without auth. Doing the installed checks and the chain first would keep an unauthenticated caller from making the enclave verify an arbitrary committee under the control lock.

packages/hashi/sources/core/intent.move
```diff
@@ -1,16 +1,16 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-/// Domain-separation intents for everything signed by Hashi member BLS keys.
+/// Domain-separation intents for Hashi member BLS signatures verified by Move.
```
zhouwfang: nit: 0x0006 is used by Rust now but not reserved here, so a future Move intent could land on the same domain.

crates/hashi/src/leader/guardian.rs
dlukegordon: Fixed. The explicit Guardian-ahead branch is restored: it logs the ahead state and returns success while the local on-chain view catches up. The exact post-chain epoch check remains for malformed or overshooting stored chains.

crates/hashi-guardian/src/withdraw_mode/committee_update.rs
dlukegordon: Fixed. Guardian now reads the installed committee before certificate verification. Exact retries must match the installed committee before activation is verified against that trusted committee. Advancing requests validate the full transition chain from the installed trust root before verifying activation against the chain-derived final committee. I also added a bad-chain + bad-activation precedence case proving the chain fails first without activation verification.

packages/hashi/sources/core/intent.move
dlukegordon: Fixed by restoring the synchronized Move registry entry: `COMMITTEE_ACTIVATION = 0x0006` plus the package accessor. Rust and Move now reserve the same domain, and the registry comments again state that they must stay in sync.

crates/hashi/src/mpc/service.rs
dlukegordon: Agreed. I created [SEC-575](https://linear.app/mysten-labs/issue/SEC-575/hashi-stop-retrying-an-already-submitted-committee-handoff-after-dry) to track the simulation-time already-submitted handoff retry path. It is assigned to me, in the Hashi project, and includes the mirrored pending-handoff state design and acceptance coverage. Leaving that fix out of this PR.

PR #1082, opened 2026-09-01
address I03

bmwill:


Bridgerz:
Move side looks good to me. One concern before we merge though: we measured this exact pattern on testnet a few days ago and same-epoch duplicate approvals are common. Over ~29h, 590 of 6,282 approval PTBs re-approved something an earlier PTB already approved, and 1,103 of the 1,136 re-approvals came from a different sender (median gap 8s). It's the new leader re-approving across a handoff because the approval path doesn't wait for the mirror to catch up, tracked in IOP-662.

With this change that duplicate aborts the whole approval PTB instead of being a harmless no-op, and the leader's retry loop resubmits the same batch. So I think we need the node-side handoff dedupe (or a wait_until_checkpoint after the approval PTB, like the deposit path does) to land first or together with this. I noticed IOP-662 got moved to Done but I don't see the fix on main, so maybe you have something in flight?

Alternative if we want this in sooner: make the same-epoch case a silent no-op on-chain and keep the strictly-newer-epoch replacement, then flip it to an abort once the leader side is fixed.

bmwill:
I'm going to just rework the withdrawal approval code to match what deposits do, that way we have a bit of uniformity.

Bridgerz (approved):
LGTM

PR #1083, opened 2026-09-01
Make `MpcManager`'s target-epoch identity optional

zhouwfang:
#95

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -177,15 +177,20 @@ pub struct NoncePartyOutcome {
     pub local_skips: u32,
 }
 
+struct TargetIdentity {
+    party_id: PartyId,
+    encryption_key: PrivateKey<EncryptionGroupElement>,
+    signing_key: Bls12381PrivateKey,
+}
+
 pub struct MpcManager {
     // Immutable during the epoch
-    pub party_id: PartyId,
+    /// Identity in the target epoch's committee, absent when this node is not one of its members.
+    identity: Option<TargetIdentity>,
     pub address: Address,
     pub mpc_config: MpcConfig,
     protocol_type: ProtocolType,
-    pub encryption_key: PrivateKey<EncryptionGroupElement>,
     pub previous_encryption_key: Option<PrivateKey<EncryptionGroupElement>>,
-    pub signing_key: Bls12381PrivateKey,
     pub committee: Committee,
     pub previous_committee: Option<Committee>,
     pub previous_nodes: Option<Nodes<EncryptionGroupElement>>,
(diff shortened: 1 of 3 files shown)


dlukegordon (commented):
Found three issues in the optional target-identity path.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -314,41 +340,51 @@ impl MpcManager {
             nonce_generation_protocol,
             committee.mpc_nonce_accumulation_window_ms(),
         );
-        let party_id = committee
-            .index_of(&address)
-            .expect("address not in committee") as u16;
+        let party_id_opt = committee.index_of(&address).map(|i| i as u16);
```
dlukegordon, line 343: A validator removed from the target committee still cannot reach this optional-identity path. `Hashi::create_mpc_manager` resolves both target encryption and signing keys first (`lib.rs:634-651`), and those lookups reject an address that is not in the committee. `handle_reconfig` then keeps retrying setup. Please make the target keys optional through the production factory and constructor, and test the real setup path with a removed validator instead of calling `MpcManager::new` directly.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -5198,9 +5237,9 @@ impl MpcManager {
                 share_index,
             )?);
             match process_avss_message(
-                &self.encryption_key,
+                self.encryption_key()?,
```
dlukegordon, line 5240: This still prevents a previous-only validator from dealing its old shares. `run_key_rotation` enters dealer mode for previous members, but this target-identity lookup returns `InvalidConfig` before their rotation certificate is published. The dealer path also unconditionally adds its own signature to the target-committee aggregator, which cannot accept a removed address. Please support rotation dealers without a target identity: skip local target-party processing/signing, send the batch to every target member, and aggregate only the signatures returned by target members.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -314,41 +340,51 @@ impl MpcManager {
...
             "MpcManager initialized"
         );
-        metrics.mpc_party_reduced_weight.set(
-            mpc_config
-                .nodes
-                .weight_of(party_id)
-                .expect("party_id was just derived from this committee") as i64,
-        );
-        if !keys_match {
+        if let Some(pid) = party_id_opt {
```
dlukegordon, line 372: When a current member becomes a nonmember, this leaves `mpc_party_reduced_weight` at the previous epoch's value because the gauge is only updated inside the `Some(pid)` branch. Please set it to zero when there is no target identity so the unlabelled current-committee gauge does not report stale membership.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Already addressed in the follow-up branch.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Already addressed in the follow-up branch.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Already addressed in the follow-up branch.

dlukegordon (approved):
All review threads are resolved.

PR #1085, opened 2026-09-02
Stop evicting peers for bad partial signatures

zhouwfang:
#179


dlukegordon (commented):
`crates/hashi/src/mpc/signing.rs:1235` — When a shared flag leaves this input with `t - 1` unflagged shares, `try_finalize_signature` can leave `required` at `t + 2`. The next honest response makes the unflagged prefix at lines 1045–1051 immediately usable, but this raw-count check still waits for `t + 2`. A slow remaining peer can therefore hold the result until its poll timeout or the signing deadline. Please include flagged-candidate readiness in the round-completion predicate for each pending input instead of checking only `partials.len()`.

dlukegordon (approved):

PR #1089, opened 2026-09-02
Harden committee handoff reconfiguration

dlukegordon:
- Remove stale handoff cache so reconfiguration retries sign current state
- Submit handoff atomically to prevent exposing unactivated committee certificates
- Handle already submitted committee handoffs

### DIFF
--- a/crates/hashi/src/mpc/service.rs
+++ b/crates/hashi/src/mpc/service.rs
@@ -14,6 +14,7 @@ use futures::future::join_all;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use sui_futures::service::Service;
+use sui_rpc::proto::sui::rpc::v2::ExecutionError;
 use sui_rpc::proto::sui::rpc::v2::execution_error::ExecutionErrorKind;
 use tokio::sync::watch;
 use tracing::debug;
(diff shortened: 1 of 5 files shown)


crates/hashi/src/mpc/service.rs
zhouwfang: Every node that loses the end_reconfig race now lands here, and a mirror more than 60s behind turns into a panic that takes the process down. On main it just kept retrying until the guard saw the change. I'd retry after the wait instead, since lag is most likely right at the epoch flip.

crates/hashi/src/withdrawals.rs
```diff
@@ -1096,9 +1092,7 @@ impl Hashi {
         // committee, so these are the only bytes worth signing.
         let transition = hashi_types::guardian::CommitteeTransitionRequest { new_committee };
 
-        let signature = self.sign_message_proto_at_epoch(&transition, from_epoch)?;
-        self.store_committee_handoff_signature(from_epoch, signature.clone());
-        Ok(signature)
+        self.sign_message_proto_at_epoch(&transition, from_epoch)
```
zhouwfang, line 1103: nit: This signs for any registered validator, including resigned ones. Requiring the caller to be in the committee would shrink the residual to actual members.

crates/hashi/src/mpc/service.rs
dlukegordon: Fixed. After the visibility wait, a still-pending mirror now stays in the inner submission loop and retries instead of returning an error to the outer panic arm.

crates/hashi/src/withdrawals.rs
dlukegordon: Fixed. The authenticated caller is now passed into transition-signature validation and must be a member of the outgoing committee before the node signs.

zhouwfang (approved):

PR #1094, opened 2026-09-02
Let a retiring committee member deal into the rotation that removes it

zhouwfang:
#95

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -1033,11 +1033,24 @@ impl MpcManager {
         // key shares and cannot generate valid rotation messages.
         let has_previous_shares = !previous.key_shares.shares.is_empty();
         if is_member_of_previous_committee && !has_previous_shares {
-            tracing::warn!(
-                "run_key_rotation: in the previous committee but holding no shares to reshare; \
-                 this node's previous-epoch weight will not reach the rotation"
-            );
-            metrics.mpc_rotation_previous_shares_missing_total.inc();
+            let owed_previous_shares = {
+                let mgr = mpc_manager.read().unwrap();
+                let address = mgr.address;
+                mgr.previous_share_ids_of(&address)
+                    .is_ok_and(|ids| !ids.is_empty())
+            };
+            if owed_previous_shares {
+                tracing::warn!(
+                    "run_key_rotation: in the previous committee but holding no shares to \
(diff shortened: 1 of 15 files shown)


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -978,52 +1027,77 @@ impl MpcManager {
...
+                ),
+            }
+        }
+        if role == RotationRole::DealerOnly {
+            let outcome = if dealt {
+                ReconfigOutcome::Dealt
+            } else if !has_previous_shares {
+                ReconfigOutcome::NoShares
+            } else {
+                ReconfigOutcome::NotNeeded
```
dlukegordon, line 1114: A transient reconstruction failure becomes terminal here. `prepare_previous_output` catches message-retrieval and complaint RPC failures and can return the public-only fallback with empty shares. `DealerOnly` maps that to `NoShares`, and `handle_reconfig` then waits for pending to clear without retrying. If this member’s previous share indices are needed for `previous.threshold`, the target parties can never finish, so pending never clears. Please distinguish genuine previous-key loss from a retryable reconstruction fallback and return an error for the latter.

crates/hashi/src/mpc/service.rs
```diff
@@ -1521,7 +1569,35 @@ impl MpcService {
...
+                        .mpc_reconfig_no_output_total
+                        .with_label_values(&[protocol_label, reason])
+                        .inc();
+                    _reconfig_timer.stop_and_discard();
+                    while self.get_pending_epoch_change() == Some(target_epoch) {
+                        metrics.task_heartbeat("mpc_service");
+                        self.sleep_if_still_pending(target_epoch).await;
+                    }
+                    if !matches!(outcome, ReconfigOutcome::NoRole) {
+                        self.post_reconfig_housekeeping(target_epoch).await;
```
dlukegordon: This can re-register a validator that deliberately resigned. A dealer-only node waits for pending to clear, then calls housekeeping and `prepare_and_register_keys`. If `remove_inactive_member` deletes the resigned `MemberInfo` before this process wakes, and `resigned.marker` has not yet been written, the latch sees neither signal and registers the active Sui validator again. Please capture the resignation before waiting, or skip only the key-registration part of housekeeping for dealer-only outcomes.

dlukegordon (commented):
Two correctness issues found in the dealer-only reconfiguration path.

dlukegordon (approved):
Both review findings have been resolved.

PR #1095, opened 2026-09-02
Decode the certificate's signer bitmap in nonce recovery

zhouwfang:
#193

### DIFF
--- a/crates/hashi/src/mpc/mpc_except_signing.rs
+++ b/crates/hashi/src/mpc/mpc_except_signing.rs
@@ -123,6 +123,7 @@ const HEDGED_RETRIEVE_INITIAL_ROUND_SIZE: usize = 2;
 const HEDGED_RETRIEVE_ROUND_GROWTH_FACTOR: usize = 2;
 const HEDGED_RETRIEVE_ROUND_TIMEOUT: Duration = Duration::from_secs(1);
 const PREVIOUS_MESSAGE_REPAIR_ATTEMPT_TIMEOUT: Duration = Duration::from_secs(20);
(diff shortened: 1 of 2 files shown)


crates/hashi/src/mpc/service.rs
```diff
@@ -1319,14 +1320,23 @@ impl MpcService {
         let (outputs, expected_size) = match protocol {
             NonceGenerationProtocol::Vanilla => {
                 let expected_size = expected_from(certified_nonce_weight(mpc_manager, &certs))?;
-                let outputs = MpcManager::reconstruct_presignatures_with_complaint_recovery(
-                    mpc_manager,
-                    epoch,
-                    batch_index,
-                    &certs,
-                    &p2p_channel,
+                let outputs = tokio::time::timeout(
```
dlukegordon: The 600s timeout covers all dealers, but complaint recovery is sequential and each dealer can use most of the existing 10 × 30s RPC retry budget. If a later dealer hits this timeout, earlier recovered outputs exist only in `dealer_nonce_outputs`; the next `sync_if_stale` attempt rebuilds `MpcManager` and clears that map. With stable slow responses and `num_consumed_presigs > 0`, recovery can repeat forever without retaining progress. Please preserve the manager/recovered outputs across retries, persist them, or scope the timeout per dealer.

zhouwfang: Yeah, switched to the third option. If an attempt fails, the complaint-recovered outputs will get redone on the next attempt.

dlukegordon (approved):
The reported recovery-timeout issue has been addressed. The updated signer-bitmap decoding and nonce-recovery changes look correct.

PR #1096, opened 2026-09-02
cli: route `hashi register` at the highest enabled package version

Bridgerz:
## Summary

`hashi register` built every `validator::*` entry call against the configured package id, which is the original publish. Every one of those entries gates on the called package's `assert_version_enabled`, so once that version is retired the whole transaction aborts with `EVersionDisabled`. On the live testnet v1 has been disabled since the 2026-09-01 upgrade, which means first registration (`hashi register --print-only`) and operator-key rotation, the two flows the runbook routes through the CLI, do not work today. The node's own startup registration was unaffected because `execute_register_or_update_validator` already routes through `active_call_package_id()`.

Closes IOP-558.

## Change

- `cli/commands/mod.rs`: `resolve_latest_enabled_package`, a one-shot governance read that returns the package of the highest version that is both governance-enabled and published. It deliberately does not intersect with `SUPPORTED_PACKAGE_VERSIONS`: that list gates what a node may decode and mutate autonomously, while the registration transaction is a fixed set of `validator::*` calls whose only version dependency is the called package's enable check, so the live package is the right target even for a CLI built before the chain's latest upgrade. No enabled published version is a hard error, never a fallback to the original id. This is the same rule `latest_package_id` already applies to the other `validator::*` lifecycle entries in the CLI client.
- `cli/mod.rs` `run_register`: resolve once, before the serialize/dry-run and execute paths split, and pass the result as the builder's `call_package`.
- `node-operator-runbook.mdx`: one sentence noting that `hashi register` resolves the highest enabled version on-chain and that a build against a retired package fails in simulation instead of being emitted.

`build_register_or_update_validator_tx` and the node path are untouched. Type tags still use the original package id, which is correct.

## Test

`test_cli_register_targets_the_enabled_package` (e2e): boots with both versions enabled, retires v1 through governance executed via the upgraded package, then runs the runbook's operator-address rotation. Building against the configured id fails in simulation with a Move abort in `versioning` (the live bug). Resolving through the new resolver yields the upgraded package, the same transaction lands, and a second build against the same config has nothing left to send.

The test drives the disable itself rather than using `with_upgraded_chain()`: on the fresh cycle the binary supports v1 only, so that boot's own assertion that active version 2 resolves cannot hold on main. Nothing on main currently uses that boot; noting it here rather than changing the harness in this PR.

## Backport

The live break is on the `testnet` maintenance branch, where operators build from. A cherry-pick of the CLI change follows as a separate PR against `testnet`; the e2e test stays on main.

### DIFF
--- a/crates/e2e-tests/src/e2e_flow.rs
+++ b/crates/e2e-tests/src/e2e_flow.rs
@@ -3698,4 +3698,165 @@ mod tests {
         info!("=== Drain Mode Max Batch Test Passed ===");
         Ok(())
     }
+
+    /// `hashi register` (the CLI path through
+    /// `build_register_or_update_validator_tx`) must target a live package:
+    /// every `validator::*` entry gates on the called package's
+    /// `assert_version_enabled`, so the original publish id aborts once its
+    /// version is retired. The node's startup registration already routes
+    /// past it; this pins the CLI resolver to the highest enabled published
+    /// version on a chain where v1 is disabled (IOP-558). The disable is
+    /// driven here rather than by the harness boot because the fresh-cycle
+    /// binary supports v1 only, and the resolver must not depend on that.
+    #[tokio::test]
+    async fn test_cli_register_targets_the_enabled_package() -> Result<()> {
+        use sui_crypto::SuiSigner as _;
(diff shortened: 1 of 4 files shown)


bmwill (approved):

jessiemongeon1:

PR #1097, opened 2026-09-02
cli: route `hashi register` at the highest enabled package version (testnet backport)

Bridgerz:
## Summary

Backport of #1096 to the `testnet` maintenance branch, where operators build the CLI from.

`hashi register` builds every `validator::*` entry call against the configured package id, which is the original v1 publish. Those entries gate on the called package's `assert_version_enabled`, and v1 has been disabled on testnet since the 2026-09-01 upgrade, so first registration (`hashi register --print-only`) and operator-key rotation through the CLI abort with `EVersionDisabled` today. Node startup registration is unaffected because it already routes through the active package.

## Change

Same as #1096: resolve the highest governance-enabled published version from a one-shot governance read and build against it on both the serialize/dry-run and execute paths, with no fallback to the original id. On this branch the resolver yields the v2 package `0x8f7efd743897fde48cc35b6203cd72c7ad4248f0eb02a9ad378e4a2d39cc2c7e`.

The e2e test from #1096 is not carried over: this branch's harness boots from the deployed snapshot rather than the fresh-cycle upgrade path the test drives.

## Operator note

Any unsigned registration transaction serialized with an earlier CLI build encodes the v1 package and will abort when broadcast. Re-generate with a build that includes this change.

Refs IOP-558.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6294,8 +6294,7 @@ checksum = "734676eb262c623cec13c3155096e08d1f8f29adce39ba17948b18dad1e54142"
 [[package]]
 name = "sui-crypto"
 version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b7b8f54c4405002c9ece67cb848479f80837185a55a59d8c0a105240f737d0da"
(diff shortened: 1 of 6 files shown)


bmwill (approved):

jessiemongeon1:

PR #1102, opened 2026-09-02
config: split the governed config into instant and epoch stores, mirror it in the node and CLI

Bridgerz:
## Summary

Split the governed config into two stores and mirror the split in the node. `Hashi.config` keeps the instant keys (Bitcoin parameters, guardian, pause, thresholds) and stays the target of `UpdateConfig`. A new `Hashi.epoch_config` holds the MPC parameters and is copied verbatim onto each new committee by `start_reconfig`, replacing the four-key `mpc_config::pin` snapshot; nothing is seeded or repaired in transit, and the one cross-key MPC rule (the weight-reduction delta stays below max-faulty) is enforced when a proposal writes the store (`mpc_config::is_consistent`, `EInconsistentMpcConfig`). Two proposal types join it: `UpdateEpochConfig` (existing epoch keys, MPC range validation, lands in the next committee) and `AddConfig { epoch, entries }` (insert-only into either store), which is how a node-side setting becomes governable without a package upgrade. Design is Brandon's.

Layout change to `Hashi`, so pre-mainnet only. Rebased from the pre-cut branch onto current main; the retired `mpc_threshold_in_basis_points` key stays rejected on add and update because the node treats its presence in a pinned config as the legacy threshold formula and the live testnet global config still carries it.

This was opened as a two-PR stack (Move here, Rust in #1106). The e2e job builds the tree's node against the tree's Move package, and `epoch_config` sits between `config` and `versioning`, so the Move half alone fails every test that decodes the `Hashi` object ("expected option type"). The two halves are one change. #1106 shows as merged only because its base branch was fast-forwarded to its head; nothing from it reached main, and its commit is the third one here. The commits still read in order: Move split, pinned keys, Rust mirror.

## Commit 4: review follow-up

Per review: `start_reconfig` no longer runs `mpc_config::pin`; the store is copied as is and the committee's field, the formation parameters and the local are renamed `epoch_config`. The seeding half of `pin` was dead on the fresh cycle (genesis seeds every key, nothing deletes one) and the per-key ranges were already refused at write, so only the cross-key rule moved: both epoch-store proposals assert `mpc_config::is_consistent` on the store they leave behind, after all entries, so a joint update of both coupled keys is judged on its result and not on entry order. Move 270 passed.

## Commit 1: Move split

`hashi.move`, `config.move`, `start_reconfig`, the `add_config` and `update_epoch_config` proposal modules, and the test utilities. `mpc_config::pin` is gone; the committee reads its parameters from the pinned copy.

## Commit 2: pinned keys on the new paths

#1062 made `UpdateConfig` refuse the guardian BTC public key and the Bitcoin chain id. The two new write paths apply the same rule, checked before the target store is consulted, so a same-named entry cannot be introduced into the epoch store either. Nothing reads a pinned key from the epoch store today; this keeps one rule in one place.

## Commit 3: Rust mirror, CLI, e2e

The on-chain mirror carries `Hashi.epoch_config`, the node reads epoch keys from the committee's pinned copy and instant keys from the root config, the CLI gains `create update-epoch-config` and `create add-config` (with `--epoch`), the mirror and GC know the two new proposal types, and the e2e config-override driver routes each key by which store holds it.

Three things the fresh cycle changed under the pre-cut branch, none flagged by the merge: the `UpgradeV2` proposal arm no longer exists, the two new `MoveType` impls carried `PACKAGE_VERSION = 2` overrides from the v2 era and now use the trait default of 1 like every other type on the squashed chain, and the proposal-type parsing test expected version 2. All three are back to the v1 baseline. The runbook's proposal table keeps main's single `Upgrade` row with the `--exclusive` semantics and adds the three config rows.

## Tests

Move: `sui move test` under the CI-pinned toolchain, 270 passed. New `add_config_tests` and `update_epoch_config_tests`, pinned-key refusals on both new paths, and the pre-genesis proposal test from #1099 adapted to the eight-argument `create_for_testing`.

Rust: clippy clean across `hashi`, `hashi-types` and `e2e-tests`; the onchain and CLI unit tests; the three config-store e2e tests (`test_create_update_config_proposal`, `test_mpc_config_defaults_match_rust`, `test_varying_t_and_allowed_delta_across_epochs`) against a fresh v1 boot.

Closes IOP-455.

### DIFF
--- a/packages/hashi/sources/core/mpc_config.move
+++ b/packages/hashi/sources/core/mpc_config.move
@@ -86,51 +86,6 @@ public(package) fun nonce_accumulation_window_ms(config: &Config): u64 {
         .destroy_or!(DEFAULT_NONCE_ACCUMULATION_WINDOW_MS)
 }
 
-public(package) fun seed_absent_defaults(config: &mut Config) {
-    seed_if_absent(
-        config,
-        KEY_WEIGHT_REDUCTION_ALLOWED_DELTA,
-        DEFAULT_WEIGHT_REDUCTION_ALLOWED_DELTA,
-    );
-    seed_if_absent(config, KEY_MAX_FAULTY_IN_BASIS_POINTS, DEFAULT_MAX_FAULTY_IN_BASIS_POINTS);
-    seed_if_absent(config, KEY_NONCE_GENERATION_PROTOCOL, VANILLA_NONCE_GENERATION_PROTOCOL);
-    seed_if_absent(
-        config,
-        KEY_NONCE_ACCUMULATION_WINDOW_MS,
-        DEFAULT_NONCE_ACCUMULATION_WINDOW_MS,
-    );
-}
-
-fun reset_if_out_of_range(config: &mut Config, key: vector<u8>, lo: u64, hi: u64, default: u64) {
-    config.try_get(key).map!(|v| v.as_u64()).do!(|value| if (value < lo || value > hi) {
-        config.upsert(key, config_value::new_u64(default));
-    });
-}
-
-fun seed_if_absent(config: &mut Config, key: vector<u8>, default: u64) {
-    if (!config.contains(key)) {
-        config.upsert(key, config_value::new_u64(default));
-    };
-}
-
-fun repair_out_of_range(config: &mut Config) {
-    reset_if_out_of_range(
-        config,
-        KEY_MAX_FAULTY_IN_BASIS_POINTS,
-        1,
-        MAX_FAULTY_BPS,
-        DEFAULT_MAX_FAULTY_IN_BASIS_POINTS,
-    );
-    let max_delta = max_faulty_in_basis_points(config) - 1;
-    clamp_at_most(config, KEY_WEIGHT_REDUCTION_ALLOWED_DELTA, max_delta);
-}
-
-fun clamp_at_most(config: &mut Config, key: vector<u8>, hi: u64) {
-    config.try_get(key).map!(|v| v.as_u64()).do!(|value| if (value > hi) {
-        config.upsert(key, config_value::new_u64(hi));
-    });
-}
-
 public(package) fun init_defaults(config: &mut Config) {
     config.upsert(
         KEY_WEIGHT_REDUCTION_ALLOWED_DELTA,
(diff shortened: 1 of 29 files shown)


jessiemongeon1:

packages/hashi/sources/core/reconfig.move
```diff
@@ -65,10 +65,9 @@ entry fun start_reconfig(
     // Assert that we are not already reconfiguring
     assert!(!self.committee_set().is_reconfiguring());
     assert_genesis_launch_authorized(self);
-    hashi::mpc_config::seed_absent_defaults(self.config_mut());
-    // Pin the current MPC parameters so they stay fixed for the new epoch even
-    // if governance changes them mid-epoch.
-    let config = hashi::mpc_config::pin(self.config_mut());
+    // Pin the epoch config wholesale so it stays fixed for the new epoch even
+    // if governance changes it mid-epoch.
+    let config = hashi::mpc_config::pin(self.epoch_config_mut());
```
bmwill: We should remove the mpc_config::pin call and pass through the epoch_config exactly as-is.

Lets also rename this to `epoch_config` here as well as in the committee itself (as well as the params in start_reconfig)

packages/hashi/sources/core/mpc_config.move
```diff
@@ -140,26 +150,15 @@ public(package) fun init_defaults(config: &mut Config) {
...
-    mpc
+/// Normalize the MPC parameters in the epoch config (seed absent keys, repair
+/// out-of-range values in place) and return the whole store for pinning onto
+/// the committee being formed. Every epoch-scoped key, MPC or
+/// governance-added, is copied wholesale so nodes read one immutable snapshot
+/// per epoch.
+public(package) fun pin(epoch_config: &mut Config): Config {
+    seed_absent_defaults(epoch_config);
+    repair_out_of_range(epoch_config);
+    *epoch_config
```
bmwill: I think we should avoid this and instead ensure those values have to be set properly period. Its going to be confusing to have a config be "copied" in but have it modified going in.

packages/hashi/sources/core/reconfig.move
Bridgerz: Done in 6ea81caa4. start_reconfig copies the store with `*self.epoch_config()` and passes it straight through; the local, the four formation parameters in committee_set and the Committee field are all `epoch_config` now. The Rust mirror keeps its field name (BCS is positional) and only its doc comments changed.

packages/hashi/sources/core/mpc_config.move
Bridgerz: Done in 6ea81caa4. pin and the seed/repair helpers are gone; nothing modifies the store on the way onto the committee. The per-key ranges were already refused at write by is_valid_value, and seeding was dead on the fresh cycle (genesis seeds every key, nothing deletes one). The one rule the per-entry checks could not see, delta below max-faulty, is now mpc_config::is_consistent, asserted by update_epoch_config and add_config on the store they leave behind (after all entries, so a joint update of both keys is judged on its result, not entry order). A proposal that would break it aborts with EInconsistentMpcConfig. Move 270 passed; the three config-store e2e tests pass.

bmwill (approved):

PR #1103, opened 2026-09-02
Reject a cached AVID nonce output whose common does not match cert

zhouwfang:
#531


dlukegordon (approved):
Reviewed the cached AVID output validation flow and regression coverage. No actionable findings.

PR #1106, opened 2026-09-02
rust: mirror the epoch config store and the AddConfig / UpdateEpochConfig proposals

Bridgerz:
## Summary

Rust half of the IOP-455 stack, on top of the Move split. The on-chain mirror carries `Hashi.epoch_config`, the node reads epoch keys from the committee's pinned copy and instant keys from the root config, the CLI gains `create update-epoch-config` and `create add-config` (with `--epoch`), the mirror and GC know the two new proposal types, and the e2e config-override driver routes each key by which store holds it.

Rebased from the pre-cut branch onto current main. Three things the fresh cycle changed under it, none flagged by the merge: the `UpgradeV2` proposal arm no longer exists (main renamed the module), the two new `MoveType` impls carried `PACKAGE_VERSION = 2` overrides from the v2 era and now use the trait default of 1 like every other type on the squashed chain, and the proposal-type parsing test expected version 2. All three are back to the v1 baseline. The runbook's proposal table keeps main's single `Upgrade` row with the `--exclusive` semantics and adds the three config rows.

## Tests

Clippy clean across `hashi`, `hashi-types` and `e2e-tests`; the onchain and CLI unit tests; and the three config-store e2e tests (`test_create_update_config_proposal`, `test_mpc_config_defaults_match_rust`, `test_varying_t_and_allowed_delta_across_epochs`) against a fresh v1 boot.

Closes IOP-455.

### DIFF
--- a/crates/e2e-tests/src/e2e_flow.rs
+++ b/crates/e2e-tests/src/e2e_flow.rs
@@ -2650,6 +2650,12 @@ mod tests {
         use hashi::onchain::types::DEFAULT_MPC_MAX_FAULTY_IN_BASIS_POINTS;
         use hashi::onchain::types::DEFAULT_MPC_WEIGHT_REDUCTION_ALLOWED_DELTA;
 
+        // Governance-added keys the Move package knows nothing about.
+        const EPOCH_KNOB: &str = "e2e_epoch_knob";
+        const EPOCH_KNOB_VALUE: u64 = 7;
+        const INSTANT_KNOB: &str = "e2e_instant_knob";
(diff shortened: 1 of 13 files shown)


Bridgerz:
Folded into #1102. The Move layout change cannot pass the e2e job without its Rust mirror (the job builds the tree's node against the tree's Move package, and epoch_config sits between config and versioning), so the two halves are one PR. GitHub shows this PR as merged because its base branch was fast-forwarded to its head; nothing from it has reached main. Review and merge happen on #1102, where this commit is the third one.

PR #1109, opened 2026-09-02
Bound per-peer work when merging partial signatures

zhouwfang:
#179

### DIFF
--- a/crates/hashi/src/mpc/rpc/p2p_channel.rs
+++ b/crates/hashi/src/mpc/rpc/p2p_channel.rs
@@ -1,6 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use super::proto_conversions::partial_sigs_response_limit;
 use crate::communication::ChannelError;
 use crate::communication::ChannelResult;
 use crate::communication::P2PChannel;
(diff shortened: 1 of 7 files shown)


crates/hashi/src/mpc/rpc/p2p_channel.rs
```diff
@@ -141,7 +148,10 @@ impl P2PChannel for RpcP2PChannel {
         party: &Address,
         request: &GetPartialSignaturesRequest,
     ) -> ChannelResult<GetPartialSignaturesResponse> {
-        let client = self.get_client(party)?;
+        let mut client = self.get_client(party)?;
+        if let Some(limit) = self.signing_response_limit {
+            client = client.tighten_max_decoding_message_size(limit);
```
dlukegordon: This limit is based on the full withdrawal, but each peer poll normally requests only the current 64-input chunk and can request fewer IDs on later rounds. A peer can therefore send unrelated entries up to the much larger transaction-wide limit; the conversion code BCS-deserializes those entries before the collector filters unknown IDs. For example, with 400 transaction inputs and 60 owned shares, a 64-ID poll accepts 1,086,336 bytes instead of 228,864. Can we store `max_owned_count` on the channel and calculate `partial_sigs_response_limit(request.signing_ids.len(), max_owned)` here?

dlukegordon (approved):

PR #1111, opened 2026-09-03
launch, node: refuse a Sui chain / Bitcoin chain pairing the protocol never deploys

Bridgerz:
## Summary

Audit finding SEC-510 (D-13, launch blocker). The four Bitcoin values in the deploy workflow travel as a matched set, and every launch-time and startup validation was an internal consistency check: `publish.rs` only asserted that the Bitcoin chain id was one of the four known networks, and Move `finish_publish` stores whatever it is handed. Pointing the deploy at Sui mainnet with the Bitcoin values still on signet would mint a real `hashi::btc` coin against free signet deposits.

Move cannot enforce the pairing (the Sui framework exposes no chain identifier to Move), so this enforces it in every Rust path that produces or consumes the pair.

## Rule

`constants::check_sui_bitcoin_chain_pairing`: Bitcoin mainnet if and only if Sui mainnet. Sui testnet, devnet and localnet must pair with a non-mainnet Bitcoin chain. This is stricter than the ticket's wording ("refuse mainnet BTC on a non-production Sui chain"), which maps onto `is_production_sui_chain` and would have allowed Sui testnet plus Bitcoin mainnet; no deployment or runbook pairs testnet with real BTC.

The Bitcoin side is decided by the network the hash resolves to (`network_from_chain_id`), as every consumer of the id does, not by string equality, so letter case cannot dodge the rule. An empty Sui chain id fails closed, since the RPC accessor yields `""` when a fullnode omits the field, and so does an unrecognised Bitcoin id. Each direction has its own message naming the Bitcoin network and, for the common Sui testnet mistake, the fact that `bitcoin_chain_id` defaults to mainnet when unset.

## Where it runs

- `build_finish_publish_tx` takes the Sui chain id and checks before its only RPC, so a refusal costs no network round trip and the ticket's test is hermetic.
- `hashi launch` fetches the id from the fullnode (`sui_rpc_client::fetch_sui_chain_id`) after the `--status` early return, so the `LAUNCH_STATUS` contract is untouched, prints it beside the RPC line, checks before resolving the sender or the UpgradeCap, and passes it to both builder call sites (dry-run, serialize-unsigned, execute). The RPC URL defaults to the mainnet fullnode, which is the foot-gun the audit describes; the printed id makes it visible.
- `publish::finish_publish` fetches the id itself, so the e2e harness is unchanged.
- `hashi register` checks the config it loads, and confirms the config's Sui chain id against the RPC it registers through, as the node does at startup. The client it builds for that is reused for the transaction.
- Node startup: `Hashi::verify_chain_pairing` right after `verify_sui_chain_id` has confirmed the local config against the RPC. It is a separate check from `verify_bitcoin_chain_id`, whose pre-launch skip would otherwise silence it in the key-registration window. This node-side check is the load-bearing control: Move never reads the pinned Bitcoin chain id, so which chain deposits are attested against is decided by each node's local config.

## Tests

`crates/hashi/tests/chain_pairing_tests.rs`: the predicate table including an uppercase mainnet hash and an unrecognised id; the launch builder refusing signet on Sui mainnet and Bitcoin mainnet on Sui testnet against a lazily connected client to a closed port; a permitted pair getting past the check to the network; the node refusing or accepting each pair from its config, including a Sui testnet config that omits `bitcoin_chain_id` and so defaults to Bitcoin mainnet (both config keys default to mainnet, so that config now fails at boot instead of later against bitcoind).

The e2e harness publishes with the localnet digest and regtest, which the rule permits, so every existing e2e boot and launch exercises the pass path.

## Residual, permanent

`finish_publish` is a Move entry that accepts any chain id, so a raw PTB or an older `hashi` binary holding the UpgradeCap bypasses the launch check, and node binaries older than this change still run a mismatched pair. This closes the operational foot-gun in the shipped tooling; it is not a protocol guarantee. The runbook's Networks section now says the pairing is enforced.

Closes IOP-663. Closes SEC-510.

### DIFF
--- a/crates/hashi/src/constants.rs
+++ b/crates/hashi/src/constants.rs
@@ -49,6 +49,47 @@ pub fn is_production_sui_chain(chain_id: &str) -> bool {
     chain_id == SUI_MAINNET_CHAIN_ID || chain_id == SUI_TESTNET_CHAIN_ID
 }
 
+/// Refuse a Sui chain / Bitcoin chain pairing the protocol never deploys:
+/// Bitcoin mainnet if and only if Sui mainnet. On Sui mainnet anything but
+/// Bitcoin mainnet would mint a real `hashi::btc` against free signet or
+/// regtest deposits; on any other Sui network Bitcoin mainnet would lock real
+/// Bitcoin behind a non-mainnet package (SEC-510, audit D-13).
+///
+/// Move cannot enforce this (`finish_publish` stores whichever id it is
+/// handed and the Sui framework exposes no chain identifier to Move), so it
+/// runs in every Rust path that produces or consumes the pair. The Bitcoin
+/// side is decided by the network the hash resolves to, as every consumer of
+/// the id does, not by string equality, so letter case cannot dodge it. An
+/// empty or unrecognised id fails closed.
+pub fn check_sui_bitcoin_chain_pairing(
+    sui_chain_id: &str,
+    bitcoin_chain_id: &str,
+) -> anyhow::Result<()> {
+    use crate::btc_monitor::config::Network;
+    use crate::btc_monitor::config::network_from_chain_id;
+
+    anyhow::ensure!(
+        !sui_chain_id.is_empty(),
+        "Sui chain ID is empty; refusing to pair it with Bitcoin chain {bitcoin_chain_id}"
+    );
+    let network = network_from_chain_id(bitcoin_chain_id)
+        .ok_or_else(|| anyhow::anyhow!("unrecognized bitcoin chain id: {bitcoin_chain_id}"))?;
+    let sui_is_mainnet = sui_chain_id == SUI_MAINNET_CHAIN_ID;
+    let bitcoin_is_mainnet = network == Network::Bitcoin;
+    anyhow::ensure!(
+        !sui_is_mainnet || bitcoin_is_mainnet,
+        "refusing Bitcoin {network:?} ({bitcoin_chain_id}) on Sui mainnet: \
+         Sui mainnet requires Bitcoin mainnet"
+    );
+    anyhow::ensure!(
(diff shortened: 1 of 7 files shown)


jessiemongeon1:

crates/hashi/src/constants.rs
```diff
@@ -49,6 +49,47 @@ pub fn is_production_sui_chain(chain_id: &str) -> bool {
...
+        "refusing Bitcoin {network:?} ({bitcoin_chain_id}) on Sui mainnet: \
+         Sui mainnet requires Bitcoin mainnet"
+    );
+    anyhow::ensure!(
+        !bitcoin_is_mainnet || sui_is_mainnet,
+        "refusing Bitcoin mainnet ({bitcoin_chain_id}) on Sui chain {sui_chain_id}: \
+         Bitcoin mainnet is allowed on Sui mainnet only; every other Sui network \
+         needs a non-mainnet bitcoin_chain_id, which defaults to mainnet when unset"
+    );
+    Ok(())
```
bmwill: This logic is sort of difficult for me to parse. I think this may be more readable via a fully enumerated match or something like:

```
match (sui_is_mainnet, bitcoin_is_mainnet) {
(true, true) | (false, false) => {}
(true, false) | (false, true) => // error here
}
```

or as an if statement:

```
if (sui_is_mainnet && bitcoin_is_mainnet) || (!sui_is_mainnet && !bitcoin_is_mainnet) {
 return Ok(())
} else {
return Err(..)
}
```

I always have such a hard time actually reading the ensure macro

Bridgerz: Done in 4d7be2575: rewritten as an exhaustive match on (sui_is_mainnet, bitcoin_is_mainnet) with one arm per mismatch direction. Same rule and messages, tests unchanged.

bmwill (approved):

PR #1113, opened 2026-09-03
[guardian-proxy] Relay KP ceremony confirmations

0xsiddharthks:
## Summary

Since #1073 a ceremony completes only once every KP has confirmed its share to the live ceremony guardian, but the proxy answered `ConfirmCeremony` with `PERMISSION_DENIED`, so a KP could only confirm over the operator's SSM tunnel. Forward it behind the gate the relay already applies to share submissions: the KP signature must verify, and the signer must be in the ceremony's committed roster read from the S3 share log. The enclave repeats both checks and binds each confirmation to its session and ceremony digest, so the proxy adds no authority; the roster check keeps a flood of self-signed requests off the enclave's serialized control lock. Exercised live on the `testing` stack: two KPs confirmed through the public endpoint and the ceremony completed.

## Changes

- `forward.rs`: forward `ConfirmCeremony` after the signature and roster checks. `OperatorInit`, `SetupNewKey`, `ProvisionerInit`, `OperatorActivate` and `RotateKpSet` stay denied.
- `RosterCache::authorize` is the one roster gate: the relay's share submissions and both forwarded KP-signed RPCs (`ProvisionerRotateCert` included, closing its TODO) go through it.
- A miss re-reads the share log, at most once per 5 s. `SetupNewKey` and `RotateKpSet` commit a new roster over the operator's tunnel, so nothing invalidates the cache for them; without the re-read a KP the new roster adds would be denied until the 60 s TTL, and `key-provisioner ceremony` confirms once without retrying.
- Tests: a rostered confirmation reaches the backend, an unrostered or malformed one is rejected before it, a tampered one fails `UNAUTHENTICATED`; in `roster.rs`, a KP added since the last read is admitted on its first call, the first ceremony is admitted from a cached miss, the re-read is bounded, and the gate's failure modes.

### DIFF
--- a/crates/hashi-guardian-proxy/src/forward.rs
+++ b/crates/hashi-guardian-proxy/src/forward.rs
@@ -377,7 +377,7 @@ mod tests {
         let (cert_armored, secret_armored) = mock_pgp_keypair();
         let cert = PgpPublicCert::new(cert_armored).unwrap();
         let store = StubStore::default();
-        seed_roster(&store, &[&cert.fingerprint().to_hex()]);
+        seed_roster(&store, 0, &[&cert.fingerprint().to_hex()]);
         let (stub, proxy) = spawn_stub_proxy(store).await;
 
         let status = proxy
@@ -404,7 +404,7 @@ mod tests {
         let (cert_armored, secret_armored) = mock_pgp_keypair();
         let cert = PgpPublicCert::new(cert_armored).unwrap();
         let store = StubStore::default();
-        seed_roster(&store, &["AAAABBBBCCCCDDDDEEEE11112222333344445555"]);
+        seed_roster(&store, 0, &["AAAABBBBCCCCDDDDEEEE11112222333344445555"]);
         let (stub, proxy) = spawn_stub_proxy(store).await;
 
         let err = proxy
--- a/crates/hashi-guardian-proxy/src/forward.rs
+++ b/crates/hashi-guardian-proxy/src/forward.rs
@@ -5,8 +5,7 @@
 //! and rejects the operator surface with `PERMISSION_DENIED`: the proxy is
 //! internet-facing and `OperatorInit` is one-shot and unauthenticated, so
 //! exposing it would let anyone wedge the guardian. KP-signed RPCs are
-//! forwarded after a signature check, so a KP on its own machine reaches the
-//! guardian through the public endpoint. Wrapped by
+//! forwarded after a signature check. Wrapped by
 //! [`crate::cache::CachingGuardianGrpc`] to cache `StandardWithdrawal`.
 
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


crates/hashi-guardian-proxy/src/forward.rs
```diff
@@ -107,7 +113,17 @@ impl<L: LogStore> GuardianService for Forwarding<L> {
...
     }
 
-    // --- Rejected: operator/ceremony surface ---
+    // Safe to expose: the enclave binds each confirmation to its session, the
+    // ceremony digest and the dealt roster.
+    async fn confirm_ceremony(
+        &self,
+        request: Request<proto::SignedCeremonyConfirmationRequest>,
+    ) -> Result<Response<proto::CeremonyConfirmationResponse>, Status> {
+        verify_kp_signature::<CeremonyConfirmationRequest, _>(request.get_ref())?;
```
dlukegordon: Could we check the signer fingerprint against `self.roster` before forwarding this? The signature is verified against the certificate carried in the same request, so anyone can generate a keypair and pass this gate. Every such request then creates a guardian control task, waits on the global control lock, and is only rejected against the roster after serialized PGP verification. A flood can therefore block the real KP confirmations. The ceremony roster has already been committed at this point, and the shared `RosterCache` already provides the same fail-closed gate used by the provisioning relay.

0xsiddharthks: Done in bd17cdcf3. `RosterCache::authorize` is now the one gate, used by the relay and by both forwarded KP-signed RPCs (`ConfirmCeremony` and `ProvisionerRotateCert`, which closes the old TODO there as well). Checked the enclave side too: `setup_new_key` writes `kp-shares/` and the `ceremony/` record before it advances to awaiting confirmations, so the roster is committed by the time a KP can confirm. Tests cover a rostered confirmation reaching the backend, an unrostered one denied before it, and the gate's not-ready / unavailable mappings.

crates/hashi-guardian-proxy/src/roster.rs
```diff
@@ -93,6 +95,66 @@ impl<L: LogStore> RosterCache<L> {
...
+                return Err(Status::failed_precondition(
+                    "no KP share log in the guardian bucket; run the key ceremony first",
+                ))
+            }
+            Err(e) => {
+                warn!(error = %format!("{e:#}"), "KP roster read failed");
+                return Err(Status::unavailable("KP roster unavailable; retry"));
+            }
+        };
+        if roster.contains(signer) {
```
dlukegordon: Could we refresh or revalidate on a cached roster miss here? Populated entries live for 60 seconds, but `SetupNewKey` and `RotateKpSet` commit a new roster through direct guardian calls, so the proxy never invalidates this cache. If the previous roster is cached, a newly added KP gets `PERMISSION_DENIED`; `key-provisioner ceremony` makes one confirmation call without retrying, so the ceremony remains incomplete until someone reruns it after the TTL. A cached `None` creates the same five-second race for the first ceremony. Please add a bounded, single-flight refresh on a miss so arbitrary unrostered requests cannot force one S3 read each, plus a regression test covering old cached roster -> new committed roster -> new KP admitted.

0xsiddharthks: Done in 6dda66ae6: a miss (unknown signer, or a cached "no ceremony yet") now re-reads the share log at most once per 5 s, single-flight under the same mutex. The miss budget is separate from the TTL read, so a KP added right after a normal read is admitted on its first call. Tests: old cached roster -> new committed roster -> new KP admitted, the first ceremony from a cached miss, and the bound plus its renewal.

dlukegordon (approved):

PR #1114, opened 2026-09-03
withdrawal: remove status enum field

bmwill:
Remove the status enum field to reduce work done in move as the status can always be derived from other fields that were set.


Bridgerz (approved):
LGTM

PR #1116, opened 2026-09-03
cli: refuse governance actions on executed proposals before the prompt; explain Move aborts

Bridgerz:
## Summary

Voting on a proposal that has already been executed produced this, after the CLI had printed the proposal's details and asked for confirmation:

```
Error: transaction not submitted: transaction simulation failed in command 0 (MOVE_ABORT): MoveAbort(MoveLocation { module: ModuleId { address: 0000...0002, name: Identifier("dynamic_field") }, function: ..., instruction: 0, function_name: Some("borrow_child_object_mut") }, 1) in command 0 at 0x...02::dynamic_field::borrow_child_object_mut
```

Executed proposals are not deleted: `proposal::execute` moves them from the active bag to the executed bag, where they stay for inspection. The CLI's only guard looked in both bags, so it found the proposal, and the Move `vote` entry borrows the active bag with no executed-bag check, so simulation ended in the framework's `dynamic_field` abort. Nothing in that line says "executed", and abort code 1 is also `proposal::EVoteAlreadyCounted`, so an operator who greps the Move source is misled. `remove-vote` and `execute` failed the same way.

## Changes

- **Pre-flight from data the CLI already has.** `HashiClient::locate_proposal` reports active, executed or missing from the scrape taken at startup. `vote`, `remove-vote` and `execute` refuse before printing details or prompting, in every mode including `--dry-run` and `--serialize-unsigned-transaction`:

  ```
  Error: proposal 0x... (UpdateConfig) has already been executed; voting is closed and there is nothing to submit. Run `hashi proposal view 0x...` for the final tally.
  ```

  A missing id names the RPC and the expiry path (proposals expire seven days after creation and can then be deleted) instead of the bare `Proposal not found`.
- **`proposal view` gains a Status line**: `Active (expires <date>)`, `Expired on <date> (...)`, or `Executed`. `list --detailed` shows the same line. `view` still accepts executed proposals.
- **Move aborts are decoded at the shared execute path** (`execute_or_simulate`, which every governance and validator command goes through). A clever `#[error]` constant is named, rendered and given a one-line hint (`EVoteAlreadyCounted: Vote already counted (this validator has already voted on the proposal)`, quorum not reached, not a committee member, version disabled, and so on). The bare `dynamic_field` code 1 is explained as the proposal having left the active bag while the command was running, which covers the race the pre-flight cannot. Matching is on module and constant name, never on the bare code. The raw SDK error stays underneath as `Caused by`.

Not changed: the Move `vote` and `remove_vote` entries still lack the executed-bag assert that `execute` and `delete_expired` have. Adding it (body-only, upgrade compatible) would give raw PTBs and older CLIs a clever `EProposalAlreadyExecuted` instead of the framework abort; that is a separate Move change, and the decoder here already handles both shapes.

## Tests

`crates/hashi/src/cli/commands/proposal_tests.rs` (the module's existing tests move there too): the location lookup, each refusal text pinned by id, type and remedy, the three status strings, and the decoder on hand-built protos, including that a hashi abort with code 1 and no clever payload is not mistaken for the bag miss, and that a non-abort failure is left alone.

`test_create_update_config_proposal` now executes its proposal and votes on it again, asserting the real chain abort (`dynamic_field` code 1 with no clever payload, or `EProposalAlreadyExecuted` once Move gains the assert) and that the decoder explains it. Full hashi unit suite, clippy and a warnings-denied doc build are clean.

### DIFF
--- a/crates/e2e-tests/src/e2e_flow.rs
+++ b/crates/e2e-tests/src/e2e_flow.rs
@@ -2399,6 +2399,87 @@ mod tests {
         );
         info!("Transaction succeeded: {}", response.transaction().digest());
 
+        // The single member's creation vote is the whole committee, so the
+        // proposal is executable at once. Execute it, then try to vote on it
+        // again: executed proposals move to the executed bag rather than being
+        // deleted, and `proposal::vote` borrows the active bag with no
+        // executed-bag check, so the chain answers with the framework's
+        // `dynamic_field` abort (code 1, no clever payload). The CLI's abort
+        // decoder relies on exactly that shape, so it is pinned here against
+        // the real chain, and the decoder is checked to explain it.
+        let proposal_id = hashi::cli::upgrade::extract_proposal_id_from_response(&response)?;
+        let execute_tx = hashi::cli::upgrade::build_execute_proposal_transaction(
+            hashi_ids,
+            hashi_isv,
+            proposal_id,
+            execute_package_id,
+            "update_config",
+        )?;
+        let exec_resp = executor.execute(execute_tx).await?;
+        assert!(
+            exec_resp.transaction().effects().status().success(),
+            "update_config::execute failed: {:?}",
+            exec_resp.transaction().effects().status()
+        );
+
+        let type_arg = sui_sdk_types::TypeTag::Struct(Box::new(sui_sdk_types::StructTag::new(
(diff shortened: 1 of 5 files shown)


jessiemongeon1:

bmwill (approved):

PR #1121, opened 2026-09-03
committee: bind certificate preimages to the Hashi object id

Bridgerz:
## Why

`verify_certificate` signs `intent || epoch || message` and nothing identifies the deployment (SEC-553). Two Hashi instances on the same Sui network with the same registrants have byte-identical committees at identical epochs, so a `ReconfigCompletion` certificate from one can complete genesis on the other and pin the wrong MPC key. The pre-mainnet republish (IOP-700) creates exactly that precondition on testnet: same validator cohort and BLS keys, old deployment still running from the maintenance branch.

## What

The signing preimage becomes `intent (u16 LE) || bcs(hashi object id) || bcs(epoch) || bcs(message)`, byte-identical on both sides.

**Move** (two choke points, verified exhaustive): `committee::verify_certificate` and the proof-of-possession check take a `hashi_id: address`. `hashi::verify` / `verify_with_committee` supply it from `&Hashi`; zero changes to the nine cert-verifying entry functions. `update_next_epoch_public_key` threads it three levels into the PoP path.

**Rust**: `signing_message`, `sign`/`proof_of_possession`, the three `Committee` verify entry points, `BlsSignatureAggregator`, and `try_from_parts` take the id. Validators sign with it (deposit confirmations, withdrawal messages, guardian requests, committee transitions, PoP), the leader aggregates with it, and `MpcManager` carries it for all ceremony certs (DKG, rotation, nonce, AVID).

**Guardian**: the enclave pins the id in `InitConfig`, so the `config_hash` KPs sign in their PI submissions covers it with no new authorization machinery. `verify_hashi_cert` checks every certificate against the pinned id. The id is exposed as an explicit `GuardianInfo` field, stamped into the genesis and committee-update S3 records, and kp-provision cross-checks the enclave-reported id against its own config. guardian-proxy is untouched (pure relay); `pb::Committee` / `move_types::Committee` stay BCS-identical to the on-chain struct.

## Regression tests

- Move: a certificate minted with another instance's object id fails `ESigVerification` (ticket requirement)
- Rust: preimage-layout lock test extended; new cross-id test covering verify entry points and aggregation
- Suites: Move 271/271, hashi 821/821, hashi-types + guardian crates 321/321, clippy clean

## Deployment notes

- Clean-chain only: stored certs (deposit approvals, committee handoffs) are re-verified after storage, so this cannot ship to a live chain. Lands on main and reaches testnet at the pre-mainnet republish; not cherry-pickable to the maintenance branch.
- Enclave binary + `config_hash` change: new PCR measurements at the next provisioning ceremony (already implied by the fresh-cycle republish).
- Launch note: pause or decommission the old testnet committee before the new genesis completes, or verify no old `ReconfigCompletion` cert exists for the new genesis epoch. Fully mitigated going forward once this lands.

Fixes IOP-657. Related: SEC-553. Unblocks IOP-700.


packages/hashi/sources/core/intent.move
```diff
@@ -3,11 +3,13 @@
...
-/// intent value, so a certificate produced for one message type can never
-/// verify as another — regardless of whether two types happen to share a BCS
-/// layout.
+/// The signing preimage is `intent (u16 LE) || bcs(hashi object id) ||
+/// bcs(epoch) || bcs(message)`. Every message type signed under the
+/// committee's keys carries a unique intent value, so a certificate produced
+/// for one message type can never verify as another — regardless of whether
+/// two types happen to share a BCS layout — and the Hashi object id binds
+/// every signature to one deployment, so a certificate minted for another
+/// Hashi instance can never verify here (SEC-553 / IOP-657).
```
bmwill: Can we remove references to issues?

packages/hashi/sources/core/committee/committee.move
```diff
@@ -250,10 +251,15 @@ public(package) fun verify_certificate<T>(
...
-    // Signing preimage: intent (u16 LE) || bcs(epoch) || bcs(message). The
-    // intent leads so the signed bytes are domain-tagged before anything else,
-    // separating message types signed under the same keys (see hashi::intent).
+    // Signing preimage: intent (u16 LE) || bcs(hashi_id) || bcs(epoch) ||
+    // bcs(message). The intent leads so the signed bytes are domain-tagged
+    // before anything else, separating message types signed under the same
+    // keys (see hashi::intent); the Hashi object id right after binds the
+    // certificate to this deployment, so a certificate minted for another
+    // Hashi instance (byte-identical committee, same epoch) can never verify
+    // here (SEC-553 / IOP-657).
```
bmwill: same here, we don't need the issue reference

crates/hashi-types/proto/sui/hashi/v1alpha/guardian_service.proto
```diff
@@ -132,6 +132,10 @@ message GuardianInfoData {
 
   // Digest of the optional GenesisState pinned during operator_init.
   optional bytes genesis_state_hash = 13;
+
+  // The Hashi shared-object id (32 bytes) this guardian serves (set after
+  // operator_init). Certificates verified by this enclave are bound to it.
+  optional bytes hashi_object_id = 14;
```
bmwill, line 138: @0xsiddharthks @mskd12 @dlukegordon 

When provisioning the guardian can we also ensure that the hashi object id is attested to by provisioners?

dlukegordon: > Yes. The KPs do not sign the raw object ID as a separate field, but they sign expected_config_hash, and that hash now commits to hashi_object_id. Each KP also independently checks the enclave-reported ID against its own configured ID before signing. The enclave then requires threshold signed submissions for its live config hash.

bmwill (approved):

PR #1123, opened 2026-09-03
committee: never authorize the zero address as a member's sender

Bridgerz:
## Problem

`committee_set::is_authorized` accepts any sender equal to the member's validator or operator address. A validator that "clears" its operator delegation by pointing it at `@0x0` would hand that member's authority to Sui system transactions, which run with `@0x0` as their sender.

## Fix

Reject `@0x0` before the address comparison. Not reachable today, since sui-core hardcodes the framework package, so this is defence in depth as agreed in the Certora thread (2026-07-31 finding #4, signed off by Zhou).

## Tests

- `test_zero_sender_never_authorized`: an undelegated member rejects the zero sender; after the delegation is cleared to `@0x0` the zero sender is still rejected while the validator key keeps its authority.
- `test_set_operator_address_rejects_zero_sender`: a zero-sender `set_operator_address` against a cleared delegation aborts.

Full Move suite: 273 passed.

Fixes IOP-595

### DIFF
--- a/packages/hashi/sources/core/committee/committee_set.move
+++ b/packages/hashi/sources/core/committee/committee_set.move
@@ -618,11 +618,15 @@ fun new_committee_from_voting_powers(
     )
 }
 
-/// True if the tx sender is authorized to act for this member — its validator
+/// True if the tx sender is authorized to act for this member: its validator
 /// key or its delegated operator key. The validator key always retains authority.
+///
+/// The zero address is never authorized. Sui system transactions run with `@0x0`
+/// as their sender, so an operator delegation that was "cleared" to `@0x0` must
+/// not hand them authority over the member.
 fun is_authorized(self: &MemberInfo, ctx: &TxContext): bool {
     let sender = ctx.sender();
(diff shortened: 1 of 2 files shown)


0xsiddharthks (approved):

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -618,11 +618,15 @@ fun new_committee_from_voting_powers(
...
+/// True if the tx sender is authorized to act for this member: its validator
 /// key or its delegated operator key. The validator key always retains authority.
+///
+/// The zero address is never authorized. Sui system transactions run with `@0x0`
+/// as their sender, so an operator delegation that was "cleared" to `@0x0` must
+/// not hand them authority over the member.
 fun is_authorized(self: &MemberInfo, ctx: &TxContext): bool {
     let sender = ctx.sender();
-    sender == self.validator_address || sender == self.operator_address
+    sender != @0x0 && (sender == self.validator_address || sender == self.operator_address)
```
bmwill, line 629: i think this is dumb but sure :)

PR #1124, opened 2026-09-03
[guardian-init] Add the KP-set rotation commands

0xsiddharthks:
## Summary

A KP-set rotation re-deals the ceremony key to a new set of key provisioners on a fresh ceremony-mode guardian. The enclave side (`RotateKpSet`) already existed; nothing could drive it. This adds the operator and KP commands, routes ceremony confirmations to the guardian being provisioned, and runs the whole flow in the local rig. Rehearsed on `testing` with real attestation, including a replacement guardian provisioned by the new set; the local rig (ceremony, 3-of-4 rotation, provision, reprovision) passes on the rebased branch.

## Changes

- `key-provisioner rotate-kp-set`: verify the ceremony guardian, decrypt this KP's share, sign a submission binding it to the proposed roster and n/t. The submission is a file; nothing in it is secret.
- `operator rotate-kp-set init | submit | wait`: OperatorInit and the session pin, then threshold-many submissions in one `RotateKpSet`, the dealt shares and logs cross-checked, and a wait for every new KP's confirmation. `wait` resumes an interrupted `submit`; `submit` refuses a guardian that already dealt.
- Config gets an optional `new_kp_roster`; `kp_roster` stays the dealt set. The dealt set is verified as a set (#1149); the proposed set is dealt in fingerprint order, so every KP signs the same roster whatever order their configs list.
- `key-provisioner ceremony` reads the guardian through the relay's `GetProvisioningTargetInfo` when the endpoint is the proxy; the proxy forwards `ConfirmCeremony` to the provisioning target. Whichever RPC answered must be a ceremony enclave.
- The steps shared with `operator ceremony` move into `ceremony.rs`.
- Local rig: `make ceremony` has every KP confirm (it hung since #1073), plus `make rotate-kp-set` and `make reprovision`.
- README: the new commands and a lost-key recovery runbook.


crates/hashi-guardian-proxy/src/forward.rs
zhouwfang, line 122: During a rotation this and the relay share one roster cache but front different enclaves, so once RotateKpSet commits an old KP gets PERMISSION_DENIED here for a call the still-active guardian would have accepted. README:303 says the serving guardian isn't touched until switchover, which is true of the enclave but not of this gate.

crates/hashi-guardian-init/README.md
```diff
@@ -224,6 +231,101 @@ cargo run -p hashi-guardian-init -- key-provisioner rotate-cert \
...
+## recovering a lost KP key
+
+A KP whose sole YubiKey is lost or unusable cannot `rotate-cert` (that needs
+the old key). Any `t` of the remaining KPs replace the whole set instead:
+
+1. Leave the serving guardian alone: it holds the key in memory and keeps
+   signing. Nothing below touches it until the switchover.
+2. Agree on the new set: `n`, `t` and one cert per KP (a fresh YubiKey for the
+   affected KP, or a different person). It goes in `new_kp_roster`;
+   `kp_roster` stays the dealt set. The operator and every KP use one config.
```
zhouwfang: One config can't cover the whole runbook, since steps 3 to 6 verify against the ceremony EIF and step 7 against the withdraw one, and those are different PCR0s at the same commit. Worth saying that current_build gets repointed before step 7, otherwise everyone hits BuildNotCurrent halfway through a lost-key recovery.

crates/hashi-guardian-init/README.md
```diff
@@ -224,6 +231,101 @@ cargo run -p hashi-guardian-init -- key-provisioner rotate-cert \
...
+   (no `--do-genesis`), `key-provisioner provision` by the new KPs,
+   `operator activate`, switch over, retire the old guardian.
+
+The rotation does not revoke the old shares: `kp-shares/{old seq}/` stays
+readable by the old certs, so `t` old keys could still reconstruct the key
+offline. If that many old keys may be compromised, run a new key ceremony and
+migrate on-chain instead.
+
+[`docker/hashi-guardian-local`](../../docker/hashi-guardian-local)'s
+`make rotate-kp-set` then `make reprovision` run this sequence end to end
```
zhouwfang: The rig sets no standby URL, so the ConfirmCeremony reroute never actually runs here outside the stub test. That's the riskiest change in the PR, so I'd soften this line rather than claim end to end.

crates/hashi-guardian-init/src/operator_rotate_kp_set.rs
```diff
@@ -0,0 +1,541 @@
...
+                btc_master_pubkey: old.btc_master_pubkey,
+            },
+            KpShareStateLogMessage::new(new_sharing_seq, 0, response.encrypted_shares),
+        )?;
+        live.validate_sharing_params(new_kp_set.num_shares, new_kp_set.threshold)?;
+        guardian
+            .verify_published(&live, new_kp_set.num_shares, new_kp_set.threshold)
+            .await?;
+        live.secret_sharing_instance
+    } else {
```
zhouwfang: This branch never reads submission_paths even though --submission is required, so a re-run with the wrong files still prints "KP-set rotation complete". The seq and pubkey checks it also drops are enclave-enforced, so those I'm less worried about.

crates/hashi-guardian-init/src/guardian_info.rs
```diff
@@ -42,6 +43,40 @@ pub async fn verified_provisioning_target_info(
...
+        .with_context(|| format!("invalid ceremony guardian endpoint {endpoint}"))?
+        .connect()
+        .await
+        .with_context(|| format!("connect to ceremony guardian at {endpoint}"))?;
+    match GuardianRelayServiceClient::new(channel.clone())
+        .get_provisioning_target_info(pb::GetProvisioningTargetInfoRequest {})
+        .await
+    {
+        Ok(response) => Ok(response.into_inner()),
+        Err(status) if status.code() == Code::Unimplemented => {
```
zhouwfang: Tonic maps any HTTP 404 to Unimplemented, so an ingress that exposes GuardianService but not GuardianRelayService silently falls back to the active guardian. Fails closed, but the KP gets an error pointing at the wrong thing.

crates/hashi-guardian-init/src/operator_rotate_kp_set.rs
```diff
@@ -0,0 +1,541 @@
...
+    let new_kp_set = cfg.require_new_kp_roster("operator rotate-kp-set")?;
+    new_kp_set.validate()?;
+    let guardian_s3 = hashi_guardian::resolve_s3_config(&cfg.guardian_s3).await?;
+    let certs_roster = cfg.kp_roster.load_certs_roster()?;
+    let new_certs_roster = new_kp_set.load_certs_roster()?;
+
+    let mut guardian = CeremonyGuardian::init(&cfg, &guardian_s3).await?;
+    ensure!(
+        guardian.lifecycle == CeremonyStage::OperatorInitialized.into(),
+        "guardian lifecycle is {:?}; a rotation was already submitted to it",
```
zhouwfang: This says a rotation was already submitted, but it fires just as readily for a guardian that finished a genesis ceremony.

docker/hashi-guardian-local/Makefile
```diff
@@ -19,15 +26,29 @@ help: ## Show this help
...
 
+rotate-kp-set: ## Re-deal the key to a new KP set on a fresh ceremony guardian (NEW_NUM_SHARES/NEW_THRESHOLD)
+	$(COMPOSE) --profile ceremony up -d --build --force-recreate ceremony
+	NEW_NUM_SHARES="$(NEW_NUM_SHARES)" NEW_THRESHOLD="$(NEW_THRESHOLD)" \
+	  $(COMPOSE) --profile init run --rm --build -e NEW_NUM_SHARES -e NEW_THRESHOLD \
+	    init /scripts/rotate-kp-set.sh
+	$(COMPOSE) --profile ceremony stop ceremony
+
+reprovision: ## Fresh withdraw guardian, provisioned by the dealt (rotated) KP set; no genesis
+	$(COMPOSE) up -d --force-recreate enclave
```
zhouwfang: This force-recreates the enclave then immediately runs provision, which has no connect retry, so --wait would help. Unrelated nit: rotate-kp-set.sh is 644 while the other scripts in that dir are 755.

crates/hashi-guardian-init/src/ceremony.rs
```diff
@@ -0,0 +1,239 @@
...
+        info!(
+            phase = "log cross-check",
+            "ceremony/ and kp-shares/ logs match the guardian's response",
+        );
+        Ok(())
+    }
+
+    /// Block until every dealt KP has confirmed and the guardian reports
+    /// `Completed`; a line a minute shows the wait is alive.
+    pub async fn wait_for_confirmations(&mut self) -> Result<()> {
```
zhouwfang, line 210: This waits forever and needs all n new KPs rather than the threshold, so one unreachable KP blocks Completed indefinitely. Probably fine with an operator watching, but the runbook should say interrupting is safe and re-running submit resumes.

crates/hashi-guardian-proxy/src/forward.rs
0xsiddharthks: The gate matches the enclave here: the active guardian's `provisioner_rotate_cert` also reads the latest `ceremony/` + `kp-shares/` from S3 and validates the share assignment against them, so once `RotateKpSet` commits, an old KP's rotate-cert is refused on both sides. The README's "untouched" wording was the error; it now says the serving guardian keeps signing while its KP-signed surface follows the committed roster.

crates/hashi-guardian-init/README.md
0xsiddharthks: Agreed, one config covers steps 3 to 6 only. Step 2 now pins `current_build` to the ceremony EIF (the build that dealt the current shares in `prev_builds`), and step 7 says to re-render with the withdraw EIF as `current_build`, the ceremony EIF moving to `prev_builds` since it wrote the logs everyone reads.

crates/hashi-guardian-init/README.md
0xsiddharthks: Softened. The rig reaches the ceremony guardian directly, so the README now credits the reroute to the proxy's `confirms_a_ceremony_on_the_provisioning_target` test and the `testing` rehearsal. Wiring a standby URL into the rig would mean restarting the proxy mid-flow; not worth the moving parts.

crates/hashi-guardian-init/src/operator_rotate_kp_set.rs
0xsiddharthks: Split it: `submit` now refuses a guardian that already dealt, and a new `operator rotate-kp-set wait` resumes the confirmation wait after verifying the logs against `new_kp_roster`. The files are never silently ignored; the rig runs `wait` on the completed rotation.

crates/hashi-guardian-init/src/guardian_info.rs
0xsiddharthks: Kept the fallback (a bare guardian in the rig needs it), but its answer must now be a ceremony enclave. The error names the RPC that answered and the lifecycle, so a hidden relay fails as "GetGuardianInfo at <endpoint> answers for a guardian in lifecycle Withdraw(…)" instead of a downstream mismatch.

crates/hashi-guardian-init/src/operator_rotate_kp_set.rs
0xsiddharthks: Reworded to "expected operator_initialized, a fresh ceremony guardian that has dealt nothing (operator rotate-kp-set wait resumes a submitted rotation)", shared by `init` and `submit`.

docker/hashi-guardian-local/Makefile
0xsiddharthks: Added `--wait` to every enclave (re)create (`ceremony`, `rotate-kp-set`, `reprovision`; the services have health checks), and `rotate-kp-set.sh` is 755 now.

crates/hashi-guardian-init/src/ceremony.rs
0xsiddharthks: That is the enclave's rule (`confirm_ceremony` completes only at n confirmations), so the wait stays unbounded. The README now says the wait has no timeout, interrupting is safe once the batch was accepted, and `operator rotate-kp-set wait` resumes.

crates/hashi-guardian-init/src/operator_rotate_kp_set.rs
```diff
@@ -0,0 +1,571 @@
...
+        .verify_recipient_set(&new_certs_roster)?;
+    info!(
+        phase = "rotate_kp_set",
+        lifecycle = ?guardian.info.lifecycle,
+        sharing_seq = logged.secret_sharing_instance.sharing_seq(),
+        "the guardian dealt this config's new_kp_roster; resuming the wait",
+    );
+
+    guardian.wait_for_confirmations().await?;
+    report(&logged.secret_sharing_instance);
```
dlukegordon, line 213: The latest ceremony logs are not bound to the pinned guardian session. If the endpoint points to an older completed ceremony guardian on the same build/bucket, these roster checks can accept the newer rotation’s logs, then `wait_for_confirmations()` immediately accepts the older guardian’s `Completed` lifecycle. The command prints the newer rotation’s sequence as complete even while its guardian awaits confirmations. Please preserve the ceremony log’s writing-session provenance and require it to match the pinned session, or verify an authenticated ceremony digest from the live guardian before accepting completion.

crates/hashi-guardian-init/README.md
```diff
@@ -221,6 +230,121 @@ cargo run -p hashi-guardian-init -- key-provisioner rotate-cert \
...
+   against the same bucket, then `operator rotate-kp-set init`.
+4. Any `t` current KPs: `key-provisioner rotate-kp-set`, each sending the
+   operator its submission file. The lost key takes no part.
+5. Operator: `operator rotate-kp-set submit` with the files. It waits for the
+   new KPs (`wait` resumes if interrupted).
+6. Every new KP: `key-provisioner ceremony`, with `kp_roster` set to the new
+   set.
+7. The withdraw EIF has a different PCR0 at the same commit, so re-render:
+   `current_build` becomes the withdraw EIF and the ceremony EIF, which wrote
+   the logs everyone reads, moves to `prev_builds`. Then flip the standby slot
```
dlukegordon: The prescribed same-commit configuration is rejected: `PcrAllowlist::new` permits only one entry per `git_revision`, so putting these two EIFs in `current_build` and `prev_builds` fails during YAML loading. Omitting the ceremony entry also fails, because its persisted attestation then resolves to the withdraw entry’s different PCR0. Please document a workable sequence using distinct actual build revisions, or support separate build identities before recommending this sequence.

crates/hashi-guardian-init/README.md
```diff
@@ -221,6 +230,121 @@ cargo run -p hashi-guardian-init -- key-provisioner rotate-cert \
...
+   to a withdraw-mode guardian: `operator provision` (no `--do-genesis`),
+   `key-provisioner provision` by the new KPs, `operator activate`, switch
+   over, retire the old guardian.
```
dlukegordon: `operator activate` cannot succeed while the old serving guardian remains running. Activation requires all other sessions to be quiet for ten minutes, but the old guardian continues heartbeating every minute, so this ordering ends in an activation timeout. Please move stopping the old guardian before activation and explicitly account for the quiet period before switching traffic to the replacement.

docker/hashi-guardian-local/scripts/rotate-kp-set.sh
```diff
@@ -0,0 +1,78 @@
...
+
+echo "== operator rotate-kp-set submit (waits for every new KP's confirmation) =="
+render_config "${endpoint}" ""
+# Kept aside: the KP renders below overwrite ${CONFIG}, and `wait` reuses it.
+operator_config="${WORK}/operator-rotate.yaml"
+cp "${CONFIG}" "${operator_config}"
+hashi-guardian-init operator rotate-kp-set submit --config "${operator_config}" "${submissions[@]}" \
+  > "${WORK}/operator-rotate.out" 2> "${WORK}/operator-rotate.log" &
+operator=$!
+wait_for_line "${WORK}/operator-rotate.log" "waiting for every key provisioner" 120 "${operator}"
```
dlukegordon: Readiness depends on an INFO event that the supported `RUST_LOG` override can suppress. With `RUST_LOG=warn`, the operator publishes the rotation and waits for confirmations, but `wait_for_line` times out after 120 seconds before the script runs any confirmations. `ceremony.sh:40` has the same dependency. Please poll actual ceremony readiness, or explicitly enable the required event for these background operator commands rather than inheriting an arbitrary logging filter.

dlukegordon (commented):
Source-only review of the current diff and relevant enclave/storage contracts. No tests, builds, or runtime flows were run.

crates/hashi-guardian-init/src/operator_rotate_kp_set.rs
0xsiddharthks: Right, `wait` only bound the logs to the build. It now reads the latest ceremony together with the session that wrote its `ceremony/` record (the reader verifies that record under the session's attestation) and requires it to be the pinned session, so an endpoint pointed at an older ceremony guardian fails with "dealt by session X, not by the pinned session Y" instead of waiting on the wrong lifecycle. While here: `submit` and `wait` no longer go through the `OperatorInit` path at all, only `init` initializes a guardian.

crates/hashi-guardian-init/README.md
0xsiddharthks: A real ceremony enclave reports its revision as `<sha>-ceremony` (`Enclave::reported_git_revision`), precisely so the same-commit withdraw build can sit beside it in the allowlist. Today's run on `testing` was configured that way: `63c5e67…-ceremony` as `current_build` for steps 3 to 6, then the withdraw build as `current_build` with the `-ceremony` entry in `prev_builds` for step 7. The README never said so; steps 2 and 7 and the sample yaml now name both revisions.

crates/hashi-guardian-init/README.md
0xsiddharthks: Yes, that order was backwards. Step 7 now stops the old guardian before `operator activate`, says activate needs the other sessions quiet for 10 minutes and retries until then, and switches traffic after activation. That is the order we ran today: stopped at 18:00, activated at 18:09.

docker/hashi-guardian-local/scripts/rotate-kp-set.sh
0xsiddharthks: Fixed. Both backgrounded operator commands now go through one `start_operator` helper in lib.sh that runs them with `RUST_LOG="${RUST_LOG:-info},hashi_guardian_init=info"`, so the line the scripts gate on is emitted whatever filter the caller set. Polling the guardian's lifecycle instead would need a new probe in the CLI; not worth it for the rig.

dlukegordon (approved):
Reviewed the latest changes and verified that the previous feedback is addressed. No remaining blockers found in the source review. Tests, builds, and deployment flows were not run as part of this review.

PR #1125, opened 2026-09-03
Consume the sizers decided dealer set in the AVID nonce party phase

zhouwfang:
#531

### DIFF
--- a/crates/hashi/src/mpc/service.rs
+++ b/crates/hashi/src/mpc/service.rs
@@ -548,6 +548,7 @@ impl MpcService {
             .inner
             .mpc_manager()
             .ok_or_else(|| anyhow::anyhow!("MpcManager not initialized"))?;
+        mpc_manager.read().unwrap().ensure_manager_epoch(epoch)?;
         let signer = self.inner.config.operator_private_key()?;
         let p2p_channel =
             RpcP2PChannel::new(onchain_state.clone(), epoch, MPC_LABEL_NONCE_GENERATION);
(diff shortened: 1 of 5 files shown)


crates/hashi/src/mpc/service.rs
```diff
@@ -1386,19 +1409,21 @@ impl MpcService {
...
+                let admitted = mpc_manager
                     .read()
                     .unwrap()
-                    .avid_admitted_nonce_dealers(&avid_certs, settled_cutoff_ms)
-                    .cutoff_ms;
-                let mut prefetched = PrefetchedTobChannel::new(avid_certs.into_inner())
-                    .with_supersede_check(onchain_state.clone(), epoch);
-                let outcome = MpcManager::run_nonce_party_phase(
+                    .avid_admitted_nonce_dealers(&avid_certs, settled_cutoff_ms)?;
+                if !admitted.floor_reached() {
```
dlukegordon, line 1420: Recheck supersession before recording this as a below-floor batch. `fetch_final_nonce_certs(..., false)` checks the epoch before `verified_nonce_certs(...).await`, so a reconfiguration can start during verification and this branch will increment a dealer-shortage counter and return `NotEnoughParticipants` for stale work. The caller eventually notices the pending reconfiguration, but the new metric is already misattributed. A post-verification supersession check in `fetch_final_nonce_certs` would cover both the no-cutoff and confirmed-cutoff return paths.

dlukegordon (approved):

PR #1126, opened 2026-09-04
Refuse gRPC callers that map to no registered validator

zhouwfang:
#918

### DIFF
--- a/crates/hashi/src/grpc/metrics_layer.rs
+++ b/crates/hashi/src/grpc/metrics_layer.rs
@@ -14,7 +14,9 @@
 //! observes traffic and writes into those metrics.
 
 use std::borrow::Cow;
+use std::collections::HashSet;
 use std::sync::Arc;
+use std::sync::OnceLock;
 use std::time::Instant;
 
 use axum::http;
@@ -278,18 +280,7 @@ impl callback::ResponseHandler for ResponseHandler {
     fn on_response(&mut self, response: &response::Parts) {
         const GRPC_STATUS: http::HeaderName = http::HeaderName::from_static("grpc-status");
 
-        let status = if response
-            .headers
-            .get(&http::header::CONTENT_TYPE)
-            .is_some_and(|content_type| {
-                content_type
-                    .as_bytes()
-                    // check if the content-type starts_with 'application/grpc' in order to
-                    // consider this as a gRPC request. A prefix comparison is done instead of a
-                    // full equality check in order to account for the various types of
-                    // content-types that are considered as gRPC traffic.
-                    .starts_with(tonic::metadata::GRPC_CONTENT_TYPE.as_bytes())
-            }) {
+        let status = if super::is_grpc_content_type(&response.headers) {
             let code = response
                 .headers
                 .get(&GRPC_STATUS)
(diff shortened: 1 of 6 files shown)


crates/hashi/src/lib.rs
```diff
@@ -1858,19 +1858,57 @@ mod test {
...
-            .into_inner();
-        assert_eq!(resp.server.as_deref(), Some("unknown/unknown"));
+            .expect_err("no gRPC method is served to an unregistered caller");
+        assert_eq!(status.code(), tonic::Code::PermissionDenied, "{status:?}");
+
+        let oversized = hashi_types::proto::SendMessagesRequest {
+            epoch: Some(0),
+            messages: Some(
+                hashi_types::proto::send_messages_request::Messages::DkgMessage(
+                    sui_rpc::proto::sui::rpc::v2::Bcs::serialize(&vec![0u8; 8 * 1024 * 1024])
```
dlukegordon: This 8 MiB request is below Hashi's 32 MiB server decode limit. If authorization happened after tonic decoding, `send_messages` would decode the whole payload and its existing `authenticate_caller` check would still return `PermissionDenied`, so this test would pass without proving pre-decode refusal. Please configure a smaller decode limit for this test, or send more than the configured limit, so the failure would otherwise be `ResourceExhausted`.

dlukegordon (commented):
Three findings: the pre-decode regression test does not cross the configured limit, reflection paths are missing from metric normalization, and the refusal metric overstates what it counts.

crates/hashi/src/grpc/metrics_layer.rs
```diff
@@ -362,22 +353,42 @@ fn extract_path(request: &request::Parts, role: Role) -> Cow<'static, str> {
     }
 }
 
+fn known_grpc_paths() -> &'static HashSet<Box<str>> {
+    static PATHS: OnceLock<HashSet<Box<str>>> = OnceLock::new();
+    PATHS.get_or_init(|| {
+        use prost::Message as _;
+
+        let mut paths = HashSet::new();
+        for encoded in [
```
dlukegordon, line 362: Both reflection services are mounted by this server, but their descriptors are missing here. A valid `ServerReflectionInfo` call therefore gets the wildcard Axum path label and is indistinguishable from an unknown reflection method. Please include the v1 and v1alpha `tonic_reflection` descriptor sets so every served gRPC method receives its exact path label.

crates/hashi/src/metrics.rs
```diff
@@ -485,6 +486,12 @@ impl Metrics {
                 registry,
             )
             .unwrap(),
+            grpc_unknown_caller_refused_total: register_int_counter_with_registry!(
+                "hashi_grpc_unknown_caller_refused_total",
+                "Requests refused before the body was decoded, because no registered validator could be resolved: no client certificate, an unparseable one, or one absent from the member registry. Not counted in hashi_requests.",
```
dlukegordon: An unparseable certificate never reaches this counter: `ClientCertVerifier` rejects it during the TLS handshake before `require_known_validator` runs. Please remove that case from the HELP text, or add separate TLS-handshake instrumentation if it needs to be observable.

dlukegordon (approved):
All review findings are addressed in the current revision.

PR #1127, opened 2026-09-04
Agree on the nonce dealer output sequence before deriving presigs

zhouwfang:
#193

### DIFF
--- a/crates/hashi/src/mpc/signing.rs
+++ b/crates/hashi/src/mpc/signing.rs
@@ -152,6 +152,7 @@ struct SigningPoolState {
 struct PrefetchedBatch {
     batch_index: u32,
     pool: Vec<Option<(Vec<S>, G)>>,
+    identity: PresigBatchIdentity,
 }
 
 pub struct SigningManager {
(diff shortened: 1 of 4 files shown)


crates/hashi/src/mpc/signing.rs
```diff
@@ -4071,31 +4193,25 @@ mod tests {
...
-    /// batch-19 refill result was relabeled as batch 20, binding stale
-    /// nonces to fresh global presig indices).
     #[tokio::test]
     async fn test_stale_prefetch_is_not_relabeled_as_next_batch() {
         let mut setup = SigningTestSetup::new(4);
         let batch_size = setup.managers[0].initial_presig_count() as u64;
         setup.refill_rx.borrow_and_update();
 
-        // Force a stale prefetch (tagged with the already-installed batch 0)
-        // past the `set_next_batch` guard, straight into the state.
```
bmwill, line 4095: are the removal of these comments intended?

zhouwfang: Yes, I keep comments for invariants, assumptions and correctness arguments rather than obvious narration.

bmwill (approved):
lgtm awesome to have a single codepath for processing the batches now

PR #1130, opened 2026-09-04
[docs] render crawled search results as text, not HTML

0xsiddharthks:
## Summary

Fixes [SEC-568](https://linear.app/mysten-labs/issue/SEC-568/hashi-docs-site-stored-xss-and-quadratic-backtracking-regex): docs search decoded HTML entities in Algolia highlight values before passing them to `dangerouslySetInnerHTML`, and cleaned tooltip text with a backreference regex that backtracks quadratically on long hierarchy labels. Results now render as React text, and the regex is replaced with a linear scan. The indices live on a shared Algolia app and are crawled from sites that take outside contributions; today's records are entity-escaped, so none of the 2,128 live hits I sampled would have become markup, but one record with a raw `<` would have.

## Changes

- `HitSnippet` renders a hit's highlight value as React text: it splits on `<mark>`, then decodes both layers of escaping, so short snippets no longer show raw entities like `Coin&lt;T&gt;`.
- No `dangerouslySetInnerHTML` is left in the search components; the empty-results links are now JSX.
- `getSnippetParts` replaces `truncateAtWord`: it counts visible characters and no longer drops the last one when there is no space to cut at.
- `cleanTooltipText` is a linear scan over word runs; it matches the old regex for glossary labels up to 64 characters.
- `parseHitUrl` keeps non-http(s) hit URLs out of `href`, and a malformed URL no longer throws from `new URL` mid-render.


dlukegordon (approved):
Reviewed at 75aa52854f288a804b06b80db5c43c88bf69e3fe. No actionable issues found in the complete diff, affected search consumers, or InstantSearch escaping behavior. Source-only review; no tests or builds run.

PR #1133, opened 2026-09-04
Prune previous-epoch key material on nodes that have left the committee

zhouwfang:
#95


dlukegordon (approved):
Looks good. We might want to add a regression test for the NoRole path: after a node has left both the current and previous committees, completing a later reconfiguration should prune eligible old MPC data without requesting a backup.

PR #1134, opened 2026-09-08
DNS - reviewing the dkg & rotation flows, focusing on communication

benr-ml:


zhouwfang (commented):
Thanks for the comments! I went through all, and I think only the TLS one (grpc/mod.rs:332) meets the must-fix bar I've been using (loss of funds, permanent wedge, broken safety property), no?

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -361,6 +361,8 @@ public(package) fun clear_resignation(
     self.member_mut(validator_address).resigned = false;
 }
 
+// TODO(fix): if called after abort_reconfig, it may reuse the same epoch as the aborted one (while the rest of the system assumes, e.g., there is one invocation of key rotation per epoch in several places)
+// is that the expected behavior?
```
zhouwfang, line 365: Not expected. We're changing abort so it only works once Sui's epoch has moved past the pending one, which means the number can't come back. Nothing to do here beyond that.

crates/hashi/src/grpc/mod.rs
```diff
@@ -321,6 +329,18 @@ pub(super) fn is_grpc_content_type(headers: &http::HeaderMap) -> bool {
         })
 }
 
+// TODO(fix): the caller identity is resolved purely from the peer's TLS
```
zhouwfang, line 332: I think this is worse than denial. Nonce dealings go through send_messages too, so a few colluding members each shadowing a big one can stop presigs, and rotation even sooner. It costs one tx per epoch to keep up and the victim never finds out. I have it as must-fix. The plan is a PoP plus a uniqueness check on the TLS key. WDYT?

crates/hashi-types/src/committee.rs
```diff
@@ -48,6 +48,18 @@ pub type EncryptionPrivateKey = fastcrypto_tbls::ecies_v1::PrivateKey<Encryption
 pub type EncryptionPublicKey = fastcrypto_tbls::ecies_v1::PublicKey<EncryptionGroupElement>;
 
 /// A thin wrapper around min_pk::BLS12381PrivateKey needed to implement Clone.
+// TODO[fix]: why do we need implicit clone and Serialize (instead of moving?).
```
benr-ml, line 51: @zhouwfang I think this one is important

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2745,6 +2745,10 @@ impl MpcManager {
         }
     }
 
+    // TODO(fix): `dealer` is never checked against the committee (but any registered member can call it).
```
benr-ml, line 2748: it can be critical if we have a bug in one of the depending flows

packages/hashi/sources/core/cert_submission.move
```diff
@@ -244,6 +244,14 @@ fun submit_stamped_cert_internal(
 
 fun assert_can_submit(hashi: &Hashi, epoch: u64, dealer: address, ctx: &TxContext) {
     hashi.versioning().assert_version_enabled();
+    // TODO(fix): `member_authorized` only checks that `dealer` is a registered
```
benr-ml, line 247: i'm fine with this one as "defense in depth"

crates/hashi-types/src/committee.rs
zhouwfang: @benr-ml Done in #1150, apart from zeroize. The key types already zeroize on drop, and wrapping the DB serialization buffer would add a plaintext copy rather than remove one, because fjall takes ownership of the Vec. The copies that aren't zeroized are in backup.rs and metrics_push, neither of them must-fix, and they're already tracked as a separate piece of work.

crates/hashi/src/grpc/mod.rs
zhouwfang: @benr-ml Done in #1151

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: @benr-ml Done in #1152

zhouwfang:
There is no must-fix item left from this PR. Will address the rest when free.

PR #1137, opened 2026-09-08
Expire old database backups

dlukegordon:
- Run retention on epoch housekeeping, including for departed/NoRole validators that must not create new archives. The existing serialized backup worker sweeps before the write-policy and config-path gates.
- Delete archives older than 14 days while always preserving the newest archive. Generation and parsing share the authoritative filename prefix and timestamp/suffix format.
- Continue cleanup after individual entry failures. Within backup-dir, expire abandoned .hashi-backup-* files and .hashi-restore-* directories after 14 days by modification time, preserving recent entries and symlinks. Completed restores, .hashi-db-restore-* directories, and restore output elsewhere remain operator-managed.
- Finalize and sync staged encrypted backups before no-clobber publication. Sync the destination directory afterward on a best-effort basis; failure logs at ERROR without failing the already-published backup.
- Require an explicit backup-dir; generated validator configs use node-specific paths outside database storage.
- Verification: 49 backup-related tests and the resigned-member end-to-end test pass, including retention after NoRole and cleanup without a config path. Workspace Clippy and make fmt completed. Existing archive publication, retention boundary, staging safety, and restore/rejoin coverage retained.
- Fixes [IOP-713](https://linear.app/mysten-labs/issue/IOP-713/add-expiry-on-db-backups).

### DIFF
--- a/crates/e2e-tests/src/hashi_network.rs
+++ b/crates/e2e-tests/src/hashi_network.rs
@@ -584,7 +584,9 @@ impl HashiNetworkBuilder {
             config.bitcoin_chain_id = Some(hashi::constants::BITCOIN_REGTEST_CHAIN_ID.to_string());
             config.sui_chain_id = service_info.chain_id.clone();
             config.screener_endpoint = Some(screener_endpoint.clone());
-            config.db = Some(dir.join(validator_address.to_string()));
+            let node_name = validator_address.to_string();
+            config.backup_dir = dir.join("backups").join(&node_name);
+            config.db = Some(dir.join(node_name));
             configs.push(config);
         }
 
--- a/crates/hashi/src/backup.rs
+++ b/crates/hashi/src/backup.rs
@@ -41,7 +41,9 @@ pub const BACKUP_FILE_NAME_PREFIX: &str = "hashi-backup";
 pub const BACKUP_MANIFEST_FILE_NAME: &str = "hashi-backup-manifest.toml";
 pub const DB_SNAPSHOT_TAR_PREFIX: &str = "hashi-db-snapshot";
 
-const BACKUP_FILE_NAME_FORMAT: &str = "hashi-backup-%Y%m%dT%H%M%SZ.tar.asc";
+const BACKUP_FILE_NAME_SUFFIX_FORMAT: &str = "-%Y%m%dT%H%M%SZ.tar.asc";
+const BACKUP_STAGING_FILE_NAME_PREFIX: &str = ".hashi-backup-";
+pub(crate) const RESTORE_STAGING_DIR_NAME_PREFIX: &str = ".hashi-restore-";
 const BACKUP_RETENTION: jiff::SignedDuration = jiff::SignedDuration::from_hours(14 * 24);
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
@@ -51,7 +53,7 @@ pub enum BackupArchiveFormat {
 }
 
 enum BackupRequest {
-    EpochChanged(u64),
+    EpochChanged { epoch: u64, write_backup: bool },
 }
 
 #[derive(Clone, Debug)]
@@ -60,10 +62,13 @@ pub struct BackupHandle {
 }
 
 impl BackupHandle {
-    pub fn backup_after_epoch_change(&self, epoch: u64) {
+    pub fn backup_after_epoch_change(&self, epoch: u64, write_backup: bool) {
         if self
             .sender
-            .send(BackupRequest::EpochChanged(epoch))
+            .send(BackupRequest::EpochChanged {
+                epoch,
+                write_backup,
+            })
             .is_err()
         {
             warn!(
@@ -100,10 +105,13 @@ impl BackupService {
     async fn run(mut self) {
         while let Some(request) = self.receiver.recv().await {
             match request {
-                BackupRequest::EpochChanged(epoch) => {
+                BackupRequest::EpochChanged {
+                    epoch,
+                    write_backup,
+                } => {
                     let hashi = self.inner.clone();
                     match tokio::task::spawn_blocking(move || {
-                        hashi.backup_after_epoch_change(epoch)
+                        hashi.backup_after_epoch_change(epoch, write_backup)
                     })
                     .await
                     {
@@ -251,8 +259,21 @@ pub fn encrypt_files_to_pgp_archive(
     recipient: &PgpPublicCert,
     output_path: &Path,
 ) -> Result<()> {
-    let output = create_file_strict(output_path)?;
-    let mut encrypted = armored_encrypt_writer(output, recipient)?;
+    // Stage beside the destination so retention sees only fully finalized archives.
+    let parent = output_path
+        .parent()
+        .filter(|parent| !parent.as_os_str().is_empty())
+        .unwrap_or_else(|| Path::new("."));
+    let mut staging = tempfile::Builder::new()
+        .prefix(BACKUP_STAGING_FILE_NAME_PREFIX)
+        .tempfile_in(parent)
+        .with_context(|| {
+            format!(
+                "Failed to create backup staging file in {}",
+                parent.display()
+            )
+        })?;
+    let mut encrypted = armored_encrypt_writer(staging.as_file_mut(), recipient)?;
     {
         let mut archive = tar::Builder::new(&mut encrypted);
         append_backup_manifest(&mut archive, manifest)?;
@@ -272,6 +293,21 @@ pub fn encrypt_files_to_pgp_archive(
         archive.finish()?;
     }
     encrypted.finalize()?;
+    staging
+        .as_file()
+        .sync_all()
+        .context("Failed to sync backup staging file")?;
+    staging
+        .persist_noclobber(output_path)
+        // Drop the temporary file even if the caller retains the error.
+        .map_err(|e| e.error)
+        .with_context(|| format!("Failed to publish backup {}", output_path.display()))?;
+    if let Err(error) = File::open(parent).and_then(|directory| directory.sync_all()) {
+        error!(
+            directory = %parent.display(),
+            "Backup published, but syncing its directory failed: {error}"
+        );
+    }
 
     Ok(())
 }
@@ -367,38 +403,120 @@ pub fn encrypted_backup_file_name() -> PathBuf {
     // ISO 8601 basic format in UTC, e.g. 20260409T230419Z. Compact, sorts
     // lexicographically, and contains no characters that need escaping on any
     // common filesystem.
-    jiff::Timestamp::now()
-        .to_zoned(jiff::tz::TimeZone::UTC)
-        .strftime(BACKUP_FILE_NAME_FORMAT)
-        .to_string()
-        .into()
+    format!(
+        "{BACKUP_FILE_NAME_PREFIX}{}",
+        jiff::Timestamp::now()
+            .to_zoned(jiff::tz::TimeZone::UTC)
+            .strftime(BACKUP_FILE_NAME_SUFFIX_FORMAT)
+    )
+    .into()
 }
 
-/// Remove expired local archives, preserving the backup that just completed.
-pub(crate) fn cleanup_old_backups(
-    output_dir: &Path,
-    current_backup: &Path,
-    now: jiff::Timestamp,
-) -> Result<()> {
+/// Remove expired archives and abandoned staging entries, always keeping the newest archive.
+pub(crate) fn cleanup_old_backups(output_dir: &Path, now: jiff::Timestamp) -> Result<()> {
     let cutoff = now.checked_sub(BACKUP_RETENTION)?;
-    for entry in fs::read_dir(output_dir)
-        .with_context(|| format!("Failed to read backup directory {}", output_dir.display()))?
-    {
-        let entry = entry?;
+    let staging_cutoff: std::time::SystemTime = cutoff.into();
+    let entries = match fs::read_dir(output_dir) {
+        Ok(entries) => entries,
+        Err(error) if error.kind() == ErrorKind::NotFound => return Ok(()),
+        Err(error) => {
+            return Err(error).with_context(|| {
+                format!("Failed to read backup directory {}", output_dir.display())
+            });
+        }
+    };
+    let mut newest: Option<(jiff::Timestamp, PathBuf)> = None;
+    for entry in entries {
+        let entry = match entry {
+            Ok(entry) => entry,
+            Err(error) => {
+                warn!(
+                    directory = %output_dir.display(),
+                    "Failed to read backup directory entry: {error}"
+                );
+                continue;
+            }
+        };
         let file_name = entry.file_name();
+        let file_name = file_name.as_encoded_bytes();
+        let restore_staging = file_name.starts_with(RESTORE_STAGING_DIR_NAME_PREFIX.as_bytes());
+        if restore_staging || file_name.starts_with(BACKUP_STAGING_FILE_NAME_PREFIX.as_bytes()) {
+            let path = entry.path();
+            // DirEntry::metadata does not follow symlinks.
+            let modified = entry.metadata().and_then(|metadata| {
+                if (restore_staging && metadata.is_dir())
+                    || (!restore_staging && metadata.is_file())
+                {
+                    metadata.modified().map(Some)
+                } else {
+                    Ok(None)
+                }
+            });
+            match modified {
+                Ok(Some(modified)) if modified < staging_cutoff => {
+                    let removed = if restore_staging {
+                        fs::remove_dir_all(&path)
+                    } else {
+                        fs::remove_file(&path)
+                    };
+                    if let Err(error) = removed {
+                        warn!(
+                            path = %path.display(),
+                            "Failed to remove expired staging entry: {error}"
+                        );
+                    }
+                }
+                Ok(_) => {}
+                Err(error) => {
+                    warn!(
+                        path = %path.display(),
+                        "Failed to read staging entry metadata: {error}"
+                    );
+                }
+            }
+            continue;
+        }
+        let Some(suffix) = file_name.strip_prefix(BACKUP_FILE_NAME_PREFIX.as_bytes()) else {
+            continue;
+        };
         let Ok(created_at) =
-            jiff::civil::DateTime::strptime(BACKUP_FILE_NAME_FORMAT, file_name.as_encoded_bytes())
+            jiff::civil::DateTime::strptime(BACKUP_FILE_NAME_SUFFIX_FORMAT, suffix)
                 .and_then(|datetime| datetime.to_zoned(jiff::tz::TimeZone::UTC))
         else {
             continue;
         };
-        if created_at.timestamp() >= cutoff || !entry.file_type()?.is_file() {
+        let path = entry.path();
+        let file_type = match entry.file_type() {
+            Ok(file_type) => file_type,
+            Err(error) => {
+                warn!(
+                    path = %path.display(),
+                    "Failed to read backup file type: {error}"
+                );
+                continue;
+            }
+        };
+        if !file_type.is_file() {
             continue;
         }
-        let path = entry.path();
-        if path != current_backup {
-            fs::remove_file(&path)
-                .with_context(|| format!("Failed to remove expired backup {}", path.display()))?;
+        let candidate = (created_at.timestamp(), path);
+        // Only delete an archive once another valid regular archive is known to be newer.
+        let (created_at, path) = match newest.as_mut() {
+            Some(newest) if candidate.0 > newest.0 => std::mem::replace(newest, candidate),
+            Some(_) => candidate,
+            None => {
+                newest = Some(candidate);
+                continue;
+            }
+        };
+        if created_at >= cutoff {
+            continue;
+        }
+        if let Err(error) = fs::remove_file(&path) {
+            warn!(
+                path = %path.display(),
+                "Failed to remove expired backup: {error}"
+            );
         }
     }
     Ok(())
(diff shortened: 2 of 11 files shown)


bmwill (approved):

crates/hashi/src/backup.rs
```diff
@@ -364,11 +367,41 @@ pub fn encrypted_backup_file_name() -> PathBuf {
...
+        else {
+            continue;
+        };
+        if created_at.timestamp() >= cutoff || !entry.file_type()?.is_file() {
+            continue;
+        }
+        let path = entry.path();
+        if path != current_backup {
+            fs::remove_file(&path)
+                .with_context(|| format!("Failed to remove expired backup {}", path.display()))?;
```
zhouwfang: This ? bails out of the whole sweep on the first entry we can't delete, and so do the ones on entry at :387 and entry.file_type() at :395. We always reach the bad entry and always abort there, so a pass never completes and it fails the same way next epoch. Logging per entry and carrying on would cover it.

crates/hashi/src/lib.rs
```diff
@@ -361,6 +361,16 @@ impl Hashi {
             output = %output_path.display(),
             "Automatic backup completed after epoch change",
         );
+        if let Err(error) = crate::backup::cleanup_old_backups(
```
zhouwfang: Cleanup only runs on the success path, so if backups start failing the old archives stop being deleted too, which is exactly when we'd want them gone. Sweeping before the save, or on both outcomes, avoids that.

design/docs/node-backup.mdx
```diff
@@ -258,6 +258,11 @@ backup-pgp-cert = "/path/to/hashi-backup-cert.asc"
 You can also configure the location where Hashi saves backups with the field
 `backup-dir`.
 
+After each successful automatic backup, Hashi deletes local backup archives older
+than 14 days, based on their filenames. The new backup and unrelated files are
+preserved. Failed or manual backups do not trigger cleanup, so expired archives
```
zhouwfang: It reads as though manual backups are exempt, but backup save names its output the same way, so one written into backup_dir gets deleted at 14 days like anything else. Worth saying so.

zhouwfang (changes requested):

crates/hashi/src/lib.rs
```diff
@@ -361,6 +361,16 @@ impl Hashi {
...
+            self.config.backup_dir(),
```
zhouwfang: This is where backup_dir stops being somewhere we only write and becomes somewhere we delete from, and it still defaults to /tmp. Two nodes on one host would delete each other's archives, and anything local can drop a file there we can't unlink. I'd just make backup_dir required and drop the default.

dlukegordon: Fixed in 786e69f870ae. backup-dir is now a required PathBuf; the optional field, accessor, and /tmp fallback are gone. Updated the config examples and operator docs to require a persistent, node-specific directory with restricted write access. A regression test verifies configs missing backup-dir are rejected; workspace Clippy and all 43 backup-related tests pass.

crates/hashi/src/backup.rs
dlukegordon: Fixed in 786e69f870ae. Entry-read, file-type, and deletion failures now log and continue; directory-open failures still report an error. A temporary fault-injection smoke test confirmed that an undeletable expired archive does not prevent another expired archive from being removed. All 43 backup-related tests pass.

crates/hashi/src/lib.rs
dlukegordon: Fixed in 786e69f870ae: cleanup now runs before each automatic save attempt, and cleanup errors do not abort the save. The failed-save regression now verifies expired archives are deleted, and another test checks that save is attempted after a cleanup error. The docs explicitly note that all old archives can expire during a prolonged backup failure.

design/docs/node-backup.mdx
dlukegordon: Clarified in 786e69f870ae: manual backups do not trigger cleanup, but matching archives placed in backup-dir are subject to the same 14-day retention policy regardless of how they were created. Both backup docs and the operator runbook now describe this; longer-term copies should be stored elsewhere.

jessiemongeon1:

crates/hashi/src/lib.rs
```diff
@@ -349,12 +349,20 @@ impl Hashi {
             );
             return Ok(None);
         };
+        if let Err(error) =
+            crate::backup::cleanup_old_backups(&self.config.backup_dir, jiff::Timestamp::now())
```
zhouwfang: Sweeping before the save loses the floor the old code had, since it only ran after a save succeeded and so always left a fresh archive. Now failing saves empty the directory after a fortnight, and a node returning from a long outage deletes every archive before attempting a save that may itself fail. Erasure and recovery want different rules and they compose: delete past 14 days, but always keep the newest archive.

crates/hashi/src/config.rs
```diff
@@ -473,7 +465,10 @@ impl Config {
             bitcoin_trusted_peers: None,
             db: None,
             backup_pgp_cert: hashi_types::pgp::test_utils::mock_pgp_cert(),
-            backup_dir: None,
+            backup_dir: std::env::temp_dir().join(format!(
```
zhouwfang, line 468: Small one, but this puts a temp-dir path back for anything built from new_for_testing, and localnet writes it into the generated validator configs. A node started from one of those would back up somewhere that clears on reboot.

zhouwfang (changes requested):

crates/hashi/src/lib.rs
dlukegordon: Fixed and pushed: cleanup now preserves the newest regular backup archive regardless of age, while deleting other expired archives before saving. Regression coverage includes failed saves with only expired archives and excludes symlinks/directories from determining the newest archive. Updated the retention docs too. All 45 backup-related tests pass.

crates/hashi/src/config.rs
dlukegordon: Fixed and pushed: generated validator configs now explicitly set a node-specific backup directory under the selected localnet data directory. The network builder also places backups within each node’s storage directory. Verified the generated paths by launching the actual localnet CLI with two validators.

crates/hashi/src/backup.rs
```diff
@@ -364,11 +367,79 @@ pub fn encrypted_backup_file_name() -> PathBuf {
...
+                    "Failed to read backup file type: {error}"
+                );
+                continue;
+            }
+        };
+        if !file_type.is_file() {
+            continue;
+        }
+        let candidate = (created_at.timestamp(), path);
+        // Only delete an archive once another valid regular archive is known to be newer.
```
zhouwfang, line 522: "valid" isn't established. The archive is created at its final path before the encryptor is built and nothing removes it on failure, so a failed save leaves a 0-byte file with the newest timestamp and the floor pins it. Staging and renaming would fix it, and the module already does that for restore.

crates/e2e-tests/src/hashi_network.rs
```diff
@@ -584,7 +584,9 @@ impl HashiNetworkBuilder {
             config.bitcoin_chain_id = Some(hashi::constants::BITCOIN_REGTEST_CHAIN_ID.to_string());
             config.sui_chain_id = service_info.chain_id.clone();
             config.screener_endpoint = Some(screener_endpoint.clone());
-            config.db = Some(dir.join(validator_address.to_string()));
+            let node_dir = dir.join(validator_address.to_string());
+            config.backup_dir = node_dir.join("backups");
```
zhouwfang: This puts backup_dir inside the DB directory, which the doc in this same PR says not to do. The recovery test wipes that path, so it would take the backups with it.

crates/hashi/src/backup.rs
dlukegordon: Fixed and pushed in 69afe09bf10a. Encryption now writes to a temporary file in the destination directory and publishes it with persist_noclobber only after the tar archive and encryptor have finalized. Failed writes and publication collisions remove the staging file without creating a retention candidate or overwriting an existing archive. Added regressions for mid-write failure, recovery-floor preservation, and destination collisions; all 47 backup-related tests pass.

crates/e2e-tests/src/hashi_network.rs
dlukegordon: Fixed and pushed in 69afe09bf10a. Network backups now live at <hashi-root>/backups/<validator-address>, outside the unchanged DB path <hashi-root>/<validator-address>. The real backup/restore test now saves under the configured backup directory, deletes the DB, verifies the archive survives, and restores/rejoins the node. That end-to-end test passes. The backup guide also explicitly requires backup-dir to be outside the DB directory.

zhouwfang (commented):
The force-push dropped five tests while the code they cover is still there, no? Nothing exercises cleanup against a missing directory now.

crates/hashi/src/backup.rs
```diff
@@ -269,6 +285,11 @@ pub fn encrypt_files_to_pgp_archive(
         archive.finish()?;
     }
     encrypted.finalize()?;
+    staging
+        .persist_noclobber(output_path)
```
zhouwfang, line 307: Worth an fsync on the staging file before this. Nothing in the module syncs, so a crash just after the rename can leave a zero-length file under the final name, which then parses as the newest archive and becomes the recovery floor. That's the case the staging change was meant to remove.

crates/hashi/src/backup.rs
```diff
@@ -248,8 +251,21 @@ pub fn encrypt_files_to_pgp_archive(
...
 ) -> Result<()> {
-    let output = create_file_strict(output_path)?;
-    let mut encrypted = armored_encrypt_writer(output, recipient)?;
+    // Stage beside the destination so retention sees only fully finalized archives.
+    let parent = output_path
+        .parent()
+        .filter(|parent| !parent.as_os_str().is_empty())
+        .unwrap_or_else(|| Path::new("."));
+    let mut staging = tempfile::Builder::new()
+        .prefix(".hashi-backup-")
```
zhouwfang: The dot prefix means the sweep can't see these. The temp file only unlinks on drop, so an abrupt death leaves one behind and nothing ever removes it, in the directory this feature exists to bound. Worth sweeping .hashi-backup-* too.

crates/hashi/src/backup.rs
dlukegordon: Fixed and pushed in efb618893a41. The finalized staging file is now sync_all()’d before persist_noclobber, and the destination directory is synced after publication. A syscall trace of a successful backup confirmed staging-file fsync -> renameat2(RENAME_NOREPLACE) -> directory fsync, all successful. This verifies the ordering; it is not a simulated power-loss test.

crates/hashi/src/backup.rs
dlukegordon: Fixed and pushed in efb618893a41. Cleanup now removes regular .hashi-backup-* staging files older than 14 days by modification time, with no newest-file exemption. Recent staging files are preserved because another backup may still be writing them. Added deterministic coverage for expired/recent/exact-boundary staging, directories/symlinks, and staging files not displacing the newest recovery archive. All 45 backup-related tests pass.

dlukegordon:
Regarding the missing-directory coverage request (https://github.com/MystenLabs/hashi/pull/1137#pullrequestreview-5148156212): restored cleanup_old_backups_accepts_missing_directory in efb618893a41. It exercises cleanup against a nonexistent directory and verifies cleanup leaves it absent. I kept the other low-value/duplicate tests removed rather than restoring all five; the retained tests still cover the recovery floor, failed writes, no-clobber publication, and real restore/rejoin. Added targeted staging-expiry coverage for the new cleanup behavior. All 45 backup-related tests pass, workspace Clippy is clean, and make fmt completed.

crates/hashi/src/backup.rs
```diff
@@ -41,6 +41,10 @@ pub const BACKUP_FILE_NAME_PREFIX: &str = "hashi-backup";
 pub const BACKUP_MANIFEST_FILE_NAME: &str = "hashi-backup-manifest.toml";
 pub const DB_SNAPSHOT_TAR_PREFIX: &str = "hashi-db-snapshot";
 
+const BACKUP_FILE_NAME_FORMAT: &str = "hashi-backup-%Y%m%dT%H%M%SZ.tar.asc";
```
zhouwfang: This hardcodes the prefix instead of composing it from line 40. Rename that constant and the sweep silently matches nothing, with no compile error and no failing test, so archives just accumulate.

crates/hashi/src/backup.rs
```diff
@@ -364,11 +393,112 @@ pub fn encrypted_backup_file_name() -> PathBuf {
...
+                    directory = %output_dir.display(),
+                    "Failed to read backup directory entry: {error}"
+                );
+                continue;
+            }
+        };
+        let file_name = entry.file_name();
+        if file_name
+            .as_encoded_bytes()
+            .starts_with(BACKUP_STAGING_FILE_NAME_PREFIX.as_bytes())
```
zhouwfang: Worth pointing the same treatment at .hashi-restore-*. Restore stages under that prefix and only cleans up on drop, so a SIGKILL leaves the config, key files and DB dump in the clear, in the directory this feature is meant to bound.

crates/hashi/src/backup.rs
dlukegordon: Fixed and pushed in bc83c159c6d1. Generation now combines BACKUP_FILE_NAME_PREFIX with a shared timestamp/suffix format; cleanup strips that same prefix and parses the shared suffix without allocating a format string per entry. Added a regression using the actual generated filename: an aged sweep removes an older archive while retaining the generated recovery archive. All 48 backup-related tests pass.

crates/hashi/src/backup.rs
dlukegordon: Fixed and pushed in bc83c159c6d1. CLI restore and cleanup now share the .hashi-restore- prefix. Within backup-dir, cleanup removes matching directories older than 14 days by directory mtime, including their staged contents, while preserving recent/exact-cutoff entries, symlinks, and successful extraction directories. Tests cover nested key/config/DB-like files and preservation of both top-level and nested symlink targets. The docs make the scope explicit: restore output elsewhere is not swept. All 48 backup-related tests pass.

crates/hashi/src/backup.rs
zhouwfang: Looks like bc83c15 landed and then got rolled back, no? Both fixes seem off the head again.

crates/hashi/src/backup.rs
```diff
@@ -269,6 +287,18 @@ pub fn encrypt_files_to_pgp_archive(
...
+    staging
+        .as_file()
+        .sync_all()
+        .context("Failed to sync backup staging file")?;
+    staging
+        .persist_noclobber(output_path)
+        // Drop the temporary file even if the caller retains the error.
+        .map_err(|e| e.error)
+        .with_context(|| format!("Failed to publish backup {}", output_path.display()))?;
+    File::open(parent)
```
zhouwfang: This runs after the archive is already in place, so a failure here reports a backup that actually succeeded. Directory fsyn isn't supported everywhere either. Probably wants to be a warn.

zhouwfang (commented):
Scope question: .hashi-db-restore- holds the decrypted database and nothing expires it, and completed restore output is plaintext too. I don't think either belongs in this sweep, since they're restore-side and short-lived, but it does mean we expire the ciphertext on a schedule and leave the plaintext to the operator. Probably worth restore cleaning up its own staging on crash rather than leaving it to this sweep. WDYT?

crates/hashi/src/backup.rs
dlukegordon: Fixed and pushed in d6f27280be77. Failure to open or sync the destination directory now logs a warning and returns backup success, since publication has already completed. Staging-file sync remains mandatory before publication. Verified with a temporary fault-injection shim returning EINVAL for the post-publication directory fsync: backup succeeded and GPG decrypted the resulting archive. All 48 backup-related tests and workspace Clippy pass.

crates/hashi/src/backup.rs
dlukegordon: Yes, the PR head had moved back. Both fixes were restored in 7a89d573 and are retained in the current pushed head d6f27280be77: filename generation/parsing share BACKUP_FILE_NAME_PREFIX and the suffix format, and cleanup handles expired .hashi-restore-* directories using the same prefix as the restore producer. The generated-name and restore-staging regressions are present and pass.

dlukegordon:
On restore cleanup ownership (https://github.com/MystenLabs/hashi/pull/1137#pullrequestreview-5158228986): agreed that completed restore output and .hashi-db-restore-* should not be added to the automatic backup sweep. Completed output is intentional user data; DB staging lives beside the restore destination, outside the backup directory contract. Restore-owned orphan cleanup would have to run on a later invocation/recovery path with protection for active restores—SIGKILL cannot execute cleanup handlers. I have not added that separate lifecycle behavior to this retention PR. In d6f27280be77 the docs explicitly call out which plaintext copies remain operator-managed. The already-requested expiry of abandoned .hashi-restore-* directories within backup-dir remains scoped to that directory.

crates/hashi/src/lib.rs
```diff
@@ -349,12 +349,20 @@ impl Hashi {
             );
             return Ok(None);
         };
+        if let Err(error) =
```
zhouwfang: The sweep lives inside backup_after_epoch_change, which only runs when the node has a role, so a node with no role stops writing archives and stops expiring them at the same time. That's the departed-validator case, which is the one that matters most for erasure. Hoisting the sweep above that gate would cover it.

crates/hashi/src/backup.rs
```diff
@@ -269,6 +287,21 @@ pub fn encrypt_files_to_pgp_archive(
...
+        .as_file()
+        .sync_all()
+        .context("Failed to sync backup staging file")?;
+    staging
+        .persist_noclobber(output_path)
+        // Drop the temporary file even if the caller retains the error.
+        .map_err(|e| e.error)
+        .with_context(|| format!("Failed to publish backup {}", output_path.display()))?;
+    if let Err(error) = File::open(parent).and_then(|directory| directory.sync_all()) {
+        warn!(
```
zhouwfang: Maybe error! rather than warn! here. A persistent failure repeats every epoch next to an INFO line saying the backup completed, and there's no metric, so a directory that never syncs looks healthy.

dlukegordon: Changed to error! in 72b1ed8c3731. A post-publication directory-sync failure now logs at ERROR level, while still returning success for the already-published archive. Staging-file sync remains mandatory before publication. Updated the docs; all 49 backup-related tests and workspace Clippy pass.

crates/hashi/src/lib.rs
dlukegordon: Fixed and pushed in 72b1ed8c3731. Epoch housekeeping now always queues maintenance on the existing serialized backup worker, carrying a separate write-backup flag. The worker sweeps before either the write-policy or config-path gate, so NoRole nodes still expire archives without creating a new backup. Extended the real resigned-member e2e test across the second boundary where the departed node has no role: the older archive is deleted, the newest remains, and no archive is added. That test and all 49 backup-related tests pass.

crates/hashi/src/backup.rs
```diff
@@ -57,10 +62,13 @@ pub struct BackupHandle {
 }
 
 impl BackupHandle {
-    pub fn backup_after_epoch_change(&self, epoch: u64) {
+    pub fn backup_after_epoch_change(&self, epoch: u64, write_backup: bool) {
```
zhouwfang: Now that a request can be maintenance-only, this name and the logs around it still say backup. The warn at :76 fires on a role-less node where the only work was the sweep, and :56 and :120 read the same way.

crates/hashi/src/backup.rs
```diff
@@ -97,10 +105,13 @@ impl BackupService {
...
             match request {
-                BackupRequest::EpochChanged(epoch) => {
+                BackupRequest::EpochChanged {
+                    epoch,
+                    write_backup,
+                } => {
                     let hashi = self.inner.clone();
                     match tokio::task::spawn_blocking(move || {
-                        hashi.backup_after_epoch_change(epoch)
+                        hashi.backup_after_epoch_change(epoch, write_backup)
```
zhouwfang: When the sweep runs on its own it leaves no trace. The success arm just below at :118 is empty and the early return at lib.rs:357 is silent, which is why the e2e polls the filesystem instead of checking a log.

dlukegordon: Fixed in baee1d67d9e6. A completed retention sweep now emits an INFO event with epoch, write_backup, and directory before either early-return gate. Failed sweeps retain the warning instead of getting a completion event. The real resigned-validator e2e emitted “Epoch backup retention sweep completed epoch=2 write_backup=false directory=…”. I kept its filesystem assertions because they verify deletion/recovery behavior independently of logging; no wording-pinning test was added.

crates/hashi/src/backup.rs
dlukegordon: Fixed in baee1d67d9e6. Renamed both methods to maintain_backups_after_epoch_change and the request type to BackupMaintenanceRequest. Stopped-worker and failure diagnostics now describe epoch backup maintenance and include epoch/write_backup fields, so cleanup-only requests are no longer reported as attempted archive writes. All callers were migrated; 49 backup-related tests and the resigned-validator e2e pass.

crates/hashi/src/lib.rs
```diff
@@ -341,7 +341,27 @@ impl Hashi {
...
+        &self,
+        epoch: u64,
+        write_backup: bool,
+    ) -> anyhow::Result<Option<PathBuf>> {
+        match crate::backup::cleanup_old_backups(&self.config.backup_dir, jiff::Timestamp::now()) {
+            Ok(()) => tracing::info!(
+                epoch,
+                write_backup,
+                directory = %self.config.backup_dir.display(),
+                "Epoch backup retention sweep completed",
```
zhouwfang, line 356: This fires on Ok, which also covers the cases where nothing happened: the directory doesn't exist, or every delete failed and the sweep carried on. A removed/failed count would make the line mean something.

zhouwfang (approved):

PR #1138, opened 2026-09-08
reconfig: replace the abort_reconfig proposal with a permissionless e…

bmwill:
…ntry

Previously, aborting a stuck reconfiguration required an `AbortReconfig` governance proposal voted on by the current committee. Audit feedback pointed out that the committees able to vote are exactly the parties a stalled reconfiguration puts in doubt: the pending committee may never finish DKG or key rotation, and a proposal gated on the outgoing committee's quorum can be stranded by the same offline stake that stalled the reconfiguration.

With this commit the voting mechanism is gone. `reconfig::abort_reconfig` is a private entry function that anyone may call, and it succeeds only when both objective on-chain conditions hold:

1. A reconfiguration is in progress (`ENotReconfiguring` otherwise).
2. The pending epoch is not Sui's current epoch (`committee_set::EPendingEpochStillCurrent` otherwise).

`start_reconfig` pins the pending committee's epoch to Sui's epoch at formation time, so while the two still match the reconfiguration is inside its window and may complete; once Sui's epoch has moved past it the reconfiguration is presumed stuck. The epoch gate lives in `committee_set::abort_reconfig`, mirroring how `start_reconfig` owns its own epoch checks. A successful abort emits a new `ReconfigAborted` event.

The `abort_reconfig` proposal module is deleted. This tree publishes a fresh package for the pre-mainnet cycle (`Published.toml` records no deployment), so removing a module is fine here; it would not pass the compatible-upgrade policy on the deployed `testnet` maintenance branch.

On the Rust side, `ProposalType::AbortReconfig`, its payload type, the `proposal create abort-reconfig` command, and every match arm in the CLI client, proposal decoding, metrics labels, and leader garbage collection are removed. A new `hashi committee abort-reconfig` command reads Sui's epoch from the fullnode, refuses up front when nothing is pending or the pending epoch is still current, and submits the entry through the latest package. `ReconfigAborted` is mirrored in `hashi-types` for event parsing.

Tests: the proposal-based abort tests are replaced by entry-function tests in `reconfig_tests.move` covering the successful teardown and event payload, an outsider sender, the still-current-epoch rejection, the not-reconfiguring rejection, discarding an already-submitted handoff certificate, and a second abort failing. The CLI pre-flight decision is a pure function with unit tests. Docs describe the new flow and command.


packages/hashi/sources/core/reconfig.move
```diff
@@ -138,6 +145,28 @@ entry fun submit_committee_handoff(
...
+///
+/// Deliberately not a governance proposal. The committees that could vote on
+/// one are exactly the parties a stalled reconfiguration puts in doubt: the
+/// pending committee may never finish DKG or key rotation, and a proposal
+/// gated on the outgoing committee's quorum can be stranded by the same
+/// offline stake that stalled the reconfiguration. Binding the abort to an
+/// objective on-chain fact instead keeps the escape hatch usable precisely
+/// when it is needed.
+///
+/// Races with a late `end_reconfig` are benign: whichever transaction lands
```
zhouwfang: The race isn't benign on the node side. The note at :20 says ENotReconfiguring is matched by name to mean "another node already finished end_reconfig", and this adds a second cause with the opposite meaning. The loser reports success, bypasses the guard at service.rs:267, and prunes the live epoch's nonce rows. Can the abort get its own error constant?

jessiemongeon1:

packages/hashi/sources/core/reconfig.move
```diff
@@ -138,6 +145,28 @@ entry fun submit_committee_handoff(
...
+/// first wins, and the other aborts with `ENotReconfiguring`.
+entry fun abort_reconfig(self: &mut Hashi, ctx: &TxContext) {
```
zhouwfang: Could this take the target epoch and assert it matches? The old module bound the target and the test that pinned it went away with it. Also, start_reconfig calls assert_genesis_launch_authorized and this doesn't.

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -429,8 +432,24 @@ public(package) fun end_reconfig(
     (next_epoch, committee_handoff_cert)
 }
 
-public(package) fun abort_reconfig(self: &mut CommitteeSet, _ctx: &TxContext): u64 {
+/// Tear down an in-flight reconfiguration that has overrun its Sui epoch:
+/// clears the pending epoch change and removes the pending committee (and
+/// any handoff certificate it collected). The current epoch, committee, and
+/// MPC public key are untouched, so operations resume under the last
```
zhouwfang: "operations resume under the last committed committee" isn't true at genesis, which is the case this was written for.

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -429,8 +432,24 @@ public(package) fun end_reconfig(
...
+/// formation time, so a pending epoch that still equals Sui's epoch means the
+/// reconfiguration is inside its window and may legitimately complete; it
+/// must not be torn down under a committee that is mid-protocol. Only once
+/// Sui's epoch has moved past the target is the reconfiguration presumed
+/// stuck and abortable. This is the whole gate: there is no signer check,
+/// because a stalled reconfiguration is exactly the state in which no
+/// committee can be relied on to produce a certificate or a quorum.
+public(package) fun abort_reconfig(self: &mut CommitteeSet, ctx: &TxContext): u64 {
     assert!(self.is_reconfiguring());
+    assert!(self.pending_epoch_change.borrow().epoch != ctx.epoch(), EPendingEpochStillCurrent);
```
zhouwfang, line 500: end_reconfig still reads no epoch, so an overrunning reconfiguration is alive rather than stuck and anyone can kill one that's about to land. And in the other direction, a pending committee that's visibly broken five minutes in now can't be cleared by anyone for up to a day, with deposits and withdrawals halted throughout. I think the end_reconfig epoch check needs to land with this.

crates/hashi/src/cli/commands/committee.rs
```diff
@@ -170,3 +174,57 @@ pub async fn show_epoch(config: &CliConfig) -> Result<()> {
...
+        refuse_unabortable_reconfig(client.onchain_state().pending_epoch_change(), sui_epoch)?;
+
+    print_detail(&format!(
+        "\n{}",
+        "Aborting the pending reconfiguration:".bold()
+    ));
+    print_detail(&format!("  Pending Hashi epoch: {pending_epoch}"));
+    print_detail(&format!("  Current Sui epoch:   {sui_epoch}"));
+    print_detail(&format!(
+        "  Effect: the pending committee is discarded. Hashi stays at epoch {} under its current \
```
zhouwfang: The effect text omits that the retry waits for the next Sui epoch, since try_submit_start_reconfig only fires on SuiEpochChanged and already returned early this epoch.

crates/hashi-types/src/move_types/mod.rs
```diff
@@ -1844,6 +1840,24 @@ impl From<ReconfigEnded> for HashiEvent {
     }
 }
 
+/// Emitted by `reconfig::abort_reconfig`: the pending epoch was torn down and
+/// the current epoch is unchanged.
+#[derive(Debug, serde_derive::Deserialize)]
+pub struct ReconfigAborted {
```
zhouwfang, line 1831: This is decoded but nothing consumes it, and there's no abort metric either. Now that a third party can trigger a teardown, that's the event worth paging on?

packages/hashi/tests/reconfig_tests.move
```diff
@@ -215,3 +216,126 @@ fun test_submit_committee_handoff_rejects_handoff_signed_by_wrong_committee() {
...
 }
+
+// ======== abort_reconfig ========
+
+/// Hashi at epoch 0 with a pending epoch-1 committee and a set MPC key, i.e.
+/// a non-initial reconfiguration in flight.
+fun hashi_with_pending_reconfig(ctx: &mut TxContext): hashi::hashi::Hashi {
+    let mut hashi = test_utils::create_hashi_with_committee(vector[VOTER1, VOTER2, VOTER3], ctx);
+    hashi.committee_set_mut().set_pending_reconfig_for_testing(pending_committee_for_testing(1));
+    hashi.committee_set_mut().set_mpc_public_key_for_testing(vector[1, 2, 3]);
```
zhouwfang, line 246: The shared helper sets an MPC key, so every new test is steady state and genesis is untested. Nothing asserts that a fresh start_reconfig succeeds after an abort either, which is what the design rests on.

packages/hashi/sources/core/committee/committee_set.move
bmwill: yes and we can revisit addressing that, for now we're opting to eliminate some of the other ambiguity that exists  with the voting to close the epoch

crates/hashi/src/sui_tx_executor.rs
```diff
@@ -1203,6 +1203,40 @@ impl SuiTxExecutor {
         Ok(())
     }
 
+    /// Tear down a reconfiguration that has overrun its Sui epoch
+    /// (`reconfig::abort_reconfig`). Permissionless on chain; `epoch` names
+    /// the pending target so a stale submission cannot abort a newer one.
+    /// A failed status surfaces as [`TransactionExecutionError`] so the
```
zhouwfang: This breaks the docs job. TransactionExecutionError is pub(crate) at :369, so the link from a public fn trips private_intra_doc_links under -Dwarnings. Dropping the brackets fixes it.

crates/hashi/src/mpc/service.rs
```diff
@@ -1644,6 +1769,15 @@ impl MpcService {
                         );
                         self.sleep_if_still_pending(target_epoch).await;
                     }
+                    ReconfigSubmissionErrorKind::ReconfigTargetDead => {
```
zhouwfang, line 1712: This arm returns without sleeping, while NonMoveAbort right above it sleeps. When our checkpoint clock lags the chain, we re-run the whole protocol and submit a doomed transaction on every loop iteration.

packages/hashi/sources/core/reconfig.move
```diff
@@ -120,12 +148,16 @@ entry fun end_reconfig(
...
-    _ctx: &TxContext,
+    ctx: &TxContext,
 ) {
     self.versioning().assert_version_enabled();
-    assert!(self.committee_set().is_reconfiguring(), ENotReconfiguring);
+    // The certificate is signed by the outgoing committee, so its epoch is
+    // the handoff's source epoch, and a stored handoff for that epoch means
+    // the transition it approved already activated.
+    let from_epoch = committee_handoff_cert.signature_epoch();
+    let already_completed = self.committee_set().has_committee_handoff(from_epoch);
```
zhouwfang: has_committee_handoff(from_epoch) says some transition out of that epoch activated, not that this submission's target did. The comment above and the doc at :221 both claim the stronger property.

crates/hashi/src/mpc/service.rs
```diff
@@ -184,6 +191,11 @@ impl MpcService {
         } else {
             info!("Node is not in the current committee, waiting for reconfig notification...");
         }
+        // Hashi lagging Sui with nothing pending is what a manual abort, or a
+        // boundary missed while this node was down, leaves behind; catch up
+        // now rather than wait for the next boundary.
+        self.try_submit_start_reconfig(self.inner.onchain_state().latest_checkpoint_epoch())
```
zhouwfang, line 196: The deadline is epoch-aligned but starts aren't, and there's no minimum-remaining check. A node coming up near a boundary forms a committee that can't finish and gets aborted.

packages/hashi/sources/core/reconfig.move
```diff
@@ -138,6 +170,35 @@ entry fun submit_committee_handoff(
...
+/// The launch switch applies exactly as it does to `start_reconfig`, so the
+/// pre-launch state machine is the same at both ends of a reconfiguration.
+entry fun abort_reconfig(self: &mut Hashi, epoch: u64, ctx: &TxContext) {
+    self.versioning().assert_version_enabled();
+    assert!(self.committee_set().is_reconfiguring(), ENotReconfiguring);
+    assert!(
+        self.committee_set().pending_epoch_change().destroy_some() == epoch,
+        EWrongReconfigEpoch,
+    );
+    assert_genesis_launch_authorized(self);
```
zhouwfang: This assert can't fire, since start_reconfig already passed the same check at :90 and versioning.move never removes the cap.

crates/hashi/src/mpc/service.rs
```diff
@@ -2403,16 +2628,164 @@ mod pruning_reference_tests {
...
+        assert!(reconfig_target_live(None, 6, true, 6));
+        assert!(!reconfig_target_live(None, 5, true, 6));
+        assert!(!reconfig_target_live(Some(7), 5, true, 6));
+        assert!(!reconfig_target_live(None, 7, true, 6));
+    }
+
+    #[test]
+    fn an_aborted_genesis_at_sui_epoch_zero_is_not_live() {
+        // Pre-genesis: epoch 0, no committee, and the genesis target is 0.
+        assert!(reconfig_target_live(Some(0), 0, false, 0));
```
zhouwfang: The comment says no committee, but with pending Some(0) the epoch-0 committee is already inserted, so this passes on the first disjunct rather than the state it describes.

packages/hashi/tests/reconfig_tests.move
```diff
@@ -211,7 +215,439 @@ fun test_submit_committee_handoff_rejects_handoff_signed_by_wrong_committee() {
...
+        3,
+    );
+    (mpc_cert, handoff_cert)
+}
+
+#[test]
+#[expected_failure(abort_code = reconfig::EReconfigWindowClosed)]
+/// Once Sui's epoch has moved past the pending epoch a valid completion is
+/// refused: from then on the reconfiguration can only be aborted.
+fun test_end_reconfig_rejects_closed_window() {
```
zhouwfang, line 508: The e2e network's epoch is 24 hours by default and no test shortens it, so the Sui epoch never advances during a run. That means nothing exercises the window closing or the abort-and-restart cycle end to end

packages/hashi/sources/core/reconfig.move
```diff
@@ -153,6 +219,50 @@ public(package) fun assert_genesis_launch_authorized(self: &Hashi) {
...
+
+/// Whether `cert` certifies the handoff already stored out of its signing
+/// epoch. Handoffs are keyed by source epoch, and after an abort a
+/// replacement can form from the same source epoch, so the source alone does
+/// not identify a transition. The target is bound only inside the signed
+/// message (the incoming committee, epoch included), so the certificate is
+/// verified against the stored transition to read it out: a certificate for
+/// a target that was aborted and then replaced from the same source epoch
+/// fails verification here instead of passing as a completion of itself.
+fun certifies_stored_handoff(self: &Hashi, cert: CommitteeSignature): bool {
```
zhouwfang: This returns bool but can abort. Once a handoff is stored, verify_with_committee at :257 raises committee::ESigVerification, which has no classifier arm, so the node panics at service.rs:1796. On the parent the same state returned EReconfigAlreadyCompleted, so the false success became a crash. Could this raise a named reconfig abort instead?

crates/hashi/src/sui_tx_executor.rs
zhouwfang: The handoff goes into the PTB before end_reconfig at :84, so it aborts first and the new EWrongReconfigEpoch only runs at genesis or after the retry drops the cert.

packages/hashi/sources/core/reconfig.move
```diff
@@ -86,9 +109,17 @@ entry fun end_reconfig(
...
+    // The certificate is signed by the incoming committee, so its epoch is
+    // this submission's target. An activated target is the current epoch and
+    // still has its committee; an aborted one has neither.
+    let next_epoch = mpc_cert.signature_epoch();
+    let already_completed =
+        self.committee_set().epoch() == next_epoch &&
+        self.committee_set().has_committee(next_epoch);
+    let pending_epoch = pending_epoch_in_window(self, already_completed, ctx);
+    // A different pending epoch means this target was aborted and replaced.
+    assert!(pending_epoch == next_epoch, EWrongReconfigEpoch);
```
zhouwfang, line 129: This also fires on a target that completed, since already_completed is computed at :116 and discarded whenever anything is pending. The node then skips prepare_signing for an epoch it's actually serving.

crates/hashi/src/mpc/service.rs
```diff
@@ -1644,6 +1771,21 @@ impl MpcService {
...
+                        _reconfig_timer.stop_and_discard();
+                        // The chain knows the target is dead before the
+                        // mirror does. Until the mirror catches up the outer
+                        // loop still sees it pending and would re-enter here
+                        // immediately, re-collecting signatures and
+                        // submitting another doomed transaction each pass.
+                        self.sleep_if_still_pending(target_epoch).await;
+                        return;
+                    }
                     ReconfigSubmissionErrorKind::NonRetryableMoveAbort
```
zhouwfang, line 1727: A whitelist with a fatal default means any unrecognised abort in the PTB's call graph kills the node, and unnamed assert!s can never be classified at all. Giving up loudly on unknown would have contained this.

packages/hashi/tests/reconfig_tests.move
```diff
@@ -211,7 +215,503 @@ fun test_submit_committee_handoff_rejects_handoff_signed_by_wrong_committee() {
...
+    reconfig::abort_reconfig_for_testing(hashi, 1, late);
+    hashi.committee_set_mut().set_pending_reconfig_for_testing(pending_committee_for_testing(2));
+}
+
+#[test]
+#[expected_failure(abort_code = hashi::committee::ESigVerification)]
+/// A handoff certificate binds its target inside the signed message (the
+/// incoming committee, epoch included), so one for an aborted target cannot
+/// verify against the replacement now pending from the same source epoch.
+fun test_submit_committee_handoff_rejects_aborted_target_while_replacement_pending() {
```
zhouwfang, line 716: This passes against the parent too. The replacement is pending, so the new helper never runs and the ESigVerification is the pre-existing one.

crates/hashi/src/mpc/service.rs
```diff
@@ -72,10 +72,20 @@ const MAX_KEY_REREGISTRATION_BUMPS: u32 = 3;
...
-/// numeric code is not stable).
+/// Move `hashi::reconfig` abort constants, matched by their clever-error
+/// constant names (the `#[error]` abort code encodes a source line, so the
+/// numeric code is not stable). Together they tell the benign "another node
+/// already completed it" race from a dead target: aborted, aborted and
+/// replaced by a different pending epoch, or past its Sui epoch window.
 const RECONFIG_E_NOT_RECONFIGURING: &str = "ENotReconfiguring";
+const RECONFIG_E_ALREADY_COMPLETED: &str = "EReconfigAlreadyCompleted";
+const RECONFIG_E_WINDOW_CLOSED: &str = "EReconfigWindowClosed";
+/// Raised by the completion entries when a different epoch is pending (the
```
zhouwfang: "completion entries" is plural, but only end_reconfig raises it. The handoff raises a committee code instead.

crates/hashi/src/mpc/service.rs
```diff
@@ -2403,16 +2639,171 @@ mod pruning_reference_tests {
...
+    }
+
+    #[test]
+    fn an_aborted_or_overrun_target_is_dead_not_completed() {
+        for constant in [
+            "ENotReconfiguring",
+            "EReconfigWindowClosed",
+            "EWrongReconfigEpoch",
+        ] {
+            let error = move_abort("reconfig", "pending_epoch_in_window", Some(constant));
```
zhouwfang: The fixture builds the abort in pending_epoch_in_window, but that constant is raised at reconfig.move:121 and :201.

crates/hashi/src/mpc/service.rs
```diff
@@ -202,19 +216,28 @@ impl MpcService {
...
                     match notification {
                         Ok(notification) => match notification {
                             Notification::StartReconfig(epoch) => {
-                                self.handle_reconfig(epoch).await;
+                                self.drive_reconfig(epoch).await;
                             }
                             Notification::SuiEpochChanged(sui_epoch) => {
                                 self.try_submit_start_reconfig(sui_epoch).await;
                             }
+                            Notification::ReconfigAborted(epoch) => {
```
dlukegordon, line 229: After an aborted genesis DKG this node is back in `is_awaiting_genesis()` but now sits in the main loop, so the replacement `start_reconfig` only gets `try_submit_start_reconfig`'s 3 attempts (~30s). If those fail on every node (fullnode hiccup right after the abort), nothing retries until the next Sui epoch boundary or a restart, whereas the startup path (`try_submit_genesis_reconfig`) retries until a pending epoch appears. Launch-only edge case, but the fix is small since all three entry points funnel through `try_submit_start_reconfig`:

```rust
// try_submit_start_reconfig, right after `if hashi_epoch >= sui_epoch { return; }`
if self.is_awaiting_genesis() {
    // Pre-genesis: same retry-until-pending loop as startup.
    self.try_submit_genesis_reconfig().await;
    return;
}
```

packages/hashi/sources/core/committee/committee_set.move
```diff
@@ -429,8 +418,63 @@ public(package) fun end_reconfig(
...
+    self.pending_epoch_change =
+        option::some(PendingEpochChange {
+            epoch,
+            committee_handoff_cert: option::none(),
+        });
+    self.insert_committee(committee);
+    epoch
+}
+
+fun abort_reconfig_inner(self: &mut CommitteeSet): u64 {
```
dlukegordon: `abort_reconfig_inner` is only called from `abort_reconfig` right above. Unlike `start_reconfig_from_voting_powers` there's no `SuiSystemState` reason for the split; inlining it keeps the two asserts and the teardown in one place.

dlukegordon (approved):
Both of my comments are addressed on b1618470 (genesis restart now uses the retry-until-pending loop; abort_reconfig teardown inlined).

crates/hashi/src/mpc/service.rs
```diff
@@ -181,6 +190,11 @@ impl MpcService {
         } else {
             info!("Node is not in the current committee, waiting for reconfig notification...");
         }
+        // Hashi lagging Sui with nothing pending is what a manual abort, or a
+        // boundary missed while this node was down, leaves behind; catch up
+        // now rather than wait for the next boundary.
+        self.try_submit_start_reconfig(self.inner.onchain_state().latest_checkpoint_epoch())
```
zhouwfang: A wedged committee node never gets here, since the loop above only breaks once a SigningManager exists. One wedged node is fine because a healthy peer submits for it, but on a fleet-wide restart nobody does, so the epoch never advances and the presig cursor never resets.

crates/hashi/src/mpc/service.rs
```diff
@@ -1573,6 +1709,21 @@ impl MpcService {
                         );
                         self.sleep_if_still_pending(target_epoch).await;
                     }
+                    ReconfigSubmissionErrorKind::ReconfigTargetDead => {
```
zhouwfang, line 1712: Once handle_reconfig sets up for the target at :1604, every give-up exit after that leaves the MpcManager there, which after an abort is dead, and nothing puts it back since sync_if_stale returns early at :1048. Refill then fails ensure_manager_epoch at :668, so no node that took part can replenish its pool until the next successful end_reconfig.

crates/hashi/src/mpc/service.rs
zhouwfang, line 196: A wedged committee node never gets here, since the loop above only breaks once a SigningManager exists. One such node is covered by a peer, but on a fleet-wide restart with every running node in the committee nobody submits, and there's no start-reconfig command either. What wedges them is the pinned manager above, since a pool that can't refill is what stops them rebuilding.

zhouwfang (commented):
Three must-fix items. The first two weren't must-fix before this PR, because in practice a pending target only ever stopped being pending by completing, and this PR makes a dead target the normal outcome of an overrun. The third is new here, since the replacement is now submitted in the same breath as the abort.

crates/hashi/src/mpc/service.rs
```diff
@@ -1381,8 +1482,17 @@ impl MpcService {
...
-        if self.get_pending_epoch_change().is_some() {
-            return;
+        if let Some(pending) = self.get_pending_epoch_change() {
+            if !reconfig_window_closed(sui_epoch, pending) {
+                return;
+            }
+            // The pending reconfiguration overran its window, so the chain
+            // will not let it complete; abort it (any node may) so the
+            // replacement can form in the new epoch.
+            self.abort_stale_reconfig(pending).await;
```
zhouwfang, line 1492: The abort and the replacement sit in the same function, and the only gate before submitting is hashi_epoch >= sui_epoch, which is false by construction after an abort. The ReconfigAborted arm does it too, so an operator's abort-reconfig is undone within a checkpoint or two. There's then no way to run on the last committed committee while debugging, which is what committee_set.move:484 still promises.

PR #1150, opened 2026-09-09
Keep private keys and passwords out of Debug output

zhouwfang:
#918


dlukegordon (approved):
Reviewed all 15 changed files at c8193e85. No actionable findings. Checked key-wrapper serialization and dependency behavior, caller migration, configuration redaction, and share-index validation. Static review only; tests and builds were not run.

PR #1151, opened 2026-09-10
Require proof of possession and uniqueness for TLS key registration

zhouwfang:
#918


dlukegordon (approved):
Reviewed all 12 changed files and the affected TLS authentication and registry-refresh paths. No actionable findings. Source-only review; tests and builds were not run.

PR #1154, opened 2026-09-10
withdraw: drop request_ids from the finalize_withdrawal cert

bmwill:
Previously, `finalize_withdrawal` took a `request_ids` vector and bound it into the committee-signed `WithdrawalSignedMessage`. The contract never read it: the request ids already live on the `WithdrawalTransaction` keyed by `withdrawal_id`, and the finalize path only attaches the guardian signatures and flips the broadcast gate. Followers compared the leader-supplied ids against the on-chain txn, which only guarded a value nobody used.

With this commit, the entry function, the signed message, and its Rust mirror `WithdrawalTxSigning` carry only `withdrawal_id`, `signatures`, and `guardian_signatures`. The leader no longer clones the ids into the cert request or the Sui call, and the follower-side "request IDs mismatch" check is gone with the field. The gRPC `SignWithdrawalTxSigningRequest` reserves tag 2 and the `request_ids` name, and the generated bindings are regenerated.

This changes the BCS preimage signed under the `WITHDRAWAL_SIGNED` intent, so the Move package and node binary must ship together.


dlukegordon (approved):
Reviewed the full diff and affected withdrawal finalization paths at 439b1b46a72a3724060e1924f7adb1d1f36f7bbb. No actionable findings. Rust and Move signing payloads, RPC conversions, transaction arguments, and generated protobuf descriptor agree. Verification was source review and descriptor inspection; tests and builds were not run.

PR #1156, opened 2026-09-10
[Monitor] Respect Sui gRPC completion watermarks

mskd12:
## Summary

- advance the Sui monitor cursor only through `Watermark.checkpoint`, the inclusive fully covered checkpoint
- retain and validate watermark/cursor progress across paginated `ListTransactions` scans
- defer transactions beyond the completion watermark so indexing lag cannot skip an event
- use `GetServiceInfo` for the chain-head snapshot and retry transient range failures with exponential backoff
- add regression coverage for `LedgerTip` and incomplete checkpoint scans

## Incident

The monitor treated `LEDGER_TIP` as proof that its requested checkpoint range was complete. `GetCheckpoint(latest)` could be ahead of the transaction history index, allowing the monitor to advance past a real `WithdrawalPickedForProcessing` event before it became queryable.

## Verification

- `cargo fmt --all -- --check`
- `cargo check`
- `git diff --check`

`make fmt` completed Rust and Buf formatting but could not run `prettier-move` because it is not installed locally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,10 +4,10 @@ members = ["crates/*"]
 
 [workspace.dependencies]
 # sui sdk crates
-sui-sdk-types = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "d71f9d6f788afc6b30baef85a342e0e7867c0019", features = ["serde", "hash"] }
-sui-crypto = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "d71f9d6f788afc6b30baef85a342e0e7867c0019", features = ["ed25519", "secp256k1", "secp256r1", "pem", "bech32"] }
-sui-rpc = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "d71f9d6f788afc6b30baef85a342e0e7867c0019" }
-sui-transaction-builder = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "d71f9d6f788afc6b30baef85a342e0e7867c0019" }
+sui-sdk-types = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "f9b4465d8e95358b4a0bef225e611797de9b30e7", features = ["serde", "hash"] }
+sui-crypto = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "f9b4465d8e95358b4a0bef225e611797de9b30e7", features = ["ed25519", "secp256k1", "secp256r1", "pem", "bech32"] }
+sui-rpc = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "f9b4465d8e95358b4a0bef225e611797de9b30e7" }
+sui-transaction-builder = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "f9b4465d8e95358b4a0bef225e611797de9b30e7" }
 
 sui-http = "0.3.1"
 bitcoin = { version = "0.32.8", features =["serde"] }
--- a/crates/hashi-monitor/src/rpc/sui.rs
+++ b/crates/hashi-monitor/src/rpc/sui.rs
@@ -18,6 +18,7 @@ use sui_rpc::proto::sui::rpc::v2::Checkpoint;
 use sui_rpc::proto::sui::rpc::v2::Event;
 use sui_rpc::proto::sui::rpc::v2::ExecutedTransaction;
 use sui_rpc::proto::sui::rpc::v2::GetCheckpointRequest;
+use sui_rpc::proto::sui::rpc::v2::GetServiceInfoRequest;
 use sui_rpc::proto::sui::rpc::v2::ListTransactionsRequest;
 use sui_rpc::proto::sui::rpc::v2::Ordering;
 use sui_rpc::proto::sui::rpc::v2::QueryEndReason;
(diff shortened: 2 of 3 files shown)


0xsiddharthks (commented):
The watermark approach looks right to me, but a few things need fixing first. The two sui.rs ones (events dropped when the timestamp lookup fails, and poll hanging on transient errors) are the important ones, and the GuardianInfo reorder in the last commit looks like it breaks signature checks for guardians built after #1121.

crates/hashi-monitor/src/rpc/sui.rs
```diff
@@ -113,67 +159,89 @@ impl SuiEventsPoller {
...
+            // transactions for the next poll so the checkpoint is ingested
+            // atomically and cannot be skipped as indexing catches up.
+            if checkpoint > completed_checkpoint {
+                continue;
+            }
             events.extend(self.parse_transaction(transaction)?);
         }
 
-        self.next_checkpoint_to_scan = Some(end_checkpoint);
+        self.next_checkpoint_to_scan = Some(completed_checkpoint.saturating_add(1));
```
0xsiddharthks: If `checkpoint_timestamp` on the next line fails, poll returns Err after `next_checkpoint_to_scan` has already moved, and the parsed `events` are dropped. The next tick starts at `completed_checkpoint + 1`, so `[start, completed_checkpoint]` never gets ingested, which is the same missed `WithdrawalPickedForProcessing` this PR is fixing. On the LedgerTip and split paths that checkpoint usually isn't cached, so it's a real `GetCheckpoint` call. Can we fetch the timestamp first and only update both cursors once it succeeds? The old code did all the fallible calls before touching state.

crates/hashi-monitor/src/rpc/sui.rs
```diff
@@ -315,93 +399,64 @@ impl SuiEventsPoller {
...
+            ]))
+            .with_start_checkpoint(start_checkpoint)
+            .with_end_checkpoint(end_checkpoint)
+            .with_filter(self.transaction_filter())
+            .with_options(
+                QueryOptions::default()
+                    .with_limit(PAGE_SIZE)
+                    .with_ordering(Ordering::Ascending),
+            );
+        let stream = self.client.list_transactions(request);
```
0xsiddharthks, line 415: `Client::list_transactions` retries transient errors indefinitely (Unavailable, DeadlineExceeded, ResourceExhausted, Cancelled/Unknown mid-stream). `ListConfig` has no attempt cap or deadline, and with no observer the retries don't log anything. So if the fullnode is down or keeps returning 429/503, this call doesn't return until it recovers, and the MAX_RANGE_ATTEMPTS/split loop above never sees those errors. `tick_sui` is awaited inside the `select!` in continuous.rs, so guardian polling, BTC checks and violation detection wait on it too. The raw `ledger_client().list_transactions` loop from bd3d9cc didn't have this problem; otherwise a `tokio::time::timeout` around the scan would get the old fail-fast behavior back.

crates/hashi-monitor/src/rpc/sui.rs
```diff
@@ -39,6 +40,49 @@ use crate::domain::utc_timestamp;
...
+        (_, Some(completed_checkpoint)) => completed_checkpoint,
+        (end_reason, None) => {
+            anyhow::bail!("unexpected Sui transaction scan end reason {end_reason:?}");
+        }
+    };
+    anyhow::ensure!(
+        completed_checkpoint < end_checkpoint,
+        "Sui transaction scan watermark checkpoint {completed_checkpoint} is outside requested range ending at {end_checkpoint}"
+    );
+    if end_reason == QueryEndReason::CheckpointBound {
```
0xsiddharthks: Edge case, but I think this can reject a normal response. If the SDK resumes with a cursor that's already at the end of the range (e.g. an ItemLimit page ending on the range's last transaction), the server sees an empty interval and ends with CheckpointBound without setting `checkpoint` (`ScanTerminal::into_watermark` in sui-rpc-api's watermark.rs), and the SDK passes that through. `completed_checkpoint` then holds the earlier item watermark, which is at most `end_checkpoint - 2`, so this fails after the retry loop, and since the cached head doesn't refresh, the next tick sends the same range. CheckpointBound already means the whole range was scanned, so taking `end_checkpoint - 1` there like the old code did and only relying on the watermark for LedgerTip seems simpler.

crates/hashi-monitor/src/rpc/sui.rs
```diff
@@ -113,67 +159,89 @@ impl SuiEventsPoller {
...
-        for transaction in raw_transactions {
+
+        let Some(completed_checkpoint) = completed_checkpoint_for_scan(
+            start_checkpoint,
+            end_checkpoint,
+            scan.end_reason,
+            scan.completed_checkpoint,
+        )?
+        else {
+            return Ok(PollOutcome::CursorUnmoved);
```
0xsiddharthks: Minor: when LedgerTip hasn't covered anything new we return CursorUnmoved with no log, so index lag just looks like a stuck cursor. batch.rs has no sleep between polls and gives up after 5 stalled iterations, so a short index lag near the end of a batch window can fail the audit. mirror.rs sleeps 1s on the same LedgerTip case. A debug log with the end reason and watermark here, plus a short wait before re-polling, would cover it.

crates/hashi-types/src/guardian/mod.rs
```diff
@@ -138,6 +134,11 @@ pub struct GuardianInfo {
         with = "crate::guardian::serde::option_hex_32"
     )]
     pub genesis_state_hash: Option<[u8; 32]>,
+    /// The Hashi shared-object id this guardian serves (set after
+    /// operator_init). Trailing and omitted when absent to preserve signing
+    /// bytes for GuardianInfo records emitted before this field was added.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub hashi_object_id: Option<sui_sdk_types::Address>,
```
0xsiddharthks: Log signatures are over the BCS bytes of the message, so moving this field changes the signing preimage for GuardianInfo. That fixes records from pre-#1121 enclaves like the testnet fixture, but anything built between #1121 and this commit encodes `hashi_object_id` right after `current_committee_epoch` (a `0x00` byte when it's None). Those records won't verify anymore, whether the field is set or not. `VerifiedSessionInfo::read_from_s3` checks this record before anything else, so GuardianReader (monitor, init tooling) can't load those sessions at all. 63c5e6720 has the old layout, and I think that's what devnet and testing are on. If those still need to work we need to accept both layouts, otherwise it's worth calling out that they need a redeploy.

Cargo.toml
```diff
@@ -4,10 +4,10 @@ members = ["crates/*"]
 
 [workspace.dependencies]
 # sui sdk crates
-sui-sdk-types = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "d71f9d6f788afc6b30baef85a342e0e7867c0019", features = ["serde", "hash"] }
-sui-crypto = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "d71f9d6f788afc6b30baef85a342e0e7867c0019", features = ["ed25519", "secp256k1", "secp256r1", "pem", "bech32"] }
-sui-rpc = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "d71f9d6f788afc6b30baef85a342e0e7867c0019" }
-sui-transaction-builder = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "d71f9d6f788afc6b30baef85a342e0e7867c0019" }
+sui-sdk-types = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "f9b4465d8e95358b4a0bef225e611797de9b30e7", features = ["serde", "hash"] }
+sui-crypto = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "f9b4465d8e95358b4a0bef225e611797de9b30e7", features = ["ed25519", "secp256k1", "secp256r1", "pem", "bech32"] }
+sui-rpc = { git = "https://github.com/MystenLabs/sui-rust-sdk.git", rev = "f9b4465d8e95358b4a0bef225e611797de9b30e7" }
```
0xsiddharthks: This bumps the SDK for the whole workspace, so hashi-types and the guardian enclave build pick it up too, inside a monitor fix. The watermark part doesn't need it (`Watermark.checkpoint` is already on 0.3.x, mirror.rs reads it), only the move to `Client::list_transactions` does. Could we land the watermark fix on 0.3.x and do the bump on its own?

crates/hashi-types/src/guardian/mod.rs
mskd12: I reverted back the ordering and moved it to https://github.com/MystenLabs/hashi/pull/1160 where I am also adding some tests to catch issues like this.

crates/hashi-monitor/src/rpc/sui.rs
mskd12: Fixed in the latest revision. checkpoint_timestamp is now fetched before either next_checkpoint_to_scan or cursor_seconds is updated, so a failed checkpoint lookup cannot discard events and advance past their range.

crates/hashi-monitor/src/rpc/sui.rs
mskd12: Fixed. The complete SDK list_transactions scan is now wrapped in the existing 60-second request timeout. A timeout enters the monitor’s bounded retry and range-splitting path, preventing Sui unavailability from blocking Guardian polling, Bitcoin checks, and violation detection indefinitely.

crates/hashi-monitor/src/rpc/sui.rs
mskd12: Fixed. CheckpointBound now advances through end_checkpoint - 1, since it means the requested half-open range completed. The completion watermark is used only when the query ends at LedgerTip. The unit test also covers CheckpointBound with no watermark and with stale/final item watermarks.

0xsiddharthks (approved):

PR #1158, opened 2026-09-10
Deal into a nonce batch the presig cursor has already drawn from

zhouwfang:
#193


dlukegordon (approved):
Reviewed head 4f453b8407a2e7a91aba9e35dc49e558cb0280af. No actionable findings. Traced the repair, dealer, certificate, and recovery paths. Source review only; tests and builds were not run.

PR #1159, opened 2026-09-11
Reuse the previous round's signer set when re-dealing a nonce batch

zhouwfang:
#531


dlukegordon (approved):
Reviewed d1fdece474286d0ce1709d4abe959b9d5724ffd8. No actionable findings. Traced replay signer reconstruction, persistence and recovery, deterministic dispersal, and shared signature collection behavior. Source review only; tests and builds were not run.

PR #1160, opened 2026-09-11
[Guardian] Preserve deployed GuardianInfo signatures

mskd12:
## Summary

- preserve deployed V2 GuardianInfo signing bytes when `hashi_object_id` is absent
- retain separate genesis wire types: committee-only V1 and V2 with the required Hashi object id
- verify authentic deployed V1/V2 records for every currently supported Guardian reader shape
- document fixture organization and compact formatting for records larger than 5 KiB

## Incident

Adding `hashi_object_id` changed the reconstructed BCS payload for an existing signed V2 GuardianInfo record in S3, causing a valid historical record to fail signature verification. The same change made the deployed committee-only V1 genesis record unreadable.

The compatibility layer now preserves both deployed signed layouts while current writers continue emitting V2 records.

## Verification

- `cargo check`
- `cargo test -p hashi-types deployed_log_fixtures_remain_compatible`
- 20-day testnet Guardian/S3 backfill through 2026-09-11T16:00:00Z

`make fmt` completed Rust and Buf formatting but could not run `prettier-move` because it is not installed locally.


crates/hashi-types/src/guardian/mod.rs
```diff
@@ -124,6 +124,7 @@ pub struct GuardianInfo {
     /// The Hashi shared-object id this guardian serves (set after
     /// operator_init). Certificates verified by this enclave must be bound
     /// to it; operators/KPs match it against their expected deployment.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
0xsiddharthks: Ceremony enclaves log this record too, and they never set `hashi_object_id` (only withdraw mode calls `install_config`), so it's None there. Builds since the field was added signed that None as an explicit `0x00` byte, and with `skip_serializing_if` the reader now leaves it out, so those ceremony sessions won't verify. The withdraw enclave reads ceremony state through GuardianReader in `operator_init`, so a guardian built from this couldn't init against an existing ceremony from one of those builds (I think the current devnet and testing ones). All the fixtures come from testnet builds that predate the field, so CI won't catch it. Could we also accept the explicit-None encoding when the field is None, and add a ceremony GuardianInfo fixture from devnet or testing?

0xsiddharthks (commented):
LGTM

0xsiddharthks (approved):

PR #1162, opened 2026-09-11
Abort reconfig followups

bmwill:


crates/hashi/src/mpc/service.rs
```diff
@@ -1085,12 +1098,48 @@ impl MpcService {
...
+    /// Put the `MpcManager` back on the current epoch once the
+    /// reconfiguration that displaced it is dead. `handle_reconfig` installs
+    /// a manager for its target before running the protocol, and every
+    /// give-up exit leaves it there; after an abort that manager is for an
+    /// epoch that will never exist. The current epoch's `SigningManager`
+    /// survived the whole time, which is exactly why `sync_if_stale` rebuilds
+    /// nothing, yet presignature refill and batch repair both run through
+    /// `ensure_manager_epoch` and fail against the dead manager, so the pool
+    /// could not refill until the next `end_reconfig` landed. A fresh manager
+    /// is all they need: the recovery path builds the same one from scratch.
```
zhouwfang, line 1110: "The recovery path builds the same one from scratch" isn't right. recover_current_dkg and recover_current_rotation go on past setup_* to reconstruct the outputs, and a fresh manager has both None. So a node serving epoch E from previous_output while pinned on the dead target now answers NotFound, which is the path a new member uses during the replacement rotation.

zhouwfang (commented):
The third issue from chat is still open.

crates/hashi/src/mpc/service.rs
```diff
@@ -1085,12 +1098,48 @@ impl MpcService {
...
+            self.setup_key_rotation(epoch)
+        } else {
+            self.setup_initial_dkg(epoch)
+        };
+        match result {
+            Ok(()) => info!(
+                "sync_if_stale: restored the MpcManager for epoch {epoch}; a reconfiguration \
+                 that did not complete had left it on {pinned:?}"
+            ),
+            Err(e) => error!(
```
zhouwfang, line 1129: No backoff here, and sync_if_stale runs every checkpoint, so a persistent failure rebuilds the manager synchronously at checkpoint rate. The same rebuild in handle_reconfig sleeps at :1851 first.

crates/hashi/src/mpc/service.rs
```diff
@@ -220,6 +220,19 @@ impl MpcService {
...
+                // Sui with nothing pending, whether or not this node can
+                // rebuild its SigningManager for the lagging epoch. Submit
+                // from inside the loop: a committee node that cannot rebuild
+                // would otherwise never reach the submission below, and a
+                // fleet of such nodes would never recover.
+                self.try_submit_start_reconfig(
+                    self.inner.onchain_state().latest_checkpoint_epoch(),
+                )
+                .await;
+                if self.get_pending_epoch_change().is_some() {
```
zhouwfang, line 233: No visibility wait after a successful submit, so when the mirror lags this reads None and submits again ten seconds later. The duplicate aborts on chain and still costs gas. abort_stale_reconfig waits 60 seconds for exactly this.

bmwill:
@zhouwfang ok your comments should be addressed now. I still left off the 3rd item from this as i still need to think it through a bit more.

design/docs/node-operator-runbook.mdx
```diff
@@ -1092,6 +1095,22 @@ hashi committee abort-reconfig
...
+
+The chain accepts it only while nothing is pending and Hashi's epoch lags Sui's; the command checks both up front and refuses otherwise. Every running node then drives the new reconfiguration as usual.
+
+To stay on the previous committee instead, for example while the cause of the stall is fixed so the replacement does not stall the same way, have governance set the `reconfig_hold` config flag before the abort lands:
+
+```bash
+hashi proposal create update-config reconfig_hold bool:true
+```
+
+Once the proposal executes, `start_reconfig` aborts on chain, nodes stop submitting it (`hashi_reconfig_hold` reads `1`), and the previous committee keeps serving deposits and withdrawals. The overrun reconfiguration is still aborted as usual, and a reconfiguration already pending still completes. Clear the flag with another `update-config` proposal (`bool:false`) and the nodes start the replacement at once.
```
zhouwfang: Clearing the flag doesn't restart anything right away. Nodes only try again on an epoch change, so the replacement waits for the next one unless someone runs start-reconfig. The same wording is in reconfiguration.mdx.

crates/hashi/src/mpc/service.rs
```diff
@@ -1085,12 +1134,56 @@ impl MpcService {
...
+            Ok(_) => {
+                *self.next_manager_restore.lock().unwrap() = None;
+                info!(
+                    "sync_if_stale: restored the MpcManager for epoch {epoch}; a \
+                     reconfiguration that did not complete had left it on {pinned:?}"
+                );
+            }
+            Err(e) => {
+                *self.next_manager_restore.lock().unwrap() = Some((epoch, now() + RETRY_INTERVAL));
+                error!(
```
zhouwfang: This retry never fires if the failure happens after setup, because the manager is already on this epoch by then. Nodes that joined this epoch will hit that often, since they need the previous output from peers that may still be restoring.

crates/hashi/src/mpc/service.rs
```diff
@@ -1716,6 +1820,13 @@ impl MpcService {
             };
             match result.await {
                 Ok(()) => {
+                    // The chain has the pending change before the mirror
```
zhouwfang: This comment and the one at 502 shouldn't mention gas. build() simulates first, so a failing tx never gets submitted.

zhouwfang (approved):
The flag looks better than a voted abort. Have you thought about an expiry epoch instead of a bool? It would cap how long a stale committee can stay on, and the restart would then happen on an epoch change.

PR #1164, opened 2026-09-11
[Guardian] Store deployment state in genesis record

mskd12:
## Summary

- bind the initial committee, Hashi object ID, and MPC master G together in the KP-authorized `GenesisState`
- persist the complete deployment state in the undeployed V2 genesis record while retaining V1 read compatibility
- keep the current `InitConfig` inputs until a fresh V2 genesis exists after the testnet wipe

## Follow-up

[IOP-728](https://linear.app/mysten-labs/issue/IOP-728/guardian-source-deployment-state-from-the-v2-genesis-record) tracks sourcing the object ID and MPC master G from verified V2 genesis and removing their duplicate `InitConfig` inputs after the wipe.

Linear: [IOP-610](https://linear.app/mysten-labs/issue/IOP-610/guardian-store-mpc-master-g-in-the-genesis-record)

## Testing

- `cargo check`
- `cargo fmt --check -- --config imports_granularity=Item --config format_code_in_doc_comments=true`
- `buf format --diff crates/hashi-types/proto/sui/hashi/v1alpha/guardian_service.proto`
- `make -C crates/hashi-types proto`


mskd12:
@bmwill If hashi object ID is intended to be permanent, then i think a better home is the genesis S3 log record. So when doing the testnet wipe, we can remove it from InitConfig (per-instance initialization config).

crates/hashi-types/src/guardian/s3/log_messages/genesis.rs
```diff
@@ -17,10 +17,11 @@ pub struct GenesisLogMessageV1 {
...
 pub struct GenesisLogMessageV2 {
     pub committee: crate::move_types::Committee,
-    /// The Hashi shared-object id this guardian was bootstrapped for, for
-    /// audit self-description (the authoritative pin lives in `InitConfig`,
-    /// covered by `config_hash`).
+    /// The Hashi shared-object id this guardian was bootstrapped for.
     pub hashi_object_id: sui_sdk_types::Address,
+    /// DKG-derived MPC master G pinned by the operator and independently
+    /// authorized by KPs during genesis.
+    pub mpc_master_g: crate::bitcoin::HashiMasterG,
```
0xsiddharthks: This adds a required field to V2 in place rather than adding a V3, and readers pick the payload type purely off the record's `schema_version`. So a `genesis/record.json` already written as V2 (the committee + hashi_object_id shape from #1121) now fails to decode with `invalid V2 log message: missing field mpc_master_g`. `read_latest_committee` only falls back to the genesis record while no `committee-update/` success exists, so an env still in that window wouldn't be able to run operator_activate. Testnet's record is V1 so it's fine there — do we know whether devnet or testing bootstrapped genesis after #1121 landed?

crates/hashi-types/src/guardian/proto_conversions.rs
```diff
@@ -773,8 +789,13 @@ pub fn operator_init_request_to_pb(
...
                     s3_credentials: Some(s3_credentials.into()),
                     init_config: Some(init_config_to_pb(init_config)?),
-                    genesis_state: genesis_state.map(|state| pb::GenesisState {
-                        committee: Some(move_committee_to_pb(&state.into_committee())),
+                    genesis_state: genesis_state.map(|state| {
+                        let (committee, hashi_object_id, mpc_master_g) = state.into_parts();
+                        pb::GenesisState {
+                            committee: Some(move_committee_to_pb(&committee)),
+                            hashi_object_id: hashi_object_id.as_bytes().to_vec().into(),
+                            mpc_master_g: mpc_master_g.to_byte_array().to_vec().into(),
```
0xsiddharthks, line 797: Nothing covers this encode/decode pair: `OperatorInitRequest::mock_for_testing()` passes `genesis_state: None` so `operator_init_request_round_trip` skips it, and `OperatorInitTestArgs::with_genesis_state` has no callers, so no test writes a genesis record either. I checked it by hand and it does hold — round-tripping a non-normalized point keeps the `GenesisState` digest stable, since fastcrypto encodes `G` through its 33-byte compressed form — but this is the one shot per deployment, with the KPs assembled, so it seems worth locking in. Passing `Some(..)` in the mock covers the wire half in one line.

0xsiddharthks (commented):
Traced the digest path end to end (operator pin -> proto -> enclave -> KP re-derivation) and it lines up. Two things below: a compat question on the V2 record, and the genesis path having no test coverage.

0xsiddharthks (commented):
Two nits I'd left out of the last pass, neither blocking.

0xsiddharthks (approved):
Approving. The V2 record question above is the only open thing and it is not a blocker.

PR #1165, opened 2026-09-11
[Guardian] Publish ceremony logs after confirmation

mskd12:
## Summary

- stage setup and rotation output in a session-scoped `kp-shares/proposed/` record
- publish finalized KP shares followed by the ceremony commit only after every KP confirms
- have ceremony participants read the exact live-session proposal while existing readers continue consuming finalized state

## Testing

- `cargo check`
- `make fmt` *(Rust and protobuf formatting completed; Move formatting could not run because `prettier-move` is not installed locally)*

### DIFF
--- a/crates/hashi-guardian-init/src/kp_ceremony.rs
+++ b/crates/hashi-guardian-init/src/kp_ceremony.rs
@@ -64,23 +64,42 @@ pub async fn run(cfg: Config, encrypted_shares_path: &Path) -> Result<()> {
     // The selected cert identifies this KP's roster entry.
     let kp_pgp_cert_path = cfg.require_kp_pgp_cert_path("key-provisioner ceremony")?;
     let kp_cert = load_kp_cert(kp_pgp_cert_path)?;
+    let kp_fingerprint = kp_cert.fingerprint();
     certs_roster
-        .cert_for_fingerprint(&kp_cert.fingerprint())
+        .cert_for_fingerprint(&kp_fingerprint)
         .with_context(|| {
             format!(
                 "this KP's cert (fingerprint {}) is not among the configured \
                  kp_roster.kp_pgp_cert_paths",
-                kp_cert.fingerprint()
+                kp_fingerprint
             )
         })?;
     info!(
         phase = "setup",
-        fingerprint = %kp_cert.fingerprint(),
+        fingerprint = %kp_fingerprint,
         "identified this KP's configured certificate",
     );
 
-    // 1. Discover and verify the latest ceremony from the immutable log
-    //    (attestation-verified once via the reader's session-key cache).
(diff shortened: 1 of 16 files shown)


0xsiddharthks (commented):
One blocker below — I think the key-provisioner ceremony command can't get past its new precondition. The rest of the restructuring reads fine to me: the proposal/commit split, the lock tiers, and the digest agreement between the proposal and the SetupNewKeyResponse all check out.

crates/hashi-guardian-init/src/kp_ceremony.rs
```diff
@@ -64,23 +64,45 @@ pub async fn run(cfg: Config, encrypted_shares_path: &Path) -> Result<()> {
...
+            .await?;
+    ensure!(
+        verified.info.lifecycle == CeremonyStage::AwaitingKeyProvisionerConfirmations.into()
+            || verified.info.lifecycle == CeremonyStage::Completed.into(),
+        "guardian is not accepting key provisioner ceremony confirmations"
+    );
+    let expected_instance = verified
+        .info
+        .secret_sharing_instance
+        .context("live ceremony GuardianInfo missing secret_sharing_instance")?;
```
0xsiddharthks: I think this can never succeed. `GuardianInfo.secret_sharing_instance` is only ever populated from `temporary_init_state` (`enclave.rs:634`), and that's set in exactly one place — `operator_init.rs:96`, inside `OIWithdrawModeInstall::install_into`, which only runs for a withdraw-mode enclave. `operator_init_state_installed` spells it out: "a ceremony enclave installs nothing else." The `ensure!` right above this guarantees we're talking to a ceremony-mode guardian, so the field is always `None` here and every KP bails with "live ceremony GuardianInfo missing secret_sharing_instance" — setup and rotation alike.

I checked with a throwaway test rather than just reading: run `setup_new_key` on a ceremony enclave, then `enclave.info().await` at `AwaitingKeyProvisionerConfirmations`, and `secret_sharing_instance` is `None`. CI stays green because the guardian-side tests call `setup_new_key`/`confirm_ceremony` directly and never go through `info()`, and there's no test driving the KP CLI against a live ceremony enclave.

Simplest fix is probably to have `info()` fall back to the pending ceremony's instance when the enclave is in ceremony mode, which would also make the cross-check here meaningful.

mskd12: Good catch. secret_sharing_instance now comes from the signed, session-scoped ceremony proposal in S3. GetGuardianInfo is only used to verify and pin the live session and lifecycle, so I removed this precondition and the redundant cross-check.

0xsiddharthks (approved):
LGTM!

PR #1166, opened 2026-09-11
[Guardian] Discover withdrawal prefixes from S3 version history

mskd12:
When delete markers hide every current object in the newest withdrawal hour, `ListObjectsV2` omits that hour from prefix discovery. Limiter recovery can then select an older success record without reaching the existing deletion check.

Use `ListObjectVersions` with `/` as the delimiter for prefix discovery, including key/version marker pagination. The walk now discovers retained history and the existing success-record check rejects delete markers. Update the layout mocks and add a regression case covering a deleted newest success with an older success still present.

Addresses the hidden-prefix defect in [SEC-527](https://linear.app/mysten-labs/issue/SEC-527). Genesis fallback and node-side guards are unchanged; this does not detect permanently removed history.

Validation:
- Focused latest-success-bucket tests: 6 passed.
- `cargo check`: passed.
- `make fmt`: Rust and protobuf formatting completed; Move formatting could not run because `prettier-move` is not installed.

### DIFF
--- a/crates/hashi-guardian/src/s3_client.rs
+++ b/crates/hashi-guardian/src/s3_client.rs
@@ -411,31 +411,32 @@ impl GuardianS3Client {
     // S3 Reads
     // ========================================================================
 
-    /// Lists immediate subdirectories under `prefix` (S3 `CommonPrefixes`,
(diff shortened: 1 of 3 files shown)


0xsiddharthks (approved):
LGTM

PR #1168, opened 2026-09-14
[MPC] Add BlsSignatureAggregator::reduced_weight_reached

jonas-lj:
Adds `reduced_weight_reached(weight)` to the reduced-weight `BlsSignatureAggregator` and uses it for the five threshold checks in `mpc_except_signing.rs` that compared `reduced_weight()` against a `u32` threshold (DKG cert, rotation cert, AVID vote quorum, and the two grace-deadline checks in `collect_dealer_signatures`).

This replaces the mix of `u32::from(aggregator.reduced_weight()) >= ...` and `aggregator.reduced_weight() as u32 >= ...` at those sites with a single helper. Sites that reuse the weight for arithmetic or logging (`confirmed`, `collected`) are left unchanged.

Also folds the drained and timeout stop reasons in the `collect_dealer_signatures` loop into a single match on `timeout_at(..)`, instead of a match followed by a separate `let Some(..) else`.

No behavior change.


crates/hashi-types/src/committee.rs
```diff
@@ -732,6 +732,10 @@ impl<'a, T: IntentMessage + Clone> BlsSignatureAggregator<'a, T, ReducedWeight<'
     pub fn reduced_weight(&self) -> u16 {
         self.domain.signed
     }
+
+    pub fn has_reduced_weight(&self, weight: u32) -> bool {
+        self.domain.signed as u32 >= weight
```
zhouwfang: Could this use u32::from instead of as? It fails to compile if signed ever grows past u32, where the cast would quietly truncate.

crates/hashi-types/src/committee.rs
```diff
@@ -732,6 +732,10 @@ impl<'a, T: IntentMessage + Clone> BlsSignatureAggregator<'a, T, ReducedWeight<'
     pub fn reduced_weight(&self) -> u16 {
         self.domain.signed
     }
+
+    pub fn has_reduced_weight(&self, weight: u32) -> bool {
```
zhouwfang: On the function name, every other has_x here is an existence check with no argument, so has_reduced_weight(w) reads like equality. Something like reduced_weight_reached would match floor_reached nearby.

zhouwfang (approved):

PR #1170, opened 2026-09-15
[screener][dead code] Remove the hashi-screener service

0xsiddharthks:
## Summary

Screening moves into the node: each operator will screen deposits and withdrawals with their own TRM Labs API key instead of calling the Mysten-run MerkleScience proxy. This removes the proxy and leaves the deposit and withdrawal screening hooks as stubs for the next PR in the stack.

## Changes

- Delete `crates/hashi-screener`, `docker/hashi-screener`, `screener_service.proto` and its generated code
- Drop the screener image from `deterministic-build.yml`, along with the now-unused `hashi_only` input and its use in `release.yml`
- Remove the `screener-endpoint` config key, the screener client and the `hashi_screener_enabled` gauge; a config that still sets the key fails to load, and no Mysten node config sets it
- Stub `screen_deposit` and `screen_withdrawal`, keeping the AML error variants the TRM integration uses
- Remove the mock screener from the e2e network
- Remove screener references from the sanctions, deposit, withdraw and node operator runbook docs


dlukegordon (approved):
Reviewed commit 4e29b06dfdf50488bc3f1634a45d8a7d426dd6de. No actionable findings. The screener removal is consistent with the stated scope, including the intentional interim screening stubs. Verification was source-only; no tests or builds run.

PR #1174, opened 2026-09-15
[withdrawals] Stop capping the withdrawal fee rate at 30 sat/vB

0xsiddharthks:
> Stacked on #1190.

## Summary

The leader clamps bitcoind's estimate to 30 sat/vB and caps the configured floor there too. Above 30 sat/vB every batch is committed on Sui underpriced, where it can't be RBF'd, and later batches chain off its change. Once the mempool minimum passes 30, `sendrawtransaction` rejects it outright, and no config lever can raise the price. The on-chain per-request budget already bounds what users pay, so price at the estimate floored at the configured minimum, and let selection fail and retry while the budget can't cover it.

Mixed fleets are fine: validators already cap at 3x their own uncapped estimate, so old validators accept an uncapped leader. Miner fees still count against the guardian limiter, bounded by the same per-request budget.

## Changes

- `withdrawals.rs`: `withdrawal_fee_rate` (estimate floored at `withdrawal_min_fee_rate_sat_vb`) replaces the clamp and `effective_min_fee_rate` for both the leader and the validator ceiling.
- The ceiling's error message no longer calls the estimate "clamped".
- `utxo_pool`: `DEFAULT_HIGH_FEE_RATE_THRESHOLD` now only switches off consolidation; the `FeeExceedsCap`, `max_fee_per_request` and `high_fee_rate_threshold` docs all say what actually bounds the fee.
- Drop `test_configured_floor_capped_at_high_fee_threshold`; `withdrawal_fee_rate_floors_the_estimate_without_capping_it` replaces it.
- Docs: the rate is floored, not capped.
- `fees.mdx` notes that a floor rescue past 3x has to be configured across the fleet, since each validator prices the ceiling off its own floor.


bmwill (commented):
I think we'll need to have a short conversation on the impacts of this PR given our gas selection algorithm and default configured minimums for deposit and withdrawal

PR #1177, opened 2026-09-15
[sanctions check] Add TRM integration for screening deposits and withdrawals

0xsiddharthks:
## Summary

Each mainnet committee member screens deposits and withdrawals with its own TRM Labs API key (`trm-api-key`) and withholds its signature when TRM flags the transfer or an address, or can't give an answer. A deposit registers its Bitcoin transaction with Transaction Monitoring and screens the Sui address that receives the hBTC; a withdrawal screens its Bitcoin destination and the requesting Sui address with Wallet Screening.

## Changes

- Add a TRM client (`crates/hashi/src/trm`) for `POST /public/v2/screening/addresses`, `POST /public/v2/tm/transfers` and `GET /public/v2/tm/transfers/{uuid}`, using the API key for HTTP Basic auth
- Deposits: register the Bitcoin transaction as a `CRYPTO_DEPOSIT` to the deposit address under `externalId` `hashi-deposit-<request id>`, so every retry and every leader gets the same transfer back; screen the Sui address the hBTC mints to (`utxo.derivation_path`, SEC-501); then read the transfer every 2s for up to 10s
- Withdrawals: screen the Bitcoin destination and the requesting Sui address in one batch, and require a result for each address submitted
- Reject when TRM attributes a screened address itself (`OWNERSHIP`) to a High or Severe category, or when the transfer's `riskScoreLevel` is High or Severe; counterparty or indirect exposure alone doesn't reject an address, since dust from a flagged source would otherwise taint any address
- A transfer still `PROCESSING` after 10s is retried on the next Bitcoin block; a transfer TRM can't screen (`FAILED`, or any status other than `PROCESSING` and `SUCCEEDED`) is a permanent rejection, since TRM keeps the first transfer registered under an `externalId`
- Retry only timeouts, transport errors, 429 and 5xx, within one 20s deadline per screening; any other API error or unexpected response fails permanently
- Space requests to each endpoint at least 125ms apart (TRM allows 10 per second per endpoint); a request whose turn would start after its deadline fails at once as transient, and a 429 holds that endpoint for `Retry-After` (capped at 60s)
- Log every rejection, and every screening TRM can't complete, on the node that decided
- Add the optional `trm-api-key` config key; the node builds a TRM client only with a key on Bitcoin mainnet
- Export `hashi_trm_enabled`, `hashi_trm_screenings_total` and `hashi_trm_screening_duration_seconds` (by `flow` and `outcome`)
- Test the client against a mock TRM server with scripted responses, and the transfer polling and request pacing against a paused clock
- Add `#[ignore]`d live tests (`TRM_API_KEY=<key> cargo nextest run -p hashi --run-ignored only trm::tests::live`): an OFAC-listed Bitcoin destination, and the Cetus exploiter's Sui address as a withdrawal requester or deposit recipient, are rejected; a real taproot deposit and the genesis address are approved. CI runs them with the `TRM_API_KEY` secret on pull requests from this repository that change the client, and on manual runs

## Questions for TRM

- We decide on a transfer's `riskScoreLevel` as soon as `screenStatus` is `SUCCEEDED`. Your webhook guide says initial-screening alerts are issued within about 5 minutes of submission. Is every initial-screening alert reflected by `SUCCEEDED`, or can one land after it?
- Nodes have no price feed, so `fiatValue` is `"0"`, and the deposit request's creation time on Sui stands in for the Bitcoin transaction time (screening still succeeded with timestamps 2h off). Does that limit Transaction Monitoring?
- We don't register withdrawals with TM, because `request_withdrawal` moves hBTC into an object rather than to an address and the Bitcoin transaction is built later. Should nodes register the confirmed Bitcoin withdrawal transaction for monitoring, as your implementation patterns do?
- Nodes resubmit the same `externalId` and re-screen the Sui address on every retry. Are those billed as new screenings?
- Should hashi send `x-trm-partner-id`? If so, could you assign one?
- A US-tenant key gets 401 from `api.eu.trmlabs.com`. Should hashi let operators on the EU tenant choose that base URL?


crates/hashi/src/trm/mod.rs
```diff
@@ -0,0 +1,1108 @@
...
+                if status == reqwest::StatusCode::TOO_MANY_REQUESTS || status.is_server_error() {
+                    TrmError::Transient(error)
+                } else {
+                    TrmError::Permanent(error)
+                },
+            );
+        }
+        response.json().await.map_err(|e| {
+            if e.is_decode() {
+                TrmError::Permanent(anyhow!("unexpected TRM response: {e}"))
```
dlukegordon: **P1 — Keep response-body transport failures retryable**

At `crates/hashi/src/trm/mod.rs:312–316`, `is_decode()` also covers transport failures while reading the response body: reqwest 0.12.28’s `bytes()` wraps body-collection errors as decode errors. If TRM returns successful headers but the connection drops mid-body, this permanently rejects the screening, and the leader marks the deposit or withdrawal as never retry. Please read the body separately and classify read failures as transient, then deserialize the completed bytes and classify JSON/schema failures as permanent. A truncated-response regression test would catch this distinction.

crates/hashi/src/trm/mod.rs
```diff
@@ -0,0 +1,1108 @@
...
+            let turn = (*next_turn).max(Instant::now());
+            if turn > deadline {
+                return Err(TrmError::Transient(anyhow!(
+                    "TRM requests are queued past the screening deadline"
+                )));
+            }
+            *next_turn = turn + REQUEST_INTERVAL;
+            turn
+        };
+        tokio::time::sleep_until(turn).await;
```
dlukegordon: **P2 — Apply Retry-After to already-queued requests**

At `crates/hashi/src/trm/mod.rs:440–459`, a 429 cooldown does not affect requests already sleeping in `wait_turn()`. For example, B reserves a turn at +125ms, then A receives `Retry-After: 2`; B still sends at +125ms because its wake-up never rechecks the hold. Normal concurrent deposit/withdrawal processing reaches this case. Please make waiting requests honor an updated cooldown before sending, while preserving request spacing and deadline checks. Add coverage where the hold arrives after another request has queued.

.github/workflows/ci.yml
```diff
@@ -82,6 +82,23 @@ jobs:
...
+        if: github.event_name == 'pull_request'
+        with:
+          filters: |
+            trm:
+              - 'crates/hashi/src/trm/**'
+              - '.github/workflows/ci.yml'
+
+      - name: Run live TRM tests
+        # These call TRM's API, so leave them to the PRs that change the client.
+        if: steps.trm-filter.outputs.trm == 'true' || github.event_name == 'workflow_dispatch'
```
dlukegordon, line 98: **P2 — Skip live TRM tests when credentials are unavailable**

At `.github/workflows/ci.yml:97–100`, this condition also enables live TRM tests for fork PRs, where `secrets.TRM_API_KEY` is unavailable and expands to an empty string ([GitHub documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow)). `live_client()` accepts that value, and the tests fail when they unwrap the unauthenticated API response. Please gate the live-test step on credential availability or restrict it to trusted same-repository/manual runs, leaving ordinary tests enabled for forks.

dlukegordon (commented):
Three findings from reviewing the TRM screening integration: transport-error classification, rate-limit cooldown handling, and credential-dependent CI.

crates/hashi/src/trm/mod.rs
0xsiddharthks: Fixed in 24edb7a87. `send` now reads the body with `bytes()` and treats any read failure as transient, then parses with `serde_json::from_slice`, so only a complete body that doesn't parse is permanent. `a_truncated_response_is_transient` sends the headers and 2 of 64 body bytes, then closes; it fails against the old `is_decode()` split.

crates/hashi/src/trm/mod.rs
0xsiddharthks: Fixed in 974fd1a29. A 429 now sets a separate `hold_until`, and a request that wakes for a turn it took before the hold re-queues behind it, keeping the 125ms spacing and the deadline check. `a_rate_limit_holds_back_a_turn_already_taken` lands the 429 while a second request is queued; before the fix that request went out at 125ms instead of 2.05s.

.github/workflows/ci.yml
0xsiddharthks: Fixed in a180d44d6. The TRM change filter now runs only when the PR's head repo is this repository, so fork PRs skip the live tests while the rest of CI runs; manual dispatches still run them.

dlukegordon (approved):
All three review findings are addressed. Verified the fixes and regression coverage by source inspection; CI passed on this head.

PR #1182, opened 2026-09-15
[hashi-monitor] Export audit metrics

0xsiddharthks:
## Summary

Part of productionizing the monitor (IOP-721): alerting pages on metrics rather than log lines, since a dead, stuck or blind monitor logs nothing. `continuous` now serves Prometheus gauges for the latest finding per category and for how far each source has been checked, so rules can page on safety findings and on a monitor that stops verifying.

## Changes

- New `metrics.rs`: `hashi_monitor_last_finding_timestamp_seconds{category}` and `hashi_monitor_checked_through_timestamp_seconds{source}`, all series exported at 0 from startup, served at `/metrics` like the guardian proxy.
- `ContinuousAuditor` routes every finding through `report_findings` (log + gauge) and updates the Sui and guardian cursors per advance and the Bitcoin time after each successful lookup cycle. No source reports a check it has not made, so the Bitcoin gauge stays at 0 until a cycle succeeds.
- `continuous --metrics-listen-addr` (default `0.0.0.0:9184`), bound before the audit starts so a port the monitor cannot serve fails startup instead of leaving it unobserved; `batch` is unchanged.
- Adds `axum` and `prometheus` workspace dependencies to `hashi-monitor`.


bmwill (approved):

crates/hashi-monitor/src/audit/continuous.rs
```diff
@@ -70,43 +81,58 @@ impl ContinuousAuditor {
...
         while let PollOutcome::CursorAdvanced(events) = self.inner.poll_guardian().await? {
             self.ingest_batch(events);
+            self.metrics
+                .set_checked_through(SOURCE_GUARDIAN, self.inner.get_guardian_cursor());
         }
         Ok(())
     }
 
     /// Throws an error if BTC RPC infra fails.
     fn tick_btc(&mut self) -> anyhow::Result<()> {
```
mskd12: Unlike Guardian & Sui reads, we just do point lookups for Bitcoin. So "checked through" does not make sense as a metric, no?

PR #1183, opened 2026-09-15
[hashi-monitor] Default the continuous start for restarts

0xsiddharthks:
> Stacked on #1182.

## Summary

Part of productionizing the monitor (IOP-721): a restarted service has no one to pass `--start`, and the monitor keeps no state. `continuous` now defaults its start to the earliest time whose checks can still be pending, so a restart re-audits exactly the window that can still produce findings.

## Changes

- `ContinuousAuditWindow::default_start`: now minus the longest next-event delay, the clock skew, the guardian cursor's own lag, and one poll plus state tick for a missed deadline to be reported. A guardian approval's deadline is only judged once the hourly cursor passes it, so without that term a restart dropped findings the cursor had not reached yet.
- `MAX_DIR_COMPLETION_LAG` in `hashi-types` names that lag: the hour a reader is still waiting on, plus the delay before that directory counts as complete.
- `continuous --start` is optional and uses that default; its help notes that violations anchored before the default start are not reported again after a restart.
- `NextEventDelays::max_delay`, and a README note on running without `--start`.


bmwill (approved):

mskd12:
The goal should be to cover all times. So in case the monitor goes down, we should set the start time appropriately backwards to achieve that. Should we set a generous default (e.g., a week ago) to keep things simple which should suffice assuming bugs are fixed in a week?

PR #1187, opened 2026-09-15
[trm labs][sanctions check] Pace TRM requests under the rate limit

0xsiddharthks:
> Stacked on #1186.

## Summary

TRM allows 10 requests per second per endpoint for each account, but a node can start dozens of screenings at once: the leader runs up to 32 deposit validations concurrently, and every committee member screens the same fan-out. Pace each TRM endpoint under the limit so a burst waits its turn within the screening deadline instead of failing with 429 and waiting for the next Bitcoin block.

## Changes

- Space requests to each TRM endpoint (transfer registration, transfer reads, address screening) at least 125ms apart
- Compute one deadline per screening and pass it to every request; a request whose turn would start after the deadline fails at once as transient instead of queueing
- On a 429, hold that endpoint until TRM's `Retry-After` has passed (capped at 60s)
- Test the pacing against a paused clock and the 429 hold against the mock TRM server

## Questions for TRM

- Is the 10 requests per second limit per endpoint, as documented, or shared across endpoints for an account?


0xsiddharthks:
Folded into #1177.

PR #1188, opened 2026-09-15
DNS - comments on presigs 

benr-ml:


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2791,6 +2841,20 @@ impl MpcManager {
         })
     }
 
+    // TODO(Must fix): liveness hole vs the spec's second-round send rule (see pdf). The
```
zhouwfang, line 2844: Agreed, and it's worse than that. Once a round goes pessimistic, the retry is pinned to the original confirm set, so one Byzantine member going quiet wedges that dealer for the epoch, and with Byzantine dealers sitting out too, that's the whole batch. I think your second fix is the right one.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -133,6 +133,7 @@ const HEDGED_RETRIEVE_ROUND_TIMEOUT: Duration = Duration::from_secs(1);
 const PREVIOUS_MESSAGE_REPAIR_ATTEMPT_TIMEOUT: Duration = Duration::from_secs(20);
 /// How long the first phase of batch AVSS keeps waiting for unanimity once the
 /// pessimistic fallback is already assured.
+// TODO: I think that 10 or 15 should be enough
```
zhouwfang, line 136: Not sure I agree. The grace keeps slow honest nodes in the vote, so cutting it makes your second-round failure easier to hit.

crates/hashi/src/mpc/signing.rs
```diff
@@ -746,6 +759,17 @@ impl SigningManager {
         }
     }
 
+    // TODO(Must fix): the RS "mismatched" attribution is only sound when the decoded
```
zhouwfang, line 762: Right, but I think it is not must fix. I considered this gate in #1085 and went the other way. The flag that's left is additive, so a wrong one costs nothing but a counter.

crates/hashi/src/mpc/signing.rs
```diff
@@ -334,6 +334,19 @@ impl SigningManager {
     }
 
     #[allow(clippy::too_many_arguments)]
+    // TODO(defence in depth): recovery re-enables every presig slot that is still `Pending`
```
zhouwfang, line 337: Agreed. I closed this last month as not reachable, and your three triggers are right. The comment on HASHI_MPC_RECOVERY_DELAY_SECONDS says governance is expected to change it though, so I'd take your fix before that.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -661,6 +680,37 @@ impl MpcManager {
         Ok(RetrieveOutcome::NeedsStore)
     }
 
+    // TODO[defence in depth]: only non-signers may complain. Cheap.
+    // TODO[defence in depth]: bound what forged complaints can reveal. Two options:
```
zhouwfang, line 684: Agreed, and it's stronger than that. At the shipped default neither cap allows more than a couple of weight units of reveal, so this needs closing structurally rather than budgeting.

crates/hashi/src/mpc/mpc_except_signing.rs
benr-ml: i dont follow - this is the amount of time we wait for more first round votes after we already met the threshold, so it's an optimization. otho, even 1 slow party can cause us to wait 90s and block presigs.

zhouwfang: @benr-ml Agreed it's an optimization, and agreed on the cost. The problem with 10 to 15 is that it's under CALL_TIMEOUT, so it admits no peer that's a retry behind. It was 10 once and got raised to 35 for that reason. I can't defend 90 either. If we cut it, I'd go to 35 and pair it with the resend fix from your send-rule comment.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -525,6 +526,24 @@ impl MpcManager {
         Ok(manager)
     }
 
+    // TODO(Must fix): the message kind is never checked against `self.protocol_type`,
```
zhouwfang, line 529: @benr-ml Fixed in #1191.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: @benr-ml Fixed by #1202.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -661,6 +680,37 @@ impl MpcManager {
...
+    //     B*(W'-f')/(t'-f'), which must stay <= t'-f'-1, so B <= (t'-f'-1)(t'-f')/(W'-f'):
+    //     B = f is safe for f <= 20%, B = 5% is enough for f < 27%, and nothing positive is
+    //     safe near f = 33%.
+    // In both: count distinct dealers in reduced weight (the response cache re-serves a
+    // dealer's shares, so a dealer counts once), count rotation in previous share indices
+    // against the previous committee's t-f-1, persist the counter, refuse beyond the cap,
+    // alarm on the first reveal, log the evidence, and resume only through a manual config
+    // override after review.
+
+    // TODO(defence in depth): invalid complaints are a cheap way for a member to hog the
```
zhouwfang, line 700: @benr-ml Fixed by #1202.

PR #1140, opened 2026-09-09
Require YubiKey attestations for all guardian KP keys

dlukegordon:
- Require verified KP certificate bundles for setup, provisioning signatures, confirmations, and both rotation flows.
- Load sibling attestation PEMs in the CLI and verify them independently inside the guardian and proxy.
- Pin signing, encryption, and recipient checks to the actual attested subkeys, with unit coverage for key binding and trust rejection.
- Part of [IOP-689](https://linear.app/mysten-labs/issue/IOP-689/add-yubikey-openpgp-attestation-to-guardian-kp-setup-flow).

PR #1189, opened 2026-09-15
[guardian] Attest dev KP keys with a software device

0xsiddharthks:
## Summary

Since #1140 every KP certificate needs YubiKey attestations, so dev ceremonies that use gpg softkeys (the devnet and private-testnet deploy, the local replica) fail at `operator ceremony` (sui-operations run 34999910004). `non-enclave-dev` builds now also trust a software device that signs its own certificate, and a dev command writes those attestations, so dev KPs carry the same bundle files through the same verification as real ones. Production builds still trust only the pinned Yubico issuers.

Checked locally with gpg-generated Ed25519/X25519 keys: `tools dev-attest`, then a ceremony-mode guardian, `operator ceremony` and three `key-provisioner ceremony` confirmations against MinIO. Then end to end by a full devnet redeploy with sui-operations #9029 (run 35034877946): ceremony, provisioning and activation green, and `guardian.devnet.hashi.sui.io/info` reports this branch's revision.

## Changes

- `hashi-types`: under `non-enclave-dev`, KP attestation verification also accepts a self-signed device certificate
- `hashi-types` test utils: `dev_kp_attestations` mints device/SIG/DEC PEMs from a self-signed device, sharing the fixture code, and errors on unsupported keys instead of panicking
- `hashi-guardian-init`: `tools dev-attest --kp-pgp-cert-path` (only with `non-enclave-dev`) verifies the bundle and only then writes the PEMs beside the cert, so a rejected one is never left on disk
- `hashi-guardian-proxy`: `non-enclave-dev` feature, since the relay verifies each KP submission's bundle when it decodes it; its Containerfile takes a `FEATURES` build arg
- Local replica: test KPs use the Ed25519/X25519 profile plus `tools dev-attest`, guarded on the attestations themselves so a cert left unattested by an interrupted run is attested on the next one, and the proxy builds with `non-enclave-dev`
- Tests: dev bundles verify only with `non-enclave-dev` and never against the Yubico pins


dlukegordon (approved):
Reviewed at 9f1c683a90c491a78fcf0479bcab70a8ed9dc671. No actionable findings. Software-device trust and the dev-attest command are gated behind non-enclave-dev; builds without that feature retain the pinned Yubico issuers. Review was source-only; tests, builds, and deployed mainnet artifacts were not independently verified.

PR #1192, opened 2026-09-17
[operator] Collect KP public keys through an upload-only S3 bucket

0xsiddharthks:
## Summary

Key provisioner Macs aren't signed in to corporate accounts, so KPs move their public files to the operator by hand. This adds operator scripts that create an upload-only bucket and key to share with KPs, then download the uploads, verify each bundle with the attestation check certificate-loading commands use, and print a fingerprint roster that each KP confirms against their YubiKey. One key serves all KPs: it can only upload, and the verification plus roster confirmation, not its secrecy, protect the result. #1193 adds the KP upload side.

## Changes

- `hashi-guardian-init tools verify-kp-cert`: runs `load_attested_kp_cert` on one certificate and prints its primary-key fingerprint.
- `operator/scripts/create-kp-upload-bucket.sh <name>`: bucket `mysten-hashi-kp-pubkeys-<name>` (us-west-2, versioned, public access blocked) and IAM user `/hashi-kp-pubkeys/hashi-kp-pubkeys-<name>-upload` allowed only `s3:PutObject` on it; tests the new key with a real upload before printing it.
- `operator/scripts/download-kp-pubkeys.sh <name>`: builds the verifier with default features, fetches the latest version of each upload, reports each user ID as VERIFIED, INCOMPLETE, or INVALID, and writes a fingerprint-ordered `roster.txt` only when every ID verifies.
- `operator/scripts/revoke-kp-upload-key.sh <name>`: deletes the upload user and its keys, refusing users outside `/hashi-kp-pubkeys/`; the bucket is kept.
- The scripts follow `key-provisioner/scripts/provision-yubikey.sh`'s structure; `operator/README.md` is the collection runbook.


dlukegordon (approved):
Reviewed at 77d446ab. No actionable findings. Source and documentation review covered the AWS resource lifecycle, certificate verification, download handling, and ceremony roster ordering. No tests, builds, or live AWS operations were run.

PR #1193, opened 2026-09-17
[kp-tooling] Upload KP public keys to the operator's S3 bucket

0xsiddharthks:
## Summary

Lets KPs hand their five public files to the operator straight from the key provisioner Mac by typing three values the operator shares (bucket, access key ID, secret access key), instead of copying them to a corporate laptop. #1192 adds the operator side that creates the bucket and verifies the uploads.

## Changes

- `key-provisioner/scripts/upload-pubkey.sh`: checks the five files agree with each other and with the connected YubiKey, then uploads them using only the typed values, ignoring any local AWS configuration. Input is visible on purpose, since the key can only upload public files and is revoked after collection; after an error, Enter keeps each value.
- `key-provisioner/scripts/show-fingerprint.sh`: prints the connected YubiKey's serial and signing-key fingerprint, in the operator roster's format.
- `provision-yubikey.sh`: ends by printing the exact upload command.
- `provision.md`: the upload and roster-check steps.


key-provisioner/scripts/upload-pubkey.sh
```diff
@@ -0,0 +1,237 @@
...
+  say "Upload the public files"
+  upload_failed=false
+  for file in "${FILES[@]}"; do
+    printf 'Uploading %s\n' "${file##*/}"
+    if ! upload_error="$(
+      env -u AWS_PROFILE -u AWS_DEFAULT_PROFILE -u AWS_SESSION_TOKEN -u AWS_SECURITY_TOKEN \
+        AWS_CONFIG_FILE=/dev/null AWS_SHARED_CREDENTIALS_FILE=/dev/null \
+        AWS_ACCESS_KEY_ID="$ACCESS_KEY_ID" AWS_SECRET_ACCESS_KEY="$SECRET_ACCESS_KEY" \
+        aws s3api put-object --region "$REGION" --bucket "$BUCKET" --key "$USER_ID/${file##*/}" \
+        --body "$file" 2>&1 > /dev/null
```
dlukegordon, line 206: At `key-provisioner/scripts/upload-pubkey.sh:202–206`, this environment still inherits `AWS_ENDPOINT_URL` and `AWS_ENDPOINT_URL_S3`. Pointing the config files at `/dev/null` and passing `--region` does not override those settings, so a pre-existing endpoint override redirects these uploads away from the operator’s AWS bucket. That can cause retries with otherwise correct inputs, or even “Upload complete” without delivering the files if the alternate endpoint accepts them. Please set `AWS_IGNORE_CONFIGURED_ENDPOINT_URLS=true` for this invocation, or explicitly unset both endpoint variables.

0xsiddharthks: Fixed in 2fd8d65: the upload now runs with AWS_IGNORE_CONFIGURED_ENDPOINT_URLS=true. I reproduced it first against a local endpoint that accepts every PUT: with AWS_ENDPOINT_URL_S3 or AWS_ENDPOINT_URL set, the script sent all five files there and printed success; with the flag, nothing reaches that endpoint and the upload goes to S3 on awscli 2.35.11 and 2.36.43.

dlukegordon (approved):
Reviewed at 2fd8d653. The inherited AWS endpoint override finding is fixed and its thread is resolved. No remaining actionable findings from the source and dependency-documentation review. No independent tests, YubiKey interactions, or live AWS operations were run.

PR #1199, opened 2026-09-17
[operator] Walk a guardian operation step by step

0xsiddharthks:
> Stacked on #1197.

## Summary

Each guardian operation is a long sequence whose steps have to happen in order, and only some of
them belong here: the packet steps in this repository, every deployment step in sui-operations.
Keeping that order in a runbook means reading it right under time pressure.

Put it in one script instead, so the runbook is the thing you run.

## Changes

- `guardian-operation.sh <flow> <name> [step]` for the `provision`, `rotate` and `rotate-kp-set`
  flows: with no step it prints the steps and which one is next, with a step it runs that one, and
  it refuses a step out of order.
- Progress lives in `.hashi/guardian-runs/<flow>-<name>/`, which also holds the rendered bundles and
  the collected submissions.
- Deployment steps are never run here by default: the script prints the sui-operations command and
  `--done` records it. With `HASHI_GUARDIAN_DRIVER` set to that driver it offers to run each one
  after a confirmation.
- `operator/README.md`: how to run an operation, and the two kinds of step.


0xsiddharthks:
Superseded by #1201: the driver moves into hashi, so a step-walker that hands deployment steps to sui-operations is no longer the right shape.

PR #1202, opened 2026-09-18
Resend the AVID nonce dealer's round 1 to non-confirmers, persist its confirm cert, and verify complaints off the lock

zhouwfang:
#531


crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -725,7 +725,15 @@ impl MpcManager {
         let cache_is_current = request.epoch == self.mpc_config.epoch;
         if cache_is_current && let Some(cached_response) = self.complaint_responses.get(&cache_key)
         {
-            return Ok(cached_response.clone());
+            return Ok(PreparedComplaint::Ready(cached_response.clone()));
+        }
+        let reject_key = (caller, cache_key);
+        if cache_is_current && self.complaint_rejections.contains(&reject_key) {
```
bmwill: First pass with a claud review found this, it sounds plausible but i'm not sure till i dig a bit more deeply into the flow.

---

I think this rejection memo is keyed too broadly. For nonce complaints, cache_key is ComplaintResponsesKey::NonceGeneration { batch_index, dealer }, so the key covers neither the complaint's kind (AvidReveal vs. AvidBlame) nor its content. After one complaint from an accuser fails verification for a given (dealer, batch_index), every later complaint from that accuser for that dealer and batch is refused until the epoch ends, without being checked.

That's the right result when the accuser resends the same complaint. But the accuser side doesn't always send the same one. The laggard path builds the complaint from whichever echoes it verified in that attempt, so a later attempt with a different set of bundles can produce a different AvidBlame complaint, or an AvidReveal complaint instead of a blame one. If the first complaint fails here, the later one gets "a complaint from this accuser for this dealer already failed verification" even if it's valid, and that accuser can't recover its shares through complaints for that batch.

The failures inside verify depend only on the complaint itself plus our local state for the round. So we could keep the memo's protection against repeat floods and still re-check a new complaint by adding a digest of the complaint to the key, for example:

let reject_key = (caller, cache_key, complaint_digest(&request.complaint));

where complaint_digest is a Blake2b256 hash of the BCS-encoded complaint. Adding just the complaint kind would be a cheaper partial fix, but it would still block a second blame complaint built from different echoes.

---

zhouwfang: I don't think this can hit an honest accuser. fastcrypto decides the complaint kind from the dealer's committed data, not from which echoes arrived, so an honest accuser's complaints for a batch all get the same verdict. A digest in the key would let an attacker dodge the memo by changing one byte per complaint.

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -218,6 +217,7 @@ pub struct MpcManager {
     pub message_responses: HashMap<MessageResponsesKey, MpcResult<SendMessagesResponse>>,
     pub complaints_to_process: HashMap<ComplaintsToProcessKey, ProtocolComplaint>,
     pub complaint_responses: HashMap<ComplaintResponsesKey, ComplaintResponse>,
+    pub complaint_rejections: HashSet<(Address, ComplaintResponsesKey)>,
```
benr-ml: I think that this cache should be more general, preferably in a follow up pr:
- add `identified_bad_nodes : HashSet<address>` 
- add nodes that sent provable bad messages (e.g., invalid complaint)
- ignore those parties in the relevant places (e.g., complaints, avss dealer's message, etc)

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -2682,7 +2780,28 @@ impl MpcManager {
             AvidNonceMessageKind::Dispersal {
                 dispersal,
                 confirm_cert,
```
benr-ml: Nothing here checks that `confirm_cert.message().dealer_address == sender` and `confirm_cert.message().batch_index == batch_index`. `AvidCertificate::confirm` only takes `messages_hash`, and `process_avid_message` / `AvidCertificate::verify` check the binding to `common`, the weight, the epoch and the BLS signature, but not these two fields. Today we rely on a `common` never being valid under two different (dealer, batch) pairs, i.e. on the random oracle binding them. Could we add the two explicit checks as defense in depth? (Pre-existing, not introduced by this PR.)

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -6606,6 +6614,18 @@ fn consume_certified_nonce_outputs<T>(
...
 }
 
+#[allow(clippy::large_enum_variant)]
+pub(crate) enum PreparedComplaint {
+    Ready(ComplaintResponse),
+    Verify {
+        cache_key: ComplaintResponsesKey,
+        reject_key: (Address, ComplaintResponsesKey),
+        epoch: u64,
+        verify: Box<dyn FnOnce() -> MpcResult<ComplaintResponse>>,
```
benr-ml: hmm, can't we use a value here instead of a callback?

crates/hashi/src/mpc/mpc_except_signing.rs
```diff
@@ -847,25 +865,89 @@ impl MpcManager {
...
                 };
-                let response = receiver.handle_complaint(
-                    complained_message,
-                    accuser_id,
-                    complaint,
-                    &complained_output,
-                )?;
-                ComplaintResponse::Rotation(response)
+                let complaint = complaint.clone();
+                Box::new(move || {
```
benr-ml: why not evaluate it here?

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Makes sense, I dropped it and will do a per-node version in a follow-up. It'll only limit what we serve that node, so dealer sets can't diverge. The previous-epoch case and the rest of the lock work aren't must-fix, since they only cost delay, so I'm tracking them as follow-ups.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Agreed, added the check. The blame vote cert needed it too, since I'd dropped main's batch binding when I moved verification off the lock.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Switched it to a value that's built under the lock and run after it's released. Behavior is the same.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: Evaluating it here would put the verification back under the write lock, which is what this change avoids.

crates/hashi/src/mpc/mpc_except_signing.rs
zhouwfang: I dropped the memo in this PR per Ben's suggestion, so this is moot here.

crates/hashi/src/mpc/mpc_except_signing.rs
benr-ml: i'm against such changes unless we see that it's an issue - the resulting code is very hard to follow and reason about (basically we implemented delayed verification instead of verifying on the callsite)

zhouwfang: @benr-ml Fair, I reverted it, so complaints are verified at the call site under the lock again. We can revisit if the lock hold ever turns out to be a problem.

PR #1203, opened 2026-09-18
Use the error-correcting aggregation from fastcrypto

jonas-lj:
Depends on MystenLabs/fastcrypto#1040, which moves the Reed-Solomon error correction into `aggregate_signatures`. This drops the local `aggregate_signatures_with_recovery` helper and merges `recover` into `aggregate`.

The fastcrypto pin points at that PR's branch commit for now.


jonas-lj:
The four failing AVID tests are fixed in #1205 and this will be rebased on it once it lands.

crates/hashi/src/mpc/signing.rs
```diff
@@ -1210,13 +1185,13 @@ async fn try_finalize_signature(
...
         match ctx.aggregate(unflagged[..t].to_vec(), metrics).await {
-            Ok(sig) => return FinalizeOutcome::Done(sig, Vec::new()),
+            Ok((sig, mismatched)) => return FinalizeOutcome::Done(sig, mismatched),
             Err(FastCryptoError::InvalidSignature) => {}
             Err(e) => return crypto_error(e),
         }
     }
     if st.recovery_attemptable(t) {
-        match ctx.recover(st.partials.clone(), metrics).await {
+        match ctx.aggregate(st.partials.clone(), metrics).await {
```
zhouwfang: I think with the t + f margin hashi will almost never get indices back in prod. We decode as soon as we can correct, so we rarely have t + f points kept unless the bad weight is close to f. That quietly turns off the mismatch metric and the flags. The flags don't need to be sound, so could aggregate_signatures also return the unproven set and we only count the proven ones?

crates/hashi/src/mpc/signing.rs
```diff
@@ -1210,13 +1185,13 @@ async fn try_finalize_signature(
...
             Ok((sig, mismatched)) => return FinalizeOutcome::Done(sig, mismatched),
             Err(FastCryptoError::TooManyErrors(_) | FastCryptoError::InvalidSignature) => {}
```
zhouwfang: TooManyErrors can't come back from aggregate_signatures anymore, so this arm and the one at :1206 can go.

crates/hashi/src/mpc/signing.rs
```diff
@@ -1194,14 +1169,14 @@ async fn try_finalize_signature(
...
-        threshold,
+        params,
     };
     let crypto_error =
         |e: FastCryptoError| FinalizeOutcome::Failed(SigningError::CryptoError(e.to_string()));
     if !st.clean_attempted {
         st.clean_attempted = true;
         match ctx.aggregate(st.partials[..t].to_vec(), metrics).await {
-            Ok(sig) => return FinalizeOutcome::Done(sig, Vec::new()),
+            Ok((sig, mismatched)) => return FinalizeOutcome::Done(sig, mismatched),
```
zhouwfang: If we keep the gated version, the Done doc at :1091, the warn at :777 and the metric help in metrics.rs still describe the old meaning, so they're worth updating.

zhouwfang (commented):
CI is failing because e2e-tests/src/lib.rs:1133 still passes t instead of Parameters.

crates/hashi/src/mpc/signing.rs
jonas-lj: Dropped from both arms. aggregate_signatures maps every decoder error to InvalidSignature.

crates/hashi/src/mpc/signing.rs
jonas-lj: Done — [MystenLabs/fastcrypto#1040 · Correct faulty partial…](https://github.com/MystenLabs/fastcrypto/pull/1040) now returns a Blame enum. Both non-empty variants are flagged here, but only `Certain` increments the metric.

crates/hashi/src/mpc/signing.rs
```diff
@@ -740,46 +748,59 @@ impl SigningManager {
         }
     }
 
+    /// Flag every share index whose owner `blame` names, so the rest of this call skips their
```
zhouwfang: nit: the full attempt still uses flagged partials, so "skips" overstates it a bit. Also nothing tests that Inconclusive flags without counting, so a small flag_mismatched test would pin that.

crates/hashi/src/mpc/signing.rs
```diff
@@ -1194,14 +1185,24 @@ async fn try_finalize_signature(
...
         vk: *verifying_key,
         deriv: st.derivation_address,
-        threshold,
+        params,
+    };
+    let crypto_error = |e: FastCryptoError| {
+        let reason = match e {
+            // The partials were ruled out as the cause, so collecting more cannot fix this.
+            FastCryptoError::InconsistentInputs => {
+                "the presigning tuple, beacon, message or verifying key does not match the one \
```
zhouwfang: Same as on fastcrypto, a wrong derivation address lands here too.

crates/hashi/src/mpc/signing.rs
jonas-lj: Re the test, it would mostly restate the match, and flag_mismatched already runs end-to-end with Blame::Certain in test_sign_one_corrupted_rs_recovery and test_sign_multiple_corrupted_rs_recovery. And it wouldn't cover the non-trivial part anyway (that we flag all of an owner's indices, not just the ones in the Blame), since SigningTestSetup is weight 1 everywhere. So I'd rather not tbh.

zhouwfang: Fair enough. One small correction: those two tests never actually reach Certain, since the decode finishes before there's enough margin. So the metric path isn't covered, but I'm fine leaving it.

zhouwfang (commented):
LGTM once the pin moves to fastcrypto main and this is rebased on #1205.

PR #1205, opened 2026-09-19
Use the AVID output type for presigning

jonas-lj:
Depends on MystenLabs/fastcrypto#1045, which removes the legacy `batch_avss` module and makes `Presignatures::new` take the AVID output type directly, without the `use_legacy` flag. This drops the `into_legacy` conversions and `USE_LEGACY_PRESIG_DERIVATION` (which was already `false`, so production presignatures are unchanged).

Bumping fastcrypto also brings in MystenLabs/fastcrypto#1027, which requires AVID vote certificates to have `W − f` weight of signers. Four AVID tests built their vote certificates with one signer too few, so they now include the dealer's own vote.

The fastcrypto pin points at the branch commit of #1045 for now.


Cargo.toml
```diff
@@ -89,8 +89,8 @@ aws-credential-types = "1"
 aws-sdk-s3 = "1.12.0"
 
 # fastcrypto — pinned to a `main` rev.
-fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto.git", rev = "e59b4a412c3f1cb2d477061e146de55276b6950d" }
-fastcrypto-tbls = { git = "https://github.com/MystenLabs/fastcrypto.git", rev = "e59b4a412c3f1cb2d477061e146de55276b6950d" }
+fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto.git", rev = "aa9653f0668833df19a8a29f44715075b11f3e33" }
```
zhouwfang, line 92: Could we rebase MystenLabs/fastcrypto#1045 onto fastcrypto main and pin that instead? Main is two commits ahead, and one of them reverts the hard errors this bump picks up. complete_dkg is the one that worries me, since we call it with .expect, so a new Err there panics rather than errors.

crates/hashi/src/mpc/mpc_except_signing_tests.rs
```diff
@@ -12479,19 +12479,19 @@ fn test_decode_avid_nonce_share_reconstructs_from_echoes() {
...
         .unwrap();
     let decoder_addr = setup.address(5);
 
-    // Four voters process their dispersal -> Vote sigs (for the W−f cert) and echoes for the decoder.
+    // The voters process their dispersal -> Vote sigs (for the W−f cert) and echoes for the decoder.
     let mut vote_sigs = Vec::new();
     let mut avid_vote = None;
     let mut echoes = Vec::new();
-    for j in [1usize, 2, 3, 4] {
+    for j in [0usize, 1, 2, 3, 4] {
```
zhouwfang, line 12494: Checked this and the tests really were wrong, production can't emit a short cert. The fixture now sits exactly on W−f=5 though, so dropping a voter later re-breaks all four with an error that reads like a crypto bug. Asserting against avid_vote_quorum would name the real cause.

crates/hashi/src/mpc/mpc_except_signing_tests.rs
```diff
@@ -12479,19 +12479,19 @@ fn test_decode_avid_nonce_share_reconstructs_from_echoes() {
     let mut rng = rand::thread_rng();
     let setup = TestSetup::new(6);
     let batch_index = 0u32;
-    // Confirmers {0..4}, decoder = node 5. Decode needs W−2f=2 echoes; the Vote cert needs W−f=4.
+    // Confirmers {0..4}, decoder = node 5. The Vote cert needs W−f weight.
```
zhouwfang, line 12482: The old numbers here were both wrong, W−2f is 4 and W−f is 5 for this 6-node f=1 fixture. That's probably what made a 4-signer cert look right. Worth keeping the decode bound rather than dropping it.

